PR #2, opened 2015-01-13
Index browser

jonasfj:
This builds on PR #1 ...

It adds a generic index browser... after that we still need to build an artifact index browser... but we can reuse the index browser part from this... It's designed specifically for this...

Note,
There are a few paper cuts we can improve later... Like it would be nice if tabs were disabled when there is no namespaces or tasks available... Perhaps we could add icons... and we could merge the breadcrumbs and input field into one widget as offered by nautilus, etc... But generally this works...

Remark, I've deployed this to `tools.taskcluster.net`, so you can try it out...
(Note self, refactor so we can deploy this on github pages etc...)


jhford:
I'm having that problem where I can't run the tools package locally.  Each page load is just a blank page with the correct favicon, <title> but a blank white page and an error rendering a component in the debug log.  I forgot what was happening before, do you remember what the fix is?

jonasfj:
Try clearing node_modules and reinstall, I updated deps....

jonasfj:
also what platform are you on?

jonasfj:
@jhford, btw, I've published this and #3 to tools.taskcluster.net, so you can try it out there...
But yeah, we really should make sure this builds on multiple computers :)

PR #3, opened 2015-01-14
Indexed artifact browser

jonasfj:
An artifact browser based on the generic index browser....

Merge if you have no objections :)

Remark, this is based PR #2 with is based on PR #1, so there is a lot stacked up here.
Also I've already published this...


jhford:
Having both the namespace and task name in the fragment is unfortunate, but I guess there's really no other way when the namespace namespace is a different namespace than the task namespace.  I wish we could name namespaces something else.

It would nice to have the browser automatically switch to task or namespaces when the selected one is empty, or showing both lists on the left side of the screen.

This looks great though, I'll merge it now.

jonasfj:
@jhford,
I was actually considering a list that showed both namespaces and task... But with two buttons for each entry, one button to browse the namespaces and another to view it as a task... or something like that...

Either solution gets a little messy... Maybe we can improve it later, at the moment it's good enough :)

jhford:
agreed, this is good and integrated in tools.

PR #4, opened 2015-02-12
Use b2gtest instead of cli so we likely have a running machine

lightsofapollo:


jonasfj:
I have no good excuse for why we didn't merge this right away...

PR #5, opened 2015-04-20
First pass of provisioner UI

jhford:
This is the first pass of the provisioner ui.  Right now, it will show the stats of the running instances, a count of pending jobs and a json display of the workerType definition.

It's not done, but I'd like to deploy this first step as a way to get visibility on the provisioner deployment.

Key things which I will be changing as time goes on:

* the top level component will be something which figures out whether to show a list, detail, edit or create.  right now, i just have the list as the top level
* use the URL fragment to decide which of these views to show
* editing of worker type definitions
* creating of worker type definitions
* a 'verify' button that will preview the launchspecificions using the server's test worker type
* a details page which shows high level stats (running, pending, etc) and a list of instances with selected metadata.
* a provisioner end point which can kill a specific spot request or instance and hook it up to the UI so that the provisioner UI can kill an instance directly.

### DIFF
--- a/aws-provisioner/awsprovisioner.jsx
+++ b/aws-provisioner/awsprovisioner.jsx
@@ -66,8 +66,8 @@ var WorkerTypeTable = React.createClass({
   load: function() {
     return {
       workerTypes: this.awsProvisioner.listWorkerTypes(),
-      //awsState: this.awsProvisioner.updateAwsState(),
-      awsState: this.awsProvisioner.awsState(),
+      awsState: this.awsProvisioner.updateAwsState(),
+      //awsState: this.awsProvisioner.awsState(),
     };
   },
   
@@ -180,7 +180,7 @@ var WorkerTypeRow = React.createClass({
         });
       });
 
-      s.pending.forEach(function(node) {
+      s.spotReq.forEach(function(node) {
         w.instanceTypes.forEach(function(itd) {
           if (itd.instanceType === node.LaunchSpecification.InstanceType) {
             console.log('Adding ' + itd.capacity + ' to requested capacity');
--- a/aws-provisioner/awsprovisioner.jsx
+++ b/aws-provisioner/awsprovisioner.jsx
@@ -66,6 +66,7 @@ var WorkerTypeTable = React.createClass({
   load: function() {
     return {
       workerTypes: this.awsProvisioner.listWorkerTypes(),
+      //awsState: this.awsProvisioner.updateAwsState(),
       awsState: this.awsProvisioner.awsState(),
     };
   },
@@ -135,12 +136,10 @@ var WorkerTypeRow = React.createClass({
 
   getInitialState: function() {
     return {
-      pendingTasks: {
-        pendingTasks: 'loading',  
-      },
+      pendingTasks: {},
       pendingTasksLoaded: true,
       pendingTasksError: undefined,
-      workerType: 'loading',
+      workerType: undefined,
       workerTypeLoaded: true,
       workerTypeError: undefined,
     };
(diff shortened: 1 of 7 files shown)


aws-provisioner/awsprovisioner.jsx
```diff
@@ -0,0 +1,372 @@
...
+            }
+            return <WorkerTypeRow key={name} awsState={realState} workerType={name} removeWorkerFromTable={that.removeWorkerFromTable}/>;
+          })
+        }
+        </tbody>
+      </bs.Table>
+    );
+  },
+  removeWorkerFromTable: function(name) {
+    this.setState({
```
jonasfj: This is a bad idea... just give the child (`WorkerTypeRow`) a reference to `this.reload` which is implemented by the taskcluster mixin...

aws-provisioner/awsprovisioner.jsx
```diff
@@ -0,0 +1,372 @@
...
+        {
+          this.renderWaitFor('workerTypes') || this.renderWaitFor('awsState') || this.state.workerTypes.map(function(name) {
+            var realState;
+            if (that.state.awsState[name]) {
+              realState = that.state.awsState[name];
+            } else {
+              realState = {
+                running: [],
+                pending: [],
+                spotReq: [],
```
jonasfj: We really should offer an API end-point that only returns the numbers... not arrays.. We rarely need the actually arrays of ids...

aws-provisioner/awsprovisioner.jsx
```diff
@@ -0,0 +1,372 @@
...
+                      worker={this.state.workerType}
+                      awsState={this.props.awsState}
+                      runningCapacity={runningCapacity}
+                      pendingCapacity={pendingCapacity}
+                      spotReqCapacity={spotReqCapacity}
+                      progressBar={progressBar}
+            />}>
+          <bs.Button bsStyle='primary' bsSize='xsmall'>View</bs.Button>
+        </bs.ModalTrigger>
+        <bs.Button bsStyle='danger' bsSize='xsmall' onClick={this.removeWorkerType}>Remove</bs.Button>
```
jonasfj: Use the `<ConfirmAction>` construct I made for schedule, rerun, and cancel task in `taskinfo.jsx`...
You give it text, icon and method to call that returns a promise on execution :)

Inside that method you call the api, and `.then` call `.reload()` on parent... So it the row is removed from the list...

jonasfj: Note, that the `<ConfirmAction>` also displays errors the usual way... Ie. using `renderWaitFor` which calls `renderError`. So if we want prettier errors, we'll implement that in `createTaskClusterMixin` then everybody all over will get that...

aws-provisioner/awsprovisioner.jsx
```diff
@@ -0,0 +1,372 @@
...
+  render: function() {
+    return (
+        <WorkerTypeTable />
+    );
+  },
+
+});
+
+// Export IndexBrowser
+module.exports = AwsProvisioner;
```
jonasfj: You probably just want to export `<WorkerTypeTable>` :)

And maybe even split `WorkerTypeRow` into another file... Or definitely consider that when you added editor and other awesome things...

jonasfj:
I'm generally very happy how well this looks... And how you managed to use `createTaskClusterMixin` I was a bit scared that it was a monster... But you seem to understand how to use it too :)

I also have logic for syncing to `window.hash` it's also a bit of a monster, but work nice once setup...

Nits. line width, please stay within 80 chars...

Anyways, I'm merging and publishing this... I don't see what possible harm this could do...
I didn't manage to very that it worked because no workerTypes existed on the provisioner. It just returned an empty array of workerTypes...

jonasfj:
deployed... please add a workerType so we can play more with it...

aws-provisioner/awsprovisioner.jsx
jhford: Well, i could have that end point, but right now I load this data once and then reuse it for the details list.  I think it's important that the details page contains info like region, instanceType, avail. zone and instance/sr ids.

I also trim down the objects sent back to exactly the subset of keys that I care about here.

I'd rather do it this way

jonasfj: I'm probably just missing the docs to know what it actually returns... :)

aws-provisioner/awsprovisioner.jsx
jhford: Done!  much nicer.  I couldn't figure out why this wasn't working, but it was a `this` vs `that` bug.  ugh.  I also made it so the whole table is a spinner until loaded instead of just the first cell.

aws-provisioner/awsprovisioner.jsx
jhford:
```javascript
/**
 * These lists are the keys for the object filtering we'll do for their
 * respective aws state objects.
 */
var filterForSR = [
  'CreateTime',
  'State',
  'LaunchSpecification:InstanceType',
  'LaunchSpecification:ImageId',
  'LaunchSpecification:Placement:AvailabilityZone'
];
var filterForInstance = [
  'InstanceId',
  'ImageId',
  'InstanceType',
  'LaunchTime',
  'Placement:AvailabilityZone',
  'SpotInstanceRequestId',
];
```

These are AWS objects, with these filters applied to limit it to a small subset of properties...  I'll probably move the filtering to the methods which get the state to reduce memory usage in the awsmanager at some point

PR #6, opened 2015-04-22
Pulse-inspector grammar fixes

djmitche:
Minor, but it bugged me..


jonasfj:
Fair enough, thanks for the fix :)

PR #8, opened 2015-04-27
Create and Update implemented!

jhford:
I've implemented the create/update portions of the UI.  The only thing that's not working right now is reloading of the table after creating a new workerType.  Theoretically, we might also want to rebuild the table for updating in case we chagned the capacity of any workerTypes or their maxCapacity values, which would change the bar graphs.

THe problem is that the reload is 'owned' by the table and both create/update fields are siblings to it, not children.  I was thinking of having the top level component have a taskcluster mixin with no client and use it's .reload() method for the table and the create/update.  This might also be crazy.

### DIFF
--- a/aws-provisioner/awsprovisioner.jsx
+++ b/aws-provisioner/awsprovisioner.jsx
@@ -307,8 +307,9 @@ var WorkerTypeRow = React.createClass({
   },
 
   removeWorkerType: function() {
+    var that = this;
     return this.awsProvisioner.removeWorkerType(this.props.workerType).then(function() {
-      return this.props.reload();
+      return that.props.reload();
     });
   },
 });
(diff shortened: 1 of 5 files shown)


jonasfj:
Let's iterate a bit on this... I don't mind deploying it... but at the moment if you first view one instanceType and then view another the editor doesn't update...

This is a bit scary because you can easily update the wrong instanceType, or overwrite an instance type with a definition from another instancetype... 

I'll play around with it today and see how far I get... We can merge and publish, but this footgun is pretty big and scary...

jonasfj:
@jhford, merged and published along with a fix for the forementioned footgun... This still gives the impression that workerType name can be changed with an update....

For details on fix please review:
https://github.com/taskcluster/taskcluster-tools/blob/586fec287a68e63f9ce3f0e409aa894216073c78/aws-provisioner/workertypeedit.jsx#L67-L79

When setting state using a property, it's usually necessary to implement `componentWillReceiveProps`.
Btw, notice how nice it is to make methods on objects with babeljs :)

jonasfj:
Pro tip, review: https://facebook.github.io/react/docs/component-specs.html

PR #9, opened 2015-04-30
Edit UserData as json instead of base64

jhford:
I've tested editing and creating.. it works for me.


jhford:
@jonasfj Hey, any reason we shouldn't be deploying this?

jonasfj:
Sorry... I have absolutely no excuse... I saw this and have failed to check it out and land it.

I'll try to land it today...

jonasfj:
@jhford,
We can now deploy just by pushing to master... So no need for me to do `grunt publish` :)

IMO; reviews here should be minimal.. It's good to have some for major changes.

PR #10, opened 2015-05-08
Reworked provisioner UI, hash state, console links and other nice thi…

jonasfj:
* Updated `font-awesome` to get `server` icon.. :)
 * Cleanup a lot in provisioner UI
 * Added links to AWS console
 * Added status overview that sums up numbers from AZ/instanceTypes
 * Synchronization to `location.hash`
 * A bunch of CSS hacks to reduce margin and tweak things..

Overall it's rather nice now... We might want to simplify format offered by provisioner... but updating this should be somewhat simple...

btw, babel + react.js is pretty awesome... With ES5 we can do `{prop() {return 4;}}` instead of `{prop: function() {return 4;}}`. This makes react.js super sweet :)

@jhford, feel free to review code in `aws-provisioner/`, IMO just skimming it should be fine... But in general just try it out.. and if you have no objections to the UI merge it to master and circle CI will roll it out...


jhford:
Looks great, thanks for the improvements!

PR #11, opened 2015-05-09
Hack to infer a script for running a task locally

jonasfj:
Example:
http://postimg.org/image/kxmhmhyhj/

@gregarndt, quick review... check if what I'm doing here is crazy...
If you have a clue about how to illustrate usage of `--link` for:
 * taskclusterProxy
 * testdroidProxy
 * balrogVPNProxy
For example which docker images to use and how to inject credentials into them that would be awesome.

Obviously, we can display credentials here :)
But we can show that they are inject by env variables, example:
```bash
docker run -d \
  --name balrog-vpn-proxy \
  -e PWD=$MY_VPN_PASSWORD \
  taskcluster/balrogvpnimage:0.2.1 \
 ;

docker run -ti --link balrog-vpn-proxy ... . .. # all the things from below...
```

@gregarndt, also on a slightly crazy note...
Would it be completely insane if every instance of docker-worker uploaded a javascript file to:
`references.taskcluster.net/workers/<provisionerId>/<workerId>/run-locally.js`

Such that we can include that file on tools.taskcluster.net and use it to generate the illustrative script.
I say illustrative because it's hard to generate something that always works, but if we could just come somewhat close that would really nice.
Obviously, if using something like a special loopback video device the script returned might include a short explanation saying that you'll need to configure this very complicated thing first, he he :)

-----
I think that even if people don't use this it's very educational, in giving a quick idea of how this all works.

### DIFF
--- a/lib/ui/taskinfo.jsx
+++ b/lib/ui/taskinfo.jsx
@@ -5,6 +5,7 @@ var format          = require('../format');
 var _               = require('lodash');
 var taskcluster     = require('taskcluster-client');
 var ConfirmAction   = require('./confirmaction');
+var path            = require('path');
 
 /** Displays information about a task in a tab page */
 var TaskInfo = React.createClass({
@@ -27,7 +28,7 @@ var TaskInfo = React.createClass({
   },
 
   /** Get initial state */
-  getInitialState: function() {
+  getInitialState() {
     return {
       // task definition
       taskLoaded:   false,
@@ -37,13 +38,13 @@ var TaskInfo = React.createClass({
   },
 
   /** Load task definition */
-  load: function() {
+  load() {
     return {
       task:         this.queue.task(this.props.status.taskId)
     };
   },
 
-  render: function() {
+  render() {
     // Easy references to values
     var status  = this.props.status;
     var task    = this.state.task;
@@ -172,7 +173,7 @@ var TaskInfo = React.createClass({
             task.scopes.length > 0 ? (
               <ul>
                 {
-                  task.scopes.map(function(scope, index) {
+                  task.scopes.map((scope, index) => {
                     return <li key={index}><code>{scope}</code></li>;
                   })
                 }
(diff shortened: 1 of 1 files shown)


jonasfj:
Oh, feel free to merge this, I've tested it and once you merge it'll roll out...

gregarndt:
In regards to the balrog and testdroid proxies, this will be much more difficult.  The balrog proxy contains the credentials and the VPN keys embedded within the image.  We could possible have a balrog vpn image where the user passes in a volume that contains the keys and credentials to use.  Perhaps in the future we can have a secret place for these things and the worker just volume mounts it in instead of worrying about images with them embedded.. Or there is a read only data container that has the creds and we expose the volume from that so that the balrog piece of it is generic enough to use anywhere. 

the testdroid proxy is a different story.  We pass in the credentials for that right from the worker and probably want to restrict how many people have access to that.  Sure people could pass in their own credentials but right now I'm not sure who we would give those out to.

gregarndt:
Oh, and if we want to get fancy, why not just supply a docker compose config  for the task container along with all the linked things? :)

gregarndt:
Regarding the JS file and setup of loopback devices: I'm all for the code to run it locally living closest to the worker that actually does the work, that doesn't worry me much.  We have a vagrant file part of docker-worker now that has the video and audio loopback devices created already.  They could just `vagrant ssh` into the instance and run the stuff in there.

jonasfj:
@gregarndt,
So my goal here is not to give a script that always works, as in copy to shell and run task locally.

My goal is to give a good idea of how this all works. So for say balrog, we just say:
```
# Get private image for balrog-vpn-proxy
# (Purely illustrative, ask for access if you really need it) 
docker login -u <username> -p <password> quay.io/secret;
docker pull query.io/secret/balrog-vpn-proxy;
```

I don't care that people can't copy paste the above... If it just illustrates there here you pull a private docker image... Then you can either give up replicating that part, or get access from one of us, if you should have that access...

gregarndt:
:+1:

PR #12, opened 2015-05-21
Update name env variable and stop script on error

gregarndt:


jonasfj:
thanks, for the fix! :)

PR #15, opened 2015-06-16
Changes to the provisioner UI to support the secrets/userdata changes

jhford:
support the new secrets schema changes.  because we just use JSON fields to edit things, we're basically just undoing the workarounds for the UserData as Base64 encoded json stuff.


jonasfj:
@jhford,
`r+` feel free to land it... It'll auto-deploy once you hit the button :)

PR #16, opened 2015-06-29
Show JSON parse errors

djmitche:
This helps newcomers to write valid JSON.

A possible extension would be to validate the JSON against the schema as well, but this will do for now.

### DIFF
--- a/task-creator/taskcreator.jsx
+++ b/task-creator/taskcreator.jsx
@@ -152,6 +152,9 @@ var TaskCreator = React.createClass({
           Reset Editor
         </bs.Button>
       </bs.ButtonToolbar>
+      <span hidden={!this.state.invalidTask}>
+        {this.state.invalidTaskError}
+      </span>
       </span>
     );
   },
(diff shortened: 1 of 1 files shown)


jonasfj:
I was actually looking at:
 1. Enabling hinting with CodeMirror, or
 2. Use ace.js which has pretty good hinting built-in

This way you would get pretty much what we have here:
http://docs.taskcluster.net/tools/task-creator/

With error boxes in the side, which is very nice...

------

I'll merge this, if I don't figure out how to get syntax validation working in the editor... which would be much nicer...

djmitche:
I'm confused -- how is the docs version newer than the version in the git repo?  And why can't you just deploy that to tools?

jonasfj:
Docs is not newer... We used a different editor for tools... And this feature slipped...

So it's more like a regression...

jonasfj:
The fix in #17 is better... It annotates at the line (or approximately the line) where the error is located... Also have slightly better messages...

A step crazier would be to add a linter that can consume JSON schemas :) that would be fun...

Reopen this, if #17 didn't solve the problem... It should be deployed a few mins...

PR #17, opened 2015-06-29
Json lint

jonasfj:
Add JSON errors to editors using a linter...

Invalidating #16


jonasfj:
Feel free to land it once review... This will auto-deploy... Basically you get the equivalent of:
https://codemirror.net/demo/lint.html

But with better JSON errors because we use a different module... The old one wasn't compatible with browserify...

PR #18, opened 2015-07-06
Interactive

EggyLv999:


package.json
```diff
@@ -43,7 +43,9 @@
     "exorcist":                         "0.1.6",
     "concat-stream":                    "1.4.6",
     "grunt-keepalive":                  "0.0.1",
-    "grunt-concurrent":                 "1.0.0"
+    "grunt-concurrent":                 "1.0.0",
+    "term.js":                          "chjj/term.js.git",
```
jonasfj: can't we use `term.js` as already imported? See line 28

jonasfj:
It would be lovely to react'ify this... Say we created an interactive terminal widget that just took a `socketUrl` as property and then setup an SSH-like session.

I wrapped `term.js` for use as livelog viewer here: https://github.com/taskcluster/taskcluster-tools/blob/master/lib/ui/terminalview.jsx

Anyways, that's just nice to have... And quite powerful as it would make it easy to embed the interactive terminal whenever we feel like it would be fun :)

jonasfj:
Also this probably need to be rebased... (or merged) I don't care, it still has merge conflicts.

jonasfj:
Note, if not clear -- I'm very happy with this, making this more react based is just nice to have, don't block on it :)

interactive/app.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+  var args = qs.parse(url.parse(window.location.href).query);
+
+  var client = new DockerExecClient({
+    url: args.socketUrl,
+    tty: true,
+    command: '/bin/bash'
+  });
+
+  client.execute().then(function () {
```
jonasfj: FYI, all this code is compiled with babeljs, so you can use async function, await and arrow functions :)

PR #19, opened 2015-07-29
Bug 1180807: don't show access tokens except when they are issued

djmitche:
Note that the tokens are still in memory in the browser -- this just
fixes the display issue.  However, it also prepares tools for a
situation where `accessToken` is omitted or blank in all responses
except createClient and resetClient.

Testing: I tested this by hand, both creating new clients and resetting passwords for existing clients.

### DIFF
--- a/auth/clienteditor.jsx
+++ b/auth/clienteditor.jsx
@@ -59,6 +59,7 @@ var ClientEditor = React.createClass({
           name:           "",
           description:    ""
         },
+        newAccessToken:   '',
         editing:          true,
         working:          false,
         error:            null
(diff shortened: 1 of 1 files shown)


jonasfj:
Hmm... Not sure about this... I think we have to fix the API first...
IMO the accessToken should just be showed in a popup box after hitting `create` button or `reset` button.

djmitche:
This patch accomplishes just that (depending on how literally you mean "popup" -- it's not `alert`, although that might be easier!).  If we fix the API to not provide `accessToken`, we'd need to apply the equivalent of this patch anyway to avoid undefinition problems.

jonasfj:
ahh, I better read it properly then.. sorry :)

djmitche:
Jonas - is there any reason not to merge and deploy this?

jonasfj:
Sorry, no reason... I forgot about it... I'll take a look now..

jonasfj:
hmm... Looks alright to me... Let's roll it out and see if it works..

Sorry for the delay... Ideally I would have loved to have the API updated (but I'm working on that).

PR #20, opened 2015-07-31
Efficiency

EggyLv999:
Well, this is my extremely sorry attempt at a tool for an efficiency calculator, since it can't really be done in grafana. Plenty of race conditions to be had in this version, but making it synchronous is not a great option either, given how long these queries are (can take 5+min). Also the ui is awful.

### DIFF
--- a/efficiency/efficiency.jsx
+++ b/efficiency/efficiency.jsx
@@ -5,12 +5,13 @@ let qs = require('querystring');
 var React = require('react');
 let request = require('superagent-promise');
 
-async function update(start, end) {
+async function update(start, end, user, pass, workerType) {
   let url = 'https://goldiewilson-onepointtwentyone-1.c.influxdb.com:8087/db/docker-worker/series';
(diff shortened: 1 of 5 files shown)


efficiency/efficiency.jsx
```diff
@@ -0,0 +1,91 @@
...
+    );
+  },
+  /** Handle form submission */
+  handleSubmit: function(e) {
+    e.preventDefault();
+    //TODO: figure out what to put in state
+    //this.setState({taskId: this.state.taskIdInput});
+    // console.log(document.getElementById("form").elements[0].value);
+    // console.log(document.getElementById("form").elements[1].value);
+    update(document.getElementById("form").elements[0].value, document.getElementById("form").elements[1].value);
```
jonasfj: `this.refs.start.getValue()` works too :)

eliperelman:
Given our movement away from Influx, we won't be taking this in. Thanks.

PR #22, opened 2015-08-10
Updated description/link to bugzilla to reflect new bugzilla product

petemoore:
I tested locally using grunt, seems to render correctly and link to bugzilla works.


petemoore:
@jonasfj @gregarndt

jonasfj:
Nice catch.. thanks...

PR #23, opened 2015-08-27
Task Inspector tasks link to Task Graph Inspector and generate nicer bash code quoting

petemoore:
Previously, TaskGraphId was shown as a plain string - with this change,
they are now shown as a hyperlink, so you can get straight from a task
to the task graph it came from.

Secondly, I've updated the "Run Locally" generated output to quote words
consistently and appropriately, and removed unnecessary ';' from end of
lines (which were redundant).

Changes tested locally and working.

### DIFF
--- a/lib/ui/taskinfo.jsx
+++ b/lib/ui/taskinfo.jsx
@@ -164,7 +164,11 @@ var TaskInfo = React.createClass({
         <dt>SchedulerId</dt>
         <dd><code>{task.schedulerId}</code></dd>
         <dt>TaskGroupId</dt>
-        <dd><code>{task.taskGroupId}</code></dd>
+        <dd>
+          <a href={'../task-graph-inspector/#' + task.taskGroupId}>
+            {task.taskGroupId}
+          </a>
+        </dd>
       </dl>
       <dl className="dl-horizontal">
         <dt>Scopes</dt>
@@ -258,14 +262,14 @@ var TaskInfo = React.createClass({
     cmds.push("# WARNING: this is experimental mileage may vary!");
     cmds.push("");
     cmds.push("# Fetch docker image");
-    cmds.push("docker pull " + payload.image);
+    cmds.push("docker pull '" + payload.image + "'");
     cmds.push("");
     cmds.push("# Find a unique container name");
-    cmds.push("export NAME='task-" + taskId + "-container';");
+    cmds.push("export NAME='task-" + taskId + "-container'");
     cmds.push("");
     cmds.push("# Run docker command");
     cmds.push("docker run -ti \\");
-    cmds.push("  --name $NAME \\");
+    cmds.push("  --name \"${NAME}\" \\");
     if (payload.capabilities && payload.capabilities.privileged) {
       cmds.push("  --privileged \\");
     }
@@ -294,7 +298,6 @@ var TaskInfo = React.createClass({
       }).join(' ');
       cmds.push("  " + command + " \\");
     }
-    cmds.push("  ;");
     cmds.push("");
     if (payload.artifacts) {
       cmds.push("# Extract Artifacts");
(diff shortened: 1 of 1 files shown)


petemoore:
@jonasfj

petemoore:
LOL, the irony! I didn't escape/quote my commit message for c13d1a5 properly, which was a commit message about a fix to the code to generate correctly quoted bash code...

lib/ui/taskinfo.jsx
```diff
@@ -164,7 +164,11 @@ var TaskInfo = React.createClass({
         <dt>SchedulerId</dt>
         <dd><code>{task.schedulerId}</code></dd>
         <dt>TaskGroupId</dt>
-        <dd><code>{task.taskGroupId}</code></dd>
+        <dd>
+          <a href={'../task-graph-inspector/#' + task.taskGroupId}>
+            {task.taskGroupId}
+          </a>
+        </dd>
```
jonasfj: This assertion only holds if `schedulerId === 'task-graph-scheduler'`, so an if/else would be very nice..

FYI, this would fix: https://bugzilla.mozilla.org/show_bug.cgi?id=1168098

lib/ui/taskinfo.jsx
```diff
@@ -294,7 +298,6 @@ var TaskInfo = React.createClass({
       }).join(' ');
       cmds.push("  " + command + " \\");
     }
-    cmds.push("  ;");
```
jonasfj: You add `;` everywhere else... why remove it from this command?

lib/ui/taskinfo.jsx
```diff
@@ -258,14 +262,14 @@ var TaskInfo = React.createClass({
...
+    cmds.push("docker pull '" + payload.image + "'");
     cmds.push("");
     cmds.push("# Find a unique container name");
-    cmds.push("export NAME='task-" + taskId + "-container';");
+    cmds.push("export NAME='task-" + taskId + "-container'");
     cmds.push("");
     cmds.push("# Run docker command");
     cmds.push("docker run -ti \\");
-    cmds.push("  --name $NAME \\");
+    cmds.push("  --name \"${NAME}\" \\");
```
jonasfj: Not sure why you wrap in `${}`, curious does it do anything?

jonasfj:
Fix the condition on `schedulerId === 'task-graph-scheduler'` for the link... And then I'll happily merge :)

lib/ui/taskinfo.jsx
petemoore: Actually, I removed it everywhere else, I didn't add it anywhere. :)

lib/ui/taskinfo.jsx
petemoore: The quotes (`""`) ensure that the shell won't interpret any of the characters in NAME. The `{}` delimeters I like to use as a matter of style - they are not necessary in this case, but in other cases would be. For example:

```bash
animal='dog'
echo "There were three ${animal}s sniffing around the park today"
```

Since often `{}` is necessary to demarcate a variable name, I tend to always use the syntax for consistency. Other benefits are often syntax highlighting in code editors, and visually I find it clearer, somehow - the variable name has more prominence, IMO. But it is not a biggy, just a preference...

jonasfj: Fine with me :)

lib/ui/taskinfo.jsx
jonasfj: Sorry cant read..

lib/ui/taskinfo.jsx
petemoore: Funnily enough I had the exact same thought on my way to school with Jasper this morning, so totally agree - I'll make the update. =)

petemoore:
| Fix the condition on `schedulerId === 'task-graph-scheduler'` for the link... And then I'll happily merge :)
Done :)

PR #26, opened 2015-09-10
Use slugid.nice() rather than slugid.v4() in taskcluster-tools

petemoore:
And finally - so that our interactive tools don't generate non-nice slugs...

### DIFF
--- a/task-graph-inspector/taskgraphinspector.jsx
+++ b/task-graph-inspector/taskgraphinspector.jsx
@@ -234,7 +234,7 @@ var TaskGraphInspector = React.createClass({
   // Render a task-graph-inspector
   render: function() {
     // Decide if input is a valid slugid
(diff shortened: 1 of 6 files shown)


task-graph-inspector/taskgraphinspector.jsx
```diff
@@ -234,7 +234,7 @@ var TaskGraphInspector = React.createClass({
   // Render a task-graph-inspector
   render: function() {
     // Decide if input is a valid slugid
-    var invalidInput = !/^[a-zA-Z0-9_-]{22}$/.test(this.state.taskGraphIdInput);
+    var invalidInput = !/^[A-Za-z0-9_-]{8}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$/.test(this.state.taskGraphIdInput);
```
jonasfj: Love it !!! :)

PR #27, opened 2015-09-21
Bug 1130049 - Unify namespaces & tasks into single list

mshal:


jonasfj:
hmm... looks sane... let's try it... It might not be perfect if we have an entry that is both task and namespace, as we'll get duplication. But we probably don't have many of those...

IMO, we should change the serverside to return a list of tasks and namespaces under the current "namespace"... But that's not possible right now I think... Or at least it'll do duplication too...

PR #28, opened 2015-10-01
Make editing tasks easier

djmitche:
* "Edit Task in Task Creator" button in task inspector
* "Update Timestamps" button in task creator


jonasfj:
Nice... will auto rollout... Thanks! :)

PR #29, opened 2015-10-02
Bug 1210393: hide the action button after success

djmitche:


lib/ui/confirmaction.jsx
```diff
@@ -69,13 +69,17 @@ var ConfirmAction = React.createClass({
...
-              disabled={this.state.executing && !this.state.resultLoaded}>
-            <bs.Glyphicon glyph={this.props.glyph}/>&nbsp;
-            <span>{this.props.label}</span>
-          </bs.Button>
+          {
+            !(this.state.executing || this.state.result) ? (
+              <bs.Button
+                  onClick={this.executeAction}
+                  bsStyle={this.props.buttonStyle}
+                  hidden={this.state.result}
```
jonasfj: missing `>`

jonasfj:
this is one of the few repositories where test works.. in fact they auto-deploy from master... :)

lib/ui/confirmaction.jsx
djmitche: Huh, I wonder how that worked for me.. >>blames grunt<<

jonasfj:
I'm guessing you used `--force` :)

jonasfj:
thanks...

PR #30, opened 2015-10-05
New auth tool

jonasfj:
New auth tool... still hard coded to use tc-auth-load-test app... for now...

almost ready to merge...

### DIFF
--- a/auth/clients/app.jsx
+++ b/auth/clients/app.jsx
@@ -1,0 +1,16 @@
+var React               = require('react');
+var ClientManager       = require('./clientmanager');
+var $                   = require('jquery');
+var utils               = require('../../lib/utils');
+
+var hashManager = utils.createHashManager({
+  separator:            '/'
+});
+
+// Render component
+$(function() {
+  React.render(
+    <ClientManager hashEntry={hashManager.root()}/>,
+    $('#container')[0]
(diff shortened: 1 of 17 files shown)


jonasfj:
@djmitche, if you have any comments on this they are welcome... but UI review isn't high on my list of priorities :)

djmitche:
For those without permissions (which is most people), this is a pretty bad first impression:

screenshot from 2015-10-06 12-16-11

I think it'd be worth intercepting that particular error and displaying a more friendly message:
----
## Permission Denied

This tool is for administrators to add, remove, and update TaskCluster credentials.  You don't have permission to view credentials.

If you would like to set up temporary credentials (with few permissions), click the "Log In" link.  If you have received a set of permanent credentials, add those in the [preferences](https://tools.taskcluster.net/preferences/) tool.
----

djmitche:
Oh, hah, now I see that you've built new tools to replace that, duh.  In my defense, the change is in the last file in the diff (menu.js)

djmitche:
So yeah, this looks good :)

jonasfj:
Yeah, new tools aren't deployed yet... But that's just hitting the merge button...

They currently operate in load-test server so won't do it now...

PR #32, opened 2015-11-16
Hooks UI

djmitche:
This includes @acmiyaguchi's work, but almost completely rewrites it.

On the left is a browser, using react-treeview to render the group/hook levels.  On the right is either a hook viewer or a hook editor, broken out by field.

I've left bits and pieces for later work -- ship early and often, after all.  The big thorn in my side is that the current schema requires that tasks have some datestamps, but the hooks service overwrites those datestamps anyway.


hooks/hookeditor.jsx
```diff
@@ -0,0 +1,573 @@
...
+        hook:    stripHookIds(hook),
+        editing: false,
+        error:   null
+      });
+    }.bind(this), function(err) {
+      this.setState({
+        error:   err
+      });
+    }.bind(this));
+  },
```
jonasfj: Just a note... `async` works here...

So you can do:
```js
async updateHook(hook) {
  try {
    await this.hooks.updateHooks(...)
    ...
  } catch(err) {
    ...
  }
}
```
(me suspects the code is inherited from long time ago...)

jonasfj:
@djmitche, `r+`.
I see nothing immediately scary here... And I don't really want to review the code in detail :)

I would rather we push it and fix bugs as they are discovered...

PR #33, opened 2015-11-25
Add noVNC support

jonasfj:
Making PR so we can easly back out... no need to review this... It's just the `display/` sub-folder... :)


jonasfj:
@gregarndt, I merged this because I'm an impatient punk... But feel free to give it a review or just rubber stamp it :)

gregarndt:
I'll just give it the rubber stamp since you've already been playing with it and I think showed it off to other people.  I tried it out too and it seemed to work.

PR #35, opened 2015-12-07
Add task images and devices to run locally section

gregarndt:
This is just a quick hack to start getting more things supported in the run locally section.  Ideally this is moved out of here and the worker can upload it's own usage script as we've talked about before, but for now this hack should be good enough.

Definitely highlights the fact that we need to add something to simplify this to our taskcluster cli client.

### DIFF
--- a/lib/ui/taskinfo.jsx
+++ b/lib/ui/taskinfo.jsx
@@ -305,29 +305,87 @@ var TaskInfo = React.createClass({
     //       (obviously we can provide the credentials)
     var taskId  = this.props.status.taskId;
     var payload = this.state.task.payload;
+    var cmds = [];
+    var imagePullCmds= [];
+    var deviceCmds = [];
+
     if (!payload.image) {
       return "# Failed to infer task payload format";
     }
-    var cmds = [];
+
     cmds.push("#!/bin/bash -e");
     cmds.push("# WARNING: this is experimental mileage may vary!");
     cmds.push("");
-    cmds.push("# Fetch docker image");
-    cmds.push("docker pull '" + payload.image + "'");
+
+    if (typeof payload.image === 'string') {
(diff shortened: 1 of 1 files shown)


lib/ui/taskinfo.jsx
```diff
@@ -305,29 +305,87 @@ var TaskInfo = React.createClass({
...
+      imagePullCmds.push("# Image appears to be a task image");
+      imagePullCmds.push("# Download image tarball from task");
+      imagePullCmds.push("curl -L -o image.tar https://queue.taskcluster.net/v1/task/" + imageTaskId + "/artifacts/" + imagePath);
+      imagePullCmds.push("");
+      imagePullCmds.push("# Extract image name and tag from image tarball");
+      imagePullCmds.push("# Note: jq is required.  Download the package appropriate");
+      imagePullCmds.push("# for your OS at https://stedolan.github.io/jq/");
+      imagePullCmds.push("image=$(tar xf image.tar -O repositories | jq -r 'keys[0]')");
+      imagePullCmds.push("image_tag=$(tar xf image.tar -O repositories | jq -r '.[keys[0]] | keys[0]')");
+      imagePullCmds.push("image_name=$image:$image_tag");
```
jonasfj: Not sure how but there must be something smarter than this... can't we overwirtr the file instead... Hmm... Or well...hmm.. Maybe not important now..

Maybe dockercli gives back the image name loaded... Hmm, maybe if we use --quiet or something...

gregarndt: I didn't find a way, and when I asked in #docker on freenode, they said that because an image.tar can contain multiple images, it doesn't return anything about the images it loaded.

I also thought of having the image building task have some task.extra information about the image name, but that would mean the task actually have that, and it's not strictly enforced.

jonasfj:
then let's land this...

PR #36, opened 2015-12-16
Fix for trailing ' in docker pull command

petemoore:
@luser r?


luser:
Looks good. Strings are hard.

PR #38, opened 2016-01-05
Fix react render grouping. r=dustin

walac:
React doesn't allow multiple component rendering, so we group them in a
div element.

r? @djmitche


djmitche:
r+

PR #40, opened 2016-01-06
Update audio devices

gregarndt:
Mistyped the audio device descriptors


jonasfj:
@gregarndt, doesn't this require that you created these devices...?

Maybe leave a comment about that... (or maybe it's already there)

jonasfj:
Ah, you did :)

PR #41, opened 2016-01-06
Remove the re-run button.

rillian:
This requires a separate permission which isn't uniformly granted.
The same thing can be achieved by just resubmitting the task.

### DIFF
--- a/lib/ui/taskinfo.jsx
+++ b/lib/ui/taskinfo.jsx
@@ -118,19 +118,6 @@ var TaskInfo = React.createClass({
             this task may cause your code to land with broken tests.
           </ConfirmAction>&nbsp;
           <ConfirmAction buttonSize="xsmall"
-                         buttonStyle="success"
-                         disabled={!isResolved}
-                         glyph="repeat"
-                         label="Rerun Task"
-                         action={this.rerunTask}
-                         success="Successfully scheduled task rerun!">
-            Are you sure you wish to rerun this task?
-            This will cause a new run of the task to be created. It will only
-            succeed if the task hasn't passed it's deadline. Notice that this
-            may interfere with listeners who only expects this tasks to be
-            resolved once.
-          </ConfirmAction>&nbsp;
-          <ConfirmAction buttonSize="xsmall"
                          buttonStyle="danger"
                          disabled={isResolved}
                          glyph="stop"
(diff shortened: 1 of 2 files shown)


rillian:
Just an attempt at a quick fix. My first contribution. r? @jonasfj

jonasfj:
I'm going to leave this open for now... There are still some people using the button.

Really we should probably just replace it with a button that re-creates the task... without the intermediate trip to the task-creator.

gregarndt:
hrm, I wonder if we could deprecate the button.  When clicking it, a popup will need to be confirmed first and that popup will explain the button will be deprecated in favor of using the task creator. And with information on how to go bug jonasfj to discuss its future.

petemoore:
Since no automated tools should be hitting the rerun button, I think we are safe to change the implementation, as it only affects human interaction. I think we all mostly agree that if a user wants to run a task again, that the workflow that is currently available via the "Edit Task" button is the correct workflow to perform, in order that the new task gets a new taskId. However "Edit Task" is also misleading - so I think we should change these two buttons into one button, and find a better name.

How about, when you hit the new button (name to be decided) it opens up a tree of the entire task graph with the current task ticked, and allows you to select other tasks in the taskgraph with tickboxes, and then you can submit a new taskgraph with new tasks, and everything will get new ids. Otherwise if you want to run a partial task graph again, there is no simple way to do it via the UI, and this hopefully serves two uses cases at once (rerunning an individual task, or rerunning a partial taskgraph).

Or is this crazy?

rillian:
I've never had to launch a task graph, so I don't have an opinion on that. However, it sounds like a differnent step. If there's consensus the re-run button should go, remove the button and file a new issue for adding new ui.

jonasfj:
Sorry, for leave this PR hanging... But I think right now consensus on this is a bit sketchy.. .Sometimes we want it... some times we want the nice task semantics that each task can only be resolved once (which rerunTask breaks).

Let's wait and see what happens when the big graph scheduler is written and people start putting it into production...

Callek:
Fwiw, "rerun" is explicity used by releng, a lot, right now with release promotion. It allows us to keep the entire graph state sane (using a specific taskID for artifacts, that are referenced later) and such.

There may be ways to make our use work in other ways, but when dealing with multiple (5-10) tasks that tend to need to be manually rerun (due to unrelated issues) having to resubmit the entire dep chain each time, and then knowing which items in the dep chain are important and should be run is VERY hard, if theyall have different taskIDs.

eliperelman:
Given that whether this should be done or not is still under contention, I'm going to close this for now. If it's determined that we want to move forward on this, please rebase and reopen. Thanks.

PR #44, opened 2016-01-27
Manipulate new client properties

djmitche:
This adds the ability to edit a client's scopes, factoring out the common code from roles.

It also adds enable/disable buttons, mostly so I could test the functionality.  I'm actually not sure these are a good idea, since that functionality should only be used by identity providers -- do you think I should remove the buttons?


jonasfj:
@djmitche, sorry for ignoring this for so long... it looks solid...

PR #45, opened 2016-01-27
Remove stray '$' in run locally

imbstack:
Hopefully this does what it says on the tin. Running locally it appears to work and `npm run test` is green.

[Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1218747)


jonasfj:
Thanks... this repo auto deploys on merge... (not all our repos does)

PR #47, opened 2016-02-04
Recreate interactive

jonasfj:
this comes ontop of the previous PR I submitted...
Have look...

The bug is here: https://bugzilla.mozilla.org/show_bug.cgi?id=1245384

Maybe you'll want to merge the other PR first... but note that this includes fixes for the other PR, I found two things that didn't work...

### DIFF
--- a/one-click-loaner/connect/connect.jsx
+++ b/one-click-loaner/connect/connect.jsx
@@ -1,0 +1,253 @@
+let React           = require('react');
+let bs              = require('react-bootstrap');
+let utils           = require('../../lib/utils');
+let format          = require('../../lib/format');
+let taskcluster     = require('taskcluster-client');
+let _               = require('lodash');
+let debug           = require('debug')('tools:one-click-loaner');
+
+const taskStateLabel = {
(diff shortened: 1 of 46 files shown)


one-click-loaner/connect/connect.jsx
```diff
@@ -0,0 +1,253 @@
...
+      );
+    }
+
+    if (this.state.shellUrl || this.state.displayUrl) {
+      return (
+        <span>
+          <h1>Loaner Ready!</h1>
+          <i>
+            You have approximately 5 minutes to connect, after that the loaner
+            will shutdown when all connections are closed.
```
imbstack: Does this mean the loaner will go away after 5 minutes or that if the connection isn't made within 5 minutes it will go away?

jonasfj: If connection isn't made... Ie if the user doesn't click the "shell" button that shows up...

imbstack:
Just a tiny nitpick I commented on. Otherwise this looks fine.

PR #48, opened 2016-02-06
Bug 1234827: a working UI for secrets

djmitche:
This uses the new listing functionality to give a more usable UI.

I couldn't figure out why the hashing isn't working.

I also couldn't figure out how to hide the secret until the user clicks a button.  I tried
```
var HiddenJson = React.createClass({
  render() {
    if (this.state.show) {
      return <pre>{JSON.stringify(this.props.jsonValue, null, 2)}</pre>;
    } else {
      return <bs.Button bsStyle="warning" onClick={this.show}>
                Reveal Secret
             </bs.Button>;
    }
  },

  show() {
    this.setState({show: true});
  }
});
```
but it caused crazy DOM errors with replicated React DOM objects, etc.

### DIFF
--- a/secrets/app.less
+++ b/secrets/app.less
@@ -1,1 +1,2 @@
 @import "../lib/format.less";
+@import "secretmanager.less";
--- a/secrets/secreteditor.jsx
+++ b/secrets/secreteditor.jsx
@@ -1,0 +1,255 @@
+var React            = require('react');
+var bs               = require('react-bootstrap');
+var utils            = require('../lib/utils');
+var taskcluster      = require('taskcluster-client');
+var DateTimePicker   = require('react-widgets').DateTimePicker;
+var format           = require('../lib/format');
+var ConfirmAction    = require('../lib/ui/confirmaction');
+var _                = require('lodash');
+
+var SecretEditor = React.createClass({
+  /** Initialize mixins */
+  mixins: [
+    utils.createTaskClusterMixin({
+      clients: {
+        secrets:       taskcluster.Secrets
+      },
+      reloadOnProps: ['currentSecretId']
+    })
+  ],
+
+  propTypes: {
+    // Method to reload a client in the parent
+    reloadSecrets:  React.PropTypes.func.isRequired
+  },
+
+  getDefaultProps() {
+    return {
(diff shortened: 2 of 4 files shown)


djmitche:
@garbas please have a look-see, too.

garbas:
@djmitche i've looked at the pr and also the repository. it would hard to comment on specific lines for this pr since the way setup my react project differently.

 - as first i would include react-redux[1] to formalize how data is passing 
 - i find passing functions as parameters an anti pattern. redux solves this.
 - i like to use immutable.js just to be on the safe side when dealing with actions
 - depending on the react version you might need to explicitly use PureRenderMixin[2]

I understand that current situation of the code is as it is. Not sure if it is worth changing it, you'll be able to judge better. If you are interested I can show you on example how redux makes things saner to maintain.

[1] http://rackt.org/redux/docs/basics/UsageWithReact.html
[2] https://facebook.github.io/react/docs/pure-render-mixin.html

garbas:
@djmitche i was just about to work on the UI part of the clobberer->taskcluster ... if you want i can implement it with react+redux and it can serve as an example which i mentioned above. if not i'm going to do it the quickest way possible.

djmitche:
Thanks!  I think you and @jonasfj should probably talk about the UI implementation details -- I've tried to submit patches, but I don't want to own the framework.

As for clobberer, I would suggest sticking to the framework it's embedded in.  Since it's a RelengAPI blueprint, it should use the Angular framework behind the RelengAPI UI.

jonasfj:
@garbas,
Current setup both the grunt, react, etc. is written by someone who saw react and thought "Wow, this is awesome I got to use this new shiny thing!!!", and then proceeded to hack up something...
I'm very guilty of this crime :)

Architecture-wise, this code is very dirty (we started it before we adopted babel).. Hopefully it can be refactored. I haven't really look too much at redux, but if you have a good idea about how to set this up... and have it all working nicely we're very interested.. You don't have to refactor it all, but I would be curious to see something that includes composition like index-artifact browser..

Also if you have any solution smarter than nasty grunt hack we have... we are all ears, I'm under no illusion that the current setup is great. It's works, but only barely :)

jonasfj:
looks okay to me...

PR #49, opened 2016-02-09
Fix scope validation to use the same regex as tc-auth

djmitche:


jonasfj:
I trust this is correct :)

PR #50, opened 2016-02-16
Add a retrigger-task button

imbstack:
I _think_ this is what [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1244844) is talking about adding, but I'm not really sure either. Figured it would be best to just handle this in a review and iterate.

### DIFF
--- a/lib/ui/retrigger-button.jsx
+++ b/lib/ui/retrigger-button.jsx
@@ -58,7 +58,6 @@ let RetriggerButton = React.createClass({
         <code>taskId</code>.<br/><br/>
         The new task will be altered as to:
         <ul>
-          <li>Strip <code>task.payload.caches</code> to avoid poisoning,</li>
           <li>Strip <code>task.routes</code> to avoid side-effects, and</li>
           <li>Update deadlines and other timestamps for the current time.</li>
         </ul>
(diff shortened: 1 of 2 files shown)


lib/ui/retrigger-button.jsx
```diff
@@ -0,0 +1,94 @@
...
+    let taskId = slugid.nice();
+    let task = _.cloneDeep(this.props.task);
+
+    delete task.routes;
+
+    // Construct message of the day
+    let msg = "\\nRetriggered task based on taskId: " +
+              this.props.taskId + "\\n"
+
+    delete task.payload.caches;
```
jonasfj: Hmm... if it's a retrigger, I think it's safe to keep caches...

lib/ui/retrigger-button.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+    task.deadline = taskcluster.fromNowJSON('12 hours');
+    task.created = taskcluster.fromNowJSON();
+    task.expires = taskcluster.fromNowJSON('7 days');
```
jonasfj: We should change this to use the orignal time stamps and just offset them as they where relative to ´task.created`

lib/ui/retrigger-button.jsx
```diff
@@ -0,0 +1,93 @@
...
+  },
+
+  async createTask() {
+    let taskId = slugid.nice();
+    let task = _.cloneDeep(this.props.task);
+
+    delete task.routes;
+
+    // Construct message of the day
+    let msg = "\\nRetriggered task based on taskId: " +
```
jonasfj: No need for the message :)

lib/ui/retrigger-button.jsx
```diff
@@ -0,0 +1,93 @@
+let React           = require('react');
+let taskcluster     = require('taskcluster-client');
+let ConfirmAction   = require('./confirmaction');
+let utils           = require('../utils');
+let slugid          = require('slugid');
+let _               = require('lodash');
+let shellescape     = require('shell-escape');
```
jonasfj: No need for this package

jonasfj:
`r+`

PR #51, opened 2016-02-16
Set environment variable for loaners

jonasfj:


jonasfj:
In talks on how to make loaners awesome:
https://public.etherpad-mozilla.org/p/jonasfj-lets-make-loaners-awesome

We concluded that we should set an environment variable on loaner tasks... and let in-tree default command take care of setting up an environment, etc... without running the actual tests...

lib/ui/loaner-button.jsx
```diff
@@ -81,15 +81,18 @@ let LoanerButton = React.createClass({
               this.props.taskId + "\\n" +
               "Original command was: " + shellescape(task.payload.command);
 
+    task.payload.env = task.payload.env || {};
+    task.payload.env.TASKCLUSTER_INTERACTIVE = 'true';
+
     // Construct new task command
-    task.payload.command = [
+    /*task.payload.command = [
```
imbstack: Why comment this out and not just delete it?

jonasfj: good point...

jonasfj: fixed...

imbstack:
r+

PR #52, opened 2016-02-17
Allow for custom shells

jonasfj:


jonasfj:
see bug https://bugzilla.mozilla.org/show_bug.cgi?id=1248991

shell/app.jsx
```diff
@@ -34,6 +34,7 @@ $(function () {
           'if [ -z `which "$SHELL"` ]; then export SHELL="/.taskclusterutils/busybox sh"; fi;',
           'SPAWN="$SHELL";',
           'if [ "$SHELL" = "bash" ]; then SPAWN="bash -li"; fi;',
+          'if [ -f "/etc/taskcluster-interactive-shell.sh" ]; then SPAWN="/etc/taskcluster-interactive-shell.sh"; fi;',
```
djmitche: Will this need `/.taskclusterutils/busybox sh` in its `#!` line?

jonasfj: nope...

it can use whatever... The only reason we have "/.taskclusterutils/busybox sh" is because we can't be sure that there is anything inside the container... so if all else fails... we can always spawn that...

As `/etc/taskcluster-interactive-shell.sh` exists in the image, it can safely assume that any other binary inside the image exists.

jonasfj: Note: unrelated bug, the first `'sh', '-c', [` should probably be:
`'/.taskclusterutils/busybox', 'sh', '-c', [` because I can't use `sh` for shell detection, as it may not even be inside the image...

Practically, this is no problem... maybe for a go binary packaged as an image without shell utils... but everybody packages with a shell typically..

djmitche: I expect we will find task images that have no shell -- it's one of the stages of learning Docker to realize that you don't always need a full OS image :)

But, given your explanation, this looks good - ship it :)

PR #54, opened 2016-02-18
Bug 1249464: fix scope editor to delete the correct scope

djmitche:


djmitche:
(the problem is that the scopes are rendered in sorted order, but the rendering function does not modify the property, which thus remains unsorted; so passing an index to `removeScope` was meaningless)

djmitche:
@jonasfj - this did not deploy because the Travis job failed.  Do you know why that didn't report here?

jonasfj:
hmm... I think @imbstack disabled build on PR... but building `taskcluster-tools` on PR is sane... as it doesn't have credentials... We were probably a bit to fast in disabling it.. .anyways, no harm...

@djmitche, just fyi, you are pretty much the only one pushing to personal repos.
For tc-tools, it's cool. For the other repos running tests fails because they often requires some secrets and we don't expose to PRs. So we're increasingly disabling build on PR, and just pushing to a branch on the `taskcluster/...` repo.

jonasfj:
I've reenabled build on PR.

djmitche:
I know, I'm representing the outside contributor here.  When we get rid of required credentials for tests, we'll get decent results for pull requests.

jonasfj:
tools don't have any secrets... so that's fine...

But things that integrate with S3 and azure queue storage is a bit of work to mock... but we are slowly moving tin the right direction... you recently added inMemory support.

PR #56, opened 2016-02-23
"Purge workers cache" button on TaskInfo component

garbas:
This is still WIP but I would need some feedback on the code since this is my first contribution.

Also any advise how I would test this would be more then welcome. Problem with testing is that I need to login with this credentials -> https://tools.taskcluster.net/auth/clients/#garbas

I went as far as manually creating ``credentials`` record in localStorage, but without luck (for now).


djmitche:
You can add a set of credentials by copy/paste with the ill-named "Preferences" tool.

lib/ui/taskinfo.jsx
```diff
@@ -33,16 +34,25 @@ var TaskInfo = React.createClass({
...
 
   /** Load task definition */
   load() {
+    var self = this;
+    var task = self.queue.task(self.props.status.taskId);
     return {
-      task:         this.queue.task(this.props.status.taskId)
+      task: task,
+      selectedCacheNames: new Promise(function(resolve, reject) {
+        var resolved = Promise.resolve(task).then(
```
djmitche: If I understand correctly, this is forcing a wait for the resolution of the `task` promise.  I don't think that's necessary -- I believe `load` resolves any promises for you.  So you could use

```js
selectedCacheNames: task.then((t) => _.keys(self.getCacheNames(t)))
```

lib/ui/taskinfo.jsx
```diff
@@ -135,6 +145,15 @@ var TaskInfo = React.createClass({
...
             scheduling process will not be able to schedule the task.
           </ConfirmAction>&nbsp;
+          <ConfirmAction buttonSize="xsmall"
+                         buttonStyle="danger"
+                         disabled={this.getCacheNames(task) === undefined}
+                         glyph="trash"
+                         label="Purge worker cache"
+                         action={this.purgeWorkerCache}
+                         success="Cache successfully purged!">
+            {this.state.purgedCacheNames === undefined ? this.listCacheNames(task) : undefined}
```
djmitche: This will need some more text!

lib/ui/taskinfo.jsx
```diff
@@ -290,6 +309,53 @@ var TaskInfo = React.createClass({
...
 
+  getCacheNames(task) {
+    return (((task || {}).payload) || {}).cache;
+  },
+  listCacheNames(task) {
+    var updateSelectedCacheNames  = this.updateSelectedCacheNames;
+    var selectedCacheNames        = this.state.selectedCacheNames;
+    return (
+      <div>
+        <span>Are you sure you wish to purge cache of this task?</span>
```
djmitche: "Are you sure you wish to purge caches used in this task across all workers?"

djmitche:
I think this is the right design -- @jonasfj?

lib/ui/taskinfo.jsx
garbas: ah right ... need to remind myself i'm in javascript context :)

garbas:
@djmitche i used ``garbas`` clientId you gave me and i get 
2016-02-23-190814_638x701_scrot

should i try to find another task to test or i need bigger scope for my clientId?

lib/ui/taskinfo.jsx
```diff
@@ -290,6 +305,58 @@ var TaskInfo = React.createClass({
...
+                                  checked={this.selectedCacheNames !== undefined
+                                      ? this.selectedCacheNames.indexOf(cacheName) !== -1
+                                      : false}
+                                  value={cacheName}
+                                  />
+                           {cacheName}
+                         </label>
+                      </li>;
+            })}
+        </ul>
```
jonasfj: I think it's better to make sub-component called `PurgeCacheButton` with the property `cacheNames`...

lib/ui/taskinfo.jsx
```diff
@@ -135,6 +141,15 @@ var TaskInfo = React.createClass({
...
           </ConfirmAction>&nbsp;
+          <ConfirmAction buttonSize="xsmall"
+                         buttonStyle="danger"
+                         disabled={this.getCacheNames(task) === undefined}
+                         glyph="trash"
+                         label="Purge worker cache"
+                         action={this.purgeWorkerCache}
+                         success="Cache successfully purged!">
+            { this.listCacheNames(task) }
+          </ConfirmAction>&nbsp;
```
jonasfj: Ie. replace this with `<PurgeCacheButton caches={_.keys(((task || {}).payload || {}).cache || {})}/>`

That way we can hide all the selection state and logic in a sub-component... I don't see any immediate places where we would reuse it... But who knows... Maybe we will want to one day...

It's certainly nice to keep it out of the taskinfo component... Like I did with `LoanerButton`...

Basically, just have `PurgeCacheButton` use `ConfirmAction` as its root element... It gets a lot less ugly that way...

lib/ui/taskinfo.jsx
```diff
@@ -33,16 +34,21 @@ var TaskInfo = React.createClass({
   getInitialState() {
     return {
       // task definition
-      taskLoaded:   false,
-      taskError:    undefined,
-      task:         undefined
+      taskLoaded:         false,
+      taskError:          undefined,
+      task:               undefined,
+      purgedCacheNames:   undefined,
```
jonasfj: This state isn't necessary... Anyways, it'll be much easier if we make a subcomponent...

lib/ui/taskinfo.jsx
jonasfj: I could imagine a future where we support inspecting a worker... In such a case we might find another use for a PurgeCacheButton... that takes a list of caches as the only property...
Just saying it's definitely something that could be reused.

lib/ui/taskinfo.jsx
```diff
@@ -290,6 +305,58 @@ var TaskInfo = React.createClass({
...
+      cacheNames.push(e.target.value);
+    } else if (e.target.checked === false) {
+      cacheNames = cacheNames.filter(i => { return i !== e.target.value});
+    }
+    this.setState({ selectedCacheNames: cacheNames });
+  },
+
+  async purgeWorkerCache() {
+    let result = [];
+    for (let cacheName of this.state.selectedCacheNames) {
```
jonasfj: The cool promise pattern is:
```js
return Promise.all(this.state.selectedCacheNames.map(cacheName => {
  return this.purgeCache.purgeCache(this.state.task.provisionerId, this.state.task.workerType, {cacheName});
}));
```

lib/ui/taskinfo.jsx
jonasfj: Okay, you would need provisionerId and workerType as properties too...

djmitche:
It looks like Jonas fixed your permissions issue?

jonasfj:
Yeah, I fixed the scope issue...

IMO can we just give purge-cache:* to releng, relops etc... So we don't have to use perma creds.

djmitche:
Hold off on that until we [namespace the workerTypes](https://bugzilla.mozilla.org/show_bug.cgi?id=1220686), and we can grant more specific scopes

garbas:
@djmitche permission was fixed and i could tested that purgeCache worked.
@jonasfj i fixed code based on your comments

one thing i discovered (as i was testing) is that once you open ConfirmAction dialog you need to reload the page to reuse the same button. i fixed this by resetting initial state when result from action comes back [967b53f](https://github.com/garbas/taskcluster-tools/commit/967b53f47bb51b7fe2520d96e0324f3dbcd6d4ee)

jonasfj:
`r+` shipping...

PR #58, opened 2016-02-24
Fixed bug...

jonasfj:
merge and it'll deploy in like 10min


gregarndt:
:+1:

garbas:
just out of curiosity is this is babel bug? or i've used "async () => { ... }" wrong?

jonasfj:
@garbas,
No Babel bug... The async... Thing you wrote is a function that returns a promise.... A function does nothing unless called. Which is all I did :)

jonasfj:
LoadState is a hack I've done in the mixin that allows to get state from a promise object...

PR #59, opened 2016-02-26
Bug 1245254 - Improve run locally message for developers

armenzg:
The current output of 'run locally' only works if you use vagrant and
we're hoping to not need to do so by using this changes.

It also helps because it does not immediately start executing the entry
script but allows for the developer to initiate which takes aways the
"execute and abort" problem.

There is like to be more changes related to this but this is a start.

@jonasfj r?

Is there an easy way for me to test this kind of changes before asking for review?
Or even run some unit tests? Thanks!


lib/ui/taskinfo.jsx
```diff
@@ -352,13 +352,22 @@ var TaskInfo = React.createClass({
...
       cmds.push("# Consult the vagrant.sh file in the docker-worker repo ");
       cmds.push("# for more information on how to install and configure ");
       cmds.push("# the loopback devices. http://www.github.com/taskcluster/docker-worker");
+      cmds.push("");
       cmds.push("# Warning: This is entirely dependent on local setup and ");
       cmds.push("# availability of devices.");
+      cmds.push("");
       if (payload.capabilities.devices['loopbackVideo']) {
-        deviceCmds.push("  --device /dev/video0:/dev/video0 \\");
+	cmds.push("# This job requires access to your video device.");
```
gregarndt: Perhaps we could get rid of the comments starting on line 352 in favor of these.

lib/ui/taskinfo.jsx
```diff
@@ -352,13 +352,22 @@ var TaskInfo = React.createClass({
...
+	cmds.push("# If this is the first time you're trying to run this locally,");
+	cmds.push("# make sure to run this first:");
+	cmds.push("# sudo apt-get install v4l2loopback-dkms");
+	cmds.push("# sudo modprobe v4l2loopback # This will create a device under /dev/video*");
+        cmds.push("last_device=`ls /dev/video* | tail -n 1`");
+        deviceCmds.push("  --device $last_device:$last_device \\");
       }
       if (payload.capabilities.devices['loopbackAudio']) {
```
gregarndt: Do we not need to mention anything like we do for video loopback?

gregarndt:
I believe that you can run "grunt" and it will start up a local server that you can point your browser at.  From there you can load up the task inspector to view an existing task and see what the run locally looks like.

jonasfj:
`grunt develop` will run it locally... All feature should work equally well on localhost  :)

Generally, this looks good...

jonasfj:
@gregarndt, you know more about how to run locally. If you like this feel free to merge...

This repo will autodeploy... but if travis is green it should at-least build. As for whether or not it crashes in browser we just don't know :)

lib/ui/taskinfo.jsx
```diff
@@ -382,9 +391,19 @@ var TaskInfo = React.createClass({
     _.keys(payload.env || {}).forEach(key => {
       cmds.push("  -e " + key + "='" + payload.env[key] + "' \\");
     });
+
+    // This allows changing behaviour on the script to be run to change
+    // behaviour from production (e.g. start VNC)
+    cmds.push("  -e RUN_LOCALLY='true' \\")
```
jonasfj: This is nice! :)

lib/ui/taskinfo.jsx
armenzg: I will look into re-adjusting.

lib/ui/taskinfo.jsx
armenzg: I can add it, however, I had ignored it because I've disabled loopbackAudio from most jobs (it seems it might have been necessary at some point).

https://bugzilla.mozilla.org/show_bug.cgi?id=1251702

gregarndt: Yea I can see how this line would be unnecessary for the tasks that no longer need loopback audio, but while loopbackAudio is an option that a task could define, we should make the documentation and behavior consistent.

lib/ui/taskinfo.jsx
gregarndt: Thank you sir!

armenzg:
I get this:
```
armenzg@armenzg-thinkpad:~/repos/taskcluster/taskcluster-tools$ grunt
Loading "Gruntfile.js" tasks...ERROR
>> Error: Cannot find module 'babelify'
Warning: Task "default" not found. Use --force to continue.

Aborted due to warnings.
```
```
armenzg@armenzg-thinkpad:~/repos/taskcluster/taskcluster-tools$ grunt develop
Loading "Gruntfile.js" tasks...ERROR
>> Error: Cannot find module 'babelify'
Warning: Task "develop" not found. Use --force to continue.

Aborted due to warnings.
```

armenzg:
r? @gregarndt 

I don't know if you prefer squashed commits or not (since it clears comments you've made in the past).

gregarndt:
The commits are fine.  Does this run locally for you now? Are you able to verify the run locally output?

armenzg:
@gregarndt after the last commit it works.

However, I can't see anything under this:
http://localhost:9000/task-inspector/

I only see this:
http://people.mozilla.org/~armenzg/sattap/eb8492cd.png

gregarndt:
I'm not exactly sure why you're having an issue but I was able to run it locally and couldn't see any noticeable problems.  I think this is safe to merge.

PR #60, opened 2016-03-02
Bug #1252981: Move tools.tc and github refs to HTTPS

claudijd:
Note: I don't fully understand the potential impacts of this.  My hope is that this is mostly meta-data for users, but consider this a fair warning to make sure this isn't going to break something.


jonasfj:
yeah, I think so.... Thanks :)

PR #61, opened 2016-03-03
Login v3 UI (part 1)

djmitche:
There's more to do here, but this is a nice point for review.

 * Use the phrases "Sign In" and "Sign Out" everywhere
 * Make a credentials menu, displaying the active clientId
 * Move preferences to a "Manage Credentials" page, put it on the credentials menu
 * Improve "Manage Credentials" page to display actual scopes, expiry, etc.
 * Don't reload on credential change, by default
 * Show a popover when signing in or out
 * Filter clientIds by prefix, defaulting to the credentials' clientId
 * Default new clients to begin with the credentials' clientId
 * Minor cleanup for client/role manager behaviors


djmitche:
OK, I added a bunch more commits that I had intended to make "part 2" here.  So this basically finishes off the tools UI for this project.

djmitche:
Note: I'm not planning to support multiple credentials in this iteration.  I don't see anyone but a few people using them, and they lead to a lot of complexity for all users.

auth/clients/clientmanager.jsx
```diff
@@ -24,7 +26,9 @@ var ClientManager = React.createClass({
 
   /** Create an initial state */
   getInitialState() {
+    let clientId = auth.loadCredentials().clientId;
```
jonasfj: Perhaps this would have been better off as a property.

Then we would have to clientmanagers, one for "Edit my clients", and the global client manager..

On the other hand I can see it as super-useful to have a prefix... in case I want to only look at `permanent/` clients...

So let's try this first :)

build-files.js
```diff
@@ -29,8 +29,8 @@ module.exports = [
   'login/app.jsx',
 
   // Preferences
-  'preferences/index.jade',
-  'preferences/app.jsx',
+  'credentials/index.jade',
+  'credentials/app.jsx',
```
jonasfj: Much better name :)

jonasfj:
This looks great... `r+` land it once you've finished merge issues...

auth/clients/clientmanager.jsx
jonasfj: We'll see if we get tired of clearing the `prefix` editor...

djmitche: In my playing around with this, I found it kind of handy.  For example, I can look at `permanent/releng`.  I'm tempted to add a similar thing to the roles editor (although OMG this all needs some better refactoring..)

jonasfj: Just remember for roles that you can do the filtering clientside..
So no need to update the server...

PR #63, opened 2016-03-03
Bug 1247008: replace lame DatePicker with TimeInput

djmitche:
All of the datepickers either don't support times, or are ugly, or both.

This replaces it with a simple text-entry box.


jonasfj:
Looks sane to me... I guess you want more accuracy for secrets...

PR #64, opened 2016-03-04
Ignore .swp files

armenzg:
.swp files are temporarily generate by Vim when editing a file.
We should be listing these files under git status.


armenzg:
Ready to be merged! @jonasfj Thanks!

PR #66, opened 2016-03-04
bug 1241176:Be less red

techchic:
make task_inspector <code> less "DANGER" colored.

### DIFF
--- a/task-inspector/taskinspector.less
+++ b/task-inspector/taskinspector.less
@@ -1,1 +1,5 @@
 @import "../lib/ui/taskview.less";
+code {
+	color: #2C3638 ;
+	background-color: #FFF ;
+}
--- a/task-inspector/taskinspector.less
+++ b/task-inspector/taskinspector.less
@@ -1,5 +1,2 @@
 @import "../lib/ui/taskview.less";
-code {
-	color: #2C3638 ;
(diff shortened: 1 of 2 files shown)


jonasfj:
I think there is a `format.less` that imported everywhere through layout.jsx, maybe that's is a better place to do this...

Maybe we could keep the background grey and just make the text black so it's still distinguishable from normal text.

djmitche:
Agreed.  I think this is the right CSS, but it needs to apply to all pages in the site.

jonasfj:
Assign @djmitche to this as I shouldn't be arguing color scheme :)

task-inspector/taskinspector.less
```diff
@@ -1 +1,5 @@
 @import "../lib/ui/taskview.less";
+code {
+	color: #2C3638 ;
+	background-color: #FFF ;
```
walac: nit: coding format should be 2 spaces.

techchic:
I have transferred the code to format.less

djmitche:
You also omitted the `background-color: #FFF`, but I think that's what @jonasfj wanted, and I like the look overall:

screenshot from 2016-03-07 17-36-33

djmitche:
Thanks @techchic!

PR #67, opened 2016-03-08
Allow pasting several scopes into the scope editor

djmitche:
This is a convenience when, for example, setting up new projects.

I could potentially add a tooltip for this, but I don't know if that's more distracting than the feature is worth - what do you think?

### DIFF
--- a/auth/clients/clienteditor.jsx
+++ b/auth/clients/clienteditor.jsx
@@ -236,6 +236,7 @@ var ClientEditor = React.createClass({
                 scopesUpdated={this.scopesUpdated} />
             </div>
           </div>
+          {(!isEditing && !isCreating) ? (
           <div className="form-group">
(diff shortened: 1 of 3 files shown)


jonasfj:
r+

Although maybe we should hide it behind a check box... Because this prevents us from making scopes with whitespace which is allowed...

Otherwise create feature :)

djmitche:
Ack, you're right -- this should just split on newlines.

djmitche:
OK, updated, and a little bit of extra stuff added

djmitche:
Duh, that's not right either.  Sorry, I forgot to test before pushing!

djmitche:
OK, I reworked it to use a textarea instead.  And actually tested it this time.

jonasfj:
I suspect we would better off using codemirror or ace without line numbers than textarea...
But while this isn't as pretty as what we had before, I think this improves productivity :)

PR #69, opened 2016-03-11
warn when credentials are about to expire

djmitche:
This is a little kinder than the existing sign-out-after-expiration behavior (which still exists, but now signs out when the creds actually expire).

This only works for temporary credentials, but I think that's OK, as that's the common case, and most permacreds have long lifetimes.


jonasfj:
Interesting... let's land it...

PR #70, opened 2016-03-12
default to the 'tutorial' workerTypes

djmitche:


jonasfj:
Nice...

So just FYI, downside of this is that there is never a "tutorial" worker type ready, which gives people a bad experience as the task isn't running immediately..

jonasfj:
r+

If you want it... I see both ups and downs to doing this...

djmitche:
I think we can add a sentence to the tutorial about that.

djmitche:
I added a note to docs

PR #71, opened 2016-03-13
Bug 1227532:added a different favicon fo reach taskcluster tool

elena-posea:
https://bugzilla.mozilla.org/show_bug.cgi?id=1227532


jonasfj:
@elena-posea, this is nice...


But it's going to be a pain to maintain this... I assume you generated these automatically?
Can we extend the `gruntfile` with a quick hack that compiles icons for everything listed in `menu.js`, then we should be able to do this as a part of the `layout.jsx` file...

From grunt docs it seems it's just something like:
```js
grunt.registerTask('render-favicons', function() {
  let menu = require("./menu");
  _.uniq(menu.map(entry => entry.icon)).forEach(icon => {
    // Some script to render:
    // lib/assets/font-awesome-favicons/<icon>.ico
    // Perhaps this is as simple as using child_process with some magic
  });
});
```

Or perhaps we can do even simpler with a simple hack like this:
https://gist.github.com/chadsmith/530f8dfa1af3640eff2b

jonasfj:
Btw, make sure to ping someone if we forget about the review...
Or file the link to the PR as a patch on the bugzilla bug and then r? someone specific, then they'll usually get around to it to move it someone else :)

elena-posea:
yes, I downloaded them from here: http://paulferrett.com/fontawesome-favicon/ . I understand what you're saying, I'll look into it today

elena-posea:
I tried the FaviconAwesome solution, but that didn't work, because it turns out trying to take the content from a font-awesome printed in a container doesn't really return the base64 encoding for that favicon. I think it will take a little longer for me to find another solution of taking the favicon as base64 string

jonasfj:
Hmm, I'm actually not sure how to do this... There has got to be some decent shell command thing we can do.. Maybe we should take a look at the svg sources...

jonasfj:
Look at icons.yml and the svg file.. I suspect it's relatively easy to do a hack with imagemagick or inkscape (from cmd) to get what we want...

elena-posea:
Ok, to I fixed some problems with that script, but it still doesn't work/ the canvas somehow doesn't use the correct font to write my hexa code / the glyph code. So I gave up that idea.
I'm having some problems with imagemagick on windows.

What do you think about this: https://github.com/encharm/Font-Awesome-SVG-PNG
I know the last commit in that repo was 2 months ago, but the author accepts pull requests just in case something changes about font-awesome and it needs to be updated. For now, it already contains all we need, and I can pass flags to only create the png for the icons I choose (the ones from the menu), and customize the size and color.

elena-posea:
Ok, so now this process of generating the png icon files happens automatically. All you need is to ensure that on the machine running this, you have installed this:
```
For Debian rsvg-convert in located in the librsvg2-bin package.
sudo apt-get install librsvg2-bin
```
as described on [their page](https://github.com/encharm/Font-Awesome-SVG-PNG)

jonasfj:
@elena-posea, this is really nice... :)
Since we are checking the icons into the git, perhaps we can make this an optional target.
Or perhaps we check if the file exists and create it, if it doesn't...

That way it won't be as painful for windows users either, as they'll only be asked to install the dependency if it's missing...

A final trick would be to ensure that we print a decent error message, if there is an error running the png generation... Something that tells people what package to install on windows, linux and mac.

We should strive to keep it easy to work on this repo... I know the grunt script is far from ideal and very slow... but it does work fairly reliably (though very very slow)...

djmitche:
Would it make sense to leave those to a follow-up pull request, if they're difficult?

jonasfj:
sure, let's land it and fix these minor issues in a follow-up...

PR #72, opened 2016-03-14
click on a row in the task graph

techchic:
The rows underline on hover, like a link, so users know they can click.

### DIFF
--- a/task-graph-inspector/taskgraphinspector.less
+++ b/task-graph-inspector/taskgraphinspector.less
@@ -1,6 +1,16 @@
 @import "../lib/ui/taskview.less";
 
 .task-graph-inspector-tasks > tbody > tr:not(.info) {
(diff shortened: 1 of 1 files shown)


jonasfj:
Seems cool to me... but maybe overkill as cursor should be set to indicate that you can click...

task-graph-inspector/taskgraphinspector.less
```diff
@@ -1,6 +1,16 @@
...
-  cursor:         pointer;
+  cursor: pointer;
+
+}
+.table > tbody > tr > td > span.markdown-view > p:last-of-type {
+  margin: 0px;
+  
+  &:hover {
+    text-decoration: underline;
+	}
```
djmitche: careful with whitespace -- it's harmless here, but can be confusing for the reader

PR #73, opened 2016-03-18
[WIP] Logviewer overhaul

ZetsubouClown:


lib/ui/terminalview.jsx
```diff
@@ -1,7 +1,105 @@
...
 
+var Viewer = React.createClass({
+    getDefaultProps: function(){
+        return {
+            cols: 120,
+            rows: 20,
+        };
+    },
+
+    getInitialState: function(){
```
jonasfj: Small thing, but in ES6 you can do:
```js
var Viewer = React.createClass({
  getDefaultProps(){
      return {
          cols: 120,
          rows: 20,
      };
  },
  ...
});
```

No need for the `function` key words.. it's pretty neat, and makes react.js feel a lot more like classes :)

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function abort(){
+    request.removeEventListener('progress', onData);
+    request.removeEventListener('load', onData);
+    request.abort();
+    close();
+}
+
+onmessage = function(e){
```
jonasfj: Could we not use: `self.onmessage` or `self.addEventListener`..

This goes for `postMessage` and `close` as well, it's much easier to see what is related to the webworker magic...
And what is internal functions...

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,57 @@
...
+        var temp = request.responseText.split('\n');
+        var lines = [];
+        temp.forEach(function(line){
+            for(var i = 0; i < line.length; i += cols)
+                lines.push(line.substr(i, cols));
+        });
+        resp.data = lines;
+      }
+    }
+  postMessage(resp);
```
jonasfj: Hmm... Architecturally, wouldn't it make more sense to send more than one message... as we are reading the lines.
That'll definitely be necessary for this to work on a livelog... as it's just an HTTP request being streamed slowly (as the log is written)...

Perhaps we should send messages as follows:
 * `self.postMessage({id: "uuid", lines: [...], eof: true || false})`

So maybe we send lines every say 100ms, and when we reach of request, we set `eof: true` and stop sending...
Obviously, we should send `lines: []` if we didn't receive any lines for say 100ms.

I think `setInterval` will do nicely for sending this way... I'm not sure 100ms is sane, but that's something we can adjust...

jonasfj:
just some preliminary feedback.. This is rather impressive... I'll have to a look bit more into depth before I have more feedback...

jonasfj:
oh, very sorry. I never saw the new changes.. Completely forgot to open an look :)

andreadelrio:
I was having some trouble getting this to work in OSX. 

Running
`npm install --save bufferutil`
`npm install --save utf-8-validate`

And then trying again fixed it.

eliperelman:
@andreadelrio I had the exact same problem when running tc-tools. I think there is a problem with the way some dependencies/peerDependencies are declared somewhere, but I haven't looked into what the problem is.

walac:
The problem is that `ws` ships with `bufferutil` and `utf-8-validate`, but they fail to build on Mac OSX with node 4. For some reason, the code path that `require` those packages were not exercised before this patch.

eliperelman:
And it looks like browserify used ws which needed to be updated accordingly. Tools is currently using a custom version of grunt-browserify.

@jonasfj do we need this custom grunt-browserify any more? It looks like you forked it so you could update the browserify version, but the upstream seems to be using a much more recent version of browserify which I think may resolve this issue.

jonasfj:
@eliperelman, not sure, feel free to try... let's keep this PR about the PR :)

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,59 @@
...
+        cols = e.data.cols;
+        request = new XMLHttpRequest();
+        request.open('get', e.data.url, true);
+        request.addEventListener('progress', onData);
+        request.addEventListener('load', onData);
+        request.send();
+    };
+    if(e.data.abort) abort();
+});
+}.toString().split('\n').slice(1,-1).join('\n');
```
andreadelrio: Not clear why first line is being removed here. 
Also, it would be more efficient to use a combination of indexOf('\n\) and slice.

ZetsubouClown: It's removed because it shouldn't be there. The first and last lines are
`function(){`
and
`}`
This is a little dirty hack to get the source code as a string while allowing it to be code instead of just a string. Of course, it's quite terrible and should be changed (with webworkify or something similar) once it becomes a bigger concern.

walac: This remove some browserify inserted code. Without it, `require`, for example, won't work. If running this module as a `Worker` is necessary, better to instantiate the worker like this
```javascript
logFetcher = new Worker('log-fetcher.js');
```

Getting the right patch to log-fetcher.js might be a tricky, however.

@eliperelman any idea of the best approach?

eliperelman: I definitely prefer the path approach over the stringification approach. Boils down to `eval`-like semantics.

jonasfj: without the stringify you don't get the babel translation...
But yes, we should be using webworkerify... This is a packaging detail..

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,59 @@
+var source = function(){
+var request;
+var dataOffset = 0;
+var cols;
+
+function divide(data){
+    var lines = [];
```
andreadelrio: nit: use 2 spaces an indentation in all files.

andreadelrio:
Please see PR #103 which continues work started in this PR.

PR #75, opened 2016-03-19
Bug 1256740 - Show accessTokens for new clients in a modal

andreadelrio:
Show "token reset" inside modal when editing Clients.

### DIFF
--- a/auth/clients/clienteditor.jsx
+++ b/auth/clients/clienteditor.jsx
@@ -173,9 +173,10 @@ var ClientEditor = React.createClass({
                       this.state.accessToken !== null ? (
                         <bs.Modal show={this.state.showDialog} onHide={this.closeDialog}>
                           <bs.Modal.Header closeButton>
-                            Token Reset 
+                            Access Token 
                           </bs.Modal.Header>
                           <bs.Modal.Body>
+                            <span>The access token for this clientId is:</span>
                             <code>{this.state.accessToken}</code>
                           </bs.Modal.Body>
                         </bs.Modal>
@@ -410,11 +411,6 @@ var ClientEditor = React.createClass({
     }
   },
 
-  /** dismiss the visible access key */
-  dismissAccessKey() {
-    this.setState({accessToken: null});
-  },
-
   /** Close modal */
   closeDialog() {
     this.setState({accessToken: null});
@@ -439,6 +435,7 @@ var ClientEditor = React.createClass({
         client:         client,
         accessToken:    client.accessToken,
         editing:        false,
+        showDialog:     true,
         working:        false,
         error:          null
       });
--- a/auth/clients/clienteditor.jsx
+++ b/auth/clients/clienteditor.jsx
@@ -157,33 +157,33 @@ var ClientEditor = React.createClass({
           }
           {
             (isEditing && !isCreating) || this.state.accessToken !== null ? (
-              <div className="form-group">
-                <label className="control-label col-md-3">AccessToken</label>
-                <div className="col-md-9">
-                  {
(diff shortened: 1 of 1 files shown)


auth/clients/clienteditor.jsx
```diff
@@ -170,11 +171,14 @@ var ClientEditor = React.createClass({
...
                     : (
                       this.state.accessToken !== null ? (
-                        <bs.Alert bsStyle="warning"
-                               onDismiss={this.dismissAccessKey}
-                               dismissAfter={5 * 60 * 1000}>
-                          <code>{this.state.accessToken}</code>
-                        </bs.Alert>
+                        <bs.Modal show={this.state.showDialog} onHide={this.closeDialog}>
+                          <bs.Modal.Header closeButton>
+                            Token Reset
```
djmitche: This should probably be "Access Token", since it appears both when the client is created and when the token is reset.

auth/clients/clienteditor.jsx
```diff
@@ -410,6 +415,11 @@ var ClientEditor = React.createClass({
     this.setState({accessToken: null});
   },
 
+  /** Close modal */
+  closeDialog() {
+    this.setState({accessToken: null});
+  },
+
```
djmitche: This function replaces dismissAccessKey, above, which is no longer required.

djmitche: Also, it won't close the dialog, since it doesn't change `this.state.showDialog`.

auth/clients/clienteditor.jsx
```diff
@@ -395,6 +399,7 @@ var ClientEditor = React.createClass({
         client:         client,
         accessToken:    client.accessToken,
         editing:        false,
+        showDialog:     true,
```
djmitche: You'll also need to add this to createClient, below.

djmitche:
A few other notes:

 * in the client display (when not editing), please remove the "Access Token" row
 * Add a little extra text to the body of the modal, maybe something like "The access token for clientId <clientId> is <accessToken>"

By the way, if you'd like to be able to login via the "sign-in" link on your development version of tools, you can use a tool called "ngrok".  I set up a (free) account at ngrok.io for myself, and then I run `grunt --force` in one window (which takes a long time to start up) and `ngrok http 9000` in another window.  The second command gives an https URL that you can use to view your site.  Login.taskcluster.net will only offer credentials to an https site.

auth/clients/clienteditor.jsx
andreadelrio: It, however, stops showing the modal because when accessToken is equal to null the condition (this.state.accessToken !== null)  no longer passes. Would this by a valid way to dismiss the modal?

djmitche: Oh, you're right -- I missed that.  Do we need both show={..} and the accessToken conditional?

andreadelrio: It turns out we didn't. There was some repetition going on. I got rid of showDialog. And now the modal works for for edit and createClient.

andreadelrio:
@djmitche is the row I should remove the one right below the modal? See attached:
screen shot 2016-03-21 at 12 15 36 pm

djmitche:
Yes -- at this point, we'll never put an actual accessToken there, so there's no need for it.  Thanks for checking :)

andreadelrio:
@djmitche I think I’ve covered all your comments. Please let me know if there’s anything else I can do.

djmitche:
Yep, looks great -- nice work!

djmitche:
That deploys automatically, but it takes a while (20 minutes or so).  So you should see your changes live soon!

PR #80, opened 2016-03-23
fixed some browser errors caused by prop show in Modal

andreadelrio:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1256740


djmitche:
Awesome, thanks!

PR #82, opened 2016-03-24
Add first sketch for new status page

walac:
This is a WIP for the new status page. Currently, it polls only a few services.


djmitche:
Looks good so far :)

build-files.js
```diff
@@ -98,6 +98,12 @@ module.exports = [
   'one-click-loaner/connect/index.jade',
   'one-click-loaner/connect/app.jsx',
   'one-click-loaner/connect/connect.less',
+
+  // Status
+  'status/index.jade',
+  'status/status.jsx',
```
jonasfj: isn't status included in app.jsx, then there is no need for this...

menu.js
```diff
@@ -207,5 +207,15 @@ module.exports = [
     description: [
       "Manage credentials stored in the browser."
     ].join("\n")
+  },
+  {
+    title:        "Taskcluster status",
```
jonasfj: Can we just call "Status" ?

menu.js
```diff
@@ -207,5 +207,15 @@ module.exports = [
...
+    link:         "/status/",
+    icon:         "cogs",
+    display:      "true",
```
jonasfj: This should be a boolean

status/status.jsx
```diff
@@ -0,0 +1,201 @@
...
+];
+
+export let StatusChecker = React.createClass({
+  propTypes: {
+    up: React.PropTypes.bool.isRequired
+  },
+  render: function() {
+    let image = this.props.up ? "green-check.png" : "red-check.png";
+    return (
+      <img className="img-check img-fluid center-block" src={image}/>
```
jonasfj: If you get bored of png, play with awesome font:
```
<span className="fa-stack">
              <i className="fa fa-square fa-stack-2x fa-fw"></i>
              <i className="fa fa-terminal fa-stack-1x fa-fw fa-inverse"></i>
</span>
```

This is what I did in loaner to make the big icon fonts... I also had some css to change font-size :)


Anyways, that's not blocking this... Just something we can play around with some day...
As far as I'm aware if there is no XSS issues in the `/status` folder you are now the owner and designer of that tool :)

status/status.jsx
```diff
@@ -0,0 +1,201 @@
...
+            return <Service name={service.name} key={service.name} poll={service.poll}/>;
+          })}
+        </bs.ButtonToolbar>
+      </div>
+    );
+  }
+});
+
+export let TaskclusterStatus = React.createClass({
+  render: function() {
```
jonasfj: note... You can do: `render() {` in es6

jonasfj:
@walac fix issues in `build-files.js` and `menu.js` then merge...

r+

status/status.jsx
walac: drevil

PR #83, opened 2016-03-24
Bug 1247951 - In tools UI, hide secret content behind a click

andreadelrio:
Secret only shows after clicking a button.

### DIFF
--- a/secrets/secreteditor.jsx
+++ b/secrets/secreteditor.jsx
@@ -42,7 +42,8 @@ var SecretEditor = React.createClass({
 
       // Operation details, if currently doing anything
       working:          false,
-      error:            null
+      error:            null,
+      showSecret:       false
     };
   },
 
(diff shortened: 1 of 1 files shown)


andreadelrio:
@djmitche What do you think of this?

secrets/secreteditor.jsx
```diff
@@ -58,7 +59,7 @@ var SecretEditor = React.createClass({
         },
         editing:          true,
         working:          false,
-        error:            null
+        error:            null,
```
djmitche: Any reason to omit showSecret here?

andreadelrio: Yes, there's no CurrentSecretId so no secret to show.

djmitche:
I like it!

secrets/secreteditor.jsx
djmitche: Ah, thanks

PR #84, opened 2016-03-25
Remove useless whitespace in AWS provisioner

djmitche:
I don't know why bootstrap includes this margin, but it's what makes every row on the AWS provisioner take up twice as much vertical space as it should..


imbstack:
r+

PR #85, opened 2016-03-29
Add tooltip with components descriptions.

walac:
Status page should be descriptive for people who doesn't master
Taskcluster details. We add a few tooltips to help on that.


status/status.jsx
```diff
@@ -13,7 +14,8 @@ function makeRequest(options, allowHeaders = []) {
   }
 
   return $.ajax({
-    url: 'https://cors-proxy.taskcluster.net/request',
+    //url: 'https://cors-proxy.taskcluster.net/request',
+    url: 'http://localhost:8080/request',
```
djmitche: Was this a debugging change?

walac: err, fixed. I will add a debug config in a follow up PR.

package.json
```diff
@@ -11,51 +11,52 @@
...
-    "grunt-concurrent":                 "1.0.0",
-    "durable-json-lint":                "0.0.1",
-    "docker-exec-websocket-server":     "^1.1.0",
-    "hterm-umd":                        "^1.0.1",
-    "react-treeview":                   "^0.4.2",
-    "envify":                           "^3.4.0",
-    "shell-escape":                     "^0.2.0",
-    "babel-plugin-object-assign":       "1.2.1",
-    "classnames":                       "^2.2.3",
+    "babel-plugin-object-assign": "1.2.1",
```
selenamarie: do we have a linter setting for the justification that's in the old version of this file?

status/status.jsx
```diff
@@ -34,25 +35,30 @@ function dummyPoll(cb) {
 let taskclusterServices = [
   {
     name: "Queue",
-    poll: pollTaskclusterService.bind(null, "https://queue.taskcluster.net/v1/ping")
+    poll: pollTaskclusterService.bind(null, "https://queue.taskcluster.net/v1/ping"),
+    description: "https://queue.taskcluster.net"
```
selenamarie: Is this a link?  I think I'd like a link that goes directly to the ping output.

package.json
walac: I think it has more to do with npm managing the file.

status/status.jsx
walac: Put a link in the tooltip is not possible, what I did is making the check a link. The original url doesn't make sense for docker hub, since it needs to get a token first, so I provided a link to the main site.

selenamarie:
r+

walac:
Rebased to merge successfully. Also froze react-treeview version because latest release seems broken. I [opened an issue](https://github.com/chenglou/react-treeview/issues/33) for this

PR #86, opened 2016-04-02
Improving the documentation.

Anup-Allamsetty:
Adding some details to the documentation. Still not complete. Can you please suggest me what else can I add to the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,21 @@ for details.
 Developing TaskCluster Tools
 ============================
 
+Prerequisites for building TaskCluster Tools
+--------------------------------------------
+  - Node version 4.x
+  - Grunt-cli (Allows you to test and see the changes)
+  - Jade (installed as one of the dependencies)
+
+Building
+--------
+
+``` sh
+git clone https://github.com/taskcluster/taskcluster-tools.git
+cd taskcluster-tools
+npm install
+```
+
 Code Organization
 -----------------
   - lib/    (code intended for reuse)
@@ -30,7 +45,7 @@ Testing changes
 ---------------
 Install the dependencies needed for grunt and start it up:
 * npm install
-* grunt develop
+* grunt develop or (./node_modules/.bin/grunt develop)
 
 Grunt allows you to test and see your changes.
 The grunt default task builds, watches sources and serves them from
--- a/README.md
+++ b/README.md
@@ -17,10 +17,14 @@ Prerequisites for building TaskCluster Tools
   - Grunt-cli (Allows you to test and see the changes)
   - Jade (installed as one of the dependencies)
 
+```
+Note:  Grunt-cli and Jade will be installed as dependencies when you run "npm install".
+```
+
 Building
 --------
 
-``` sh
+```
 git clone https://github.com/taskcluster/taskcluster-tools.git
 cd taskcluster-tools
 npm install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,21 @@ for details.
 Developing TaskCluster Tools
 ============================
 
+Prerequisites for building TaskCluster Tools
+--------------------------------------------
+  - Node version 4.x
+  - Grunt-cli (Allows you to test and see the changes)
+  - Jade (installed as one of the dependencies)
```
djmitche: Both Grunt and Jade should be installed by 'npm install' -- did you need to install them separatelyi?

README.md
```diff
@@ -11,6 +11,21 @@ for details.
...
+
+Building
+--------
+
+``` sh
+git clone https://github.com/taskcluster/taskcluster-tools.git
+cd taskcluster-tools
+npm install
+```
```
djmitche: This is great :)

djmitche:
That's really up to you -- did you find anything else about setting up tools to be confusing or difficult?

Something you might add is to use https://ngrok.io to get an HTTPS URL, so that the "Sign In" button will work (it won't send credentials to an HTTP URL).  It's not too hard to figure out -- you basically sign up for a (free) account at the site, then set up the ngrok client, then run `ngrok http 9000`.

README.md
Anup-Allamsetty: No, I didn't have to install them separately, but just mentioning that these packages are also needed for development is fair in my opinion. 

I will just leave a note below saying that jade and grunt-cli will be installed while you run "npm install".

Is that fine?

Anup-Allamsetty:
Sure, I will, add that also.

README.md
djmitche: I'm not sure - there are something like 50-100 packages that npm install installs, so it doesn't seem like these two deserve a special mention.  I think it will generate questions like "how do I install Jade on Ubuntu" which we will answer with "you don't, that part of the README is confusing".

Anup-Allamsetty:
@djmitche, removed "grunt-cli" and "jade" from the dependencies.

README.md
```diff
@@ -58,3 +71,13 @@ Testing
...
+
+
+Ngrok Setup (optional)
+-----------
+Ngrok allows you to expose a web server running on your local machine to the internet.
+Ngrok is used creating a https connection, so that you can login to the taskcluster-tools.
+For using ngrok:
+  - Create an account (free) on [ngrok](https://ngrok.com/).
+  - Install ngrok - npm install ngrok
+  - Run ngrok - ngrok http 9000
```
djmitche: You may want to note that you need to do this in a separate terminal from where you are running `grunt develop`

README.md
```diff
@@ -30,7 +43,7 @@ Testing changes
 ---------------
 Install the dependencies needed for grunt and start it up:
 * npm install
-* grunt develop
+* grunt develop or (./node_modules/.bin/grunt develop)
```
djmitche: Oh, I forgot something to mention here -- `grunt develop --force` helps grunt not bail out when you save a syntax error or something similarly harmless

djmitche:
Thanks for working on this -- I think it will help the next person, too!

Anup-Allamsetty:
@djmitche done!! added the changes requested. 

Happy to help anytime :)

PR #89, opened 2016-04-11
Fix typos

chinhodado:


jonasfj:
thanks for the fix... we modified dependencies.. so the build is broken (wonders of npm)... anyways, it'll get fixed soon enough I'm sure...

petemoore:
Should get fixed in PR #90 :)

PR #88, opened 2016-04-11
Changed title from Previous Tasks to Previously Viewed Tasks

petemoore:
I didn't understand what "Previous Tasks" meant, so I reworded it.

PR #90, opened 2016-04-12
Fix docker-exec-websocker-server to version 1.1.0

petemoore:
This should hopefully fix the broken CI. :)

See https://github.com/taskcluster/taskcluster-tools/pull/88#issuecomment-209037925

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "grunt-keepalive":                  "0.0.1",
     "grunt-concurrent":                 "1.0.0",
     "durable-json-lint":                "0.0.1",
-    "docker-exec-websocket-server":     "^1.1.0",
+    "docker-exec-websocket-server":     "1.1.0",
     "hterm-umd":                        "^1.0.1",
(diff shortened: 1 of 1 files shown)


ckousik:
I had to deal with this when working on the visualizer... :)

package.json
```diff
@@ -49,7 +49,7 @@
     "grunt-keepalive":                  "0.0.1",
     "grunt-concurrent":                 "1.0.0",
     "durable-json-lint":                "0.0.1",
-    "docker-exec-websocket-server":     "^1.1.0",
+    "docker-exec-websocket-server":     "1.1.0",
```
djmitche: @jonasfj so did your analysis find that d-e-w-s had violated semver?

I wish we could include comments in package.json :(

jonasfj: I'm not entirely sure the bug is in websocket exec package thing... It could be a combination...
We recently did things to websocket-exec thing so it probably broke.

PR #93, opened 2016-04-18
Bug 1265518 - Create task group inspector section

eliperelman:


jonasfj:
from a quick look, this seems fine...

You are correct that `continuationToken` handling is missing... Possibly we should only load the next set when people scroll... regardless this is a good start...

But for rails giant graphs he'll definitely want `continuationToken`  support :)

task-group-inspector/taskgroupinspector.jsx
```diff
@@ -0,0 +1,469 @@
...
+   * @param {string} taskGroupId
+   * @returns {Promise}
+   */
+  listTaskGroup: function(taskGroupId) {
+    let payload = {
+      taskGroupId,
+      tasks: []
+    };
+
+    return new Promise((resolve, reject) => {
```
jonasfj: Don't use promises like this, it's so easy to break those promises...
Always strive to use `async/await` you are much less likely to drop promises... I do sometimes use `.catch(` but only as an alternative to `try{} catch{}`.
If you haven't play with it we have a tutorial:
http://docs.taskcluster.net/learn/async-javascript/

Here I would do:
```js
async listTaskGroup(taskGroupId) {
  let continuationToken = null;
  let tasks = [];
  do {
    let result = await this.queue.listTaskGroup(taskGroupId, continuationToken ? {continuationToken} : null);
    tasks.push(...result.tasks);
    continuationToken = result.continuationToken;
  } while(continuationToken);
  return tasks;
}
```

And yes, async/await works with loops, it makes your life much better... Also yes, you can do `{f() {return true;}}` instead of `{f: function() {return true;}}`, those are the same thing.. Which reads much nicer here...

task-group-inspector/taskgroupinspector.jsx
```diff
@@ -0,0 +1,469 @@
...
+          })
+          .catch(reject);
+      };
+
+      load();
+    });
+  },
+
+  /** Return promised state for TaskClusterMixin */
+  load: function() {
```
jonasfj: Just note, you can also write:
`load() {` It's magic :)

task-group-inspector/taskgroupinspector.jsx
eliperelman: Cool, didn't know we were using future syntax in here, good to know.

task-group-inspector/taskgroupinspector.jsx
eliperelman: Heh, was just being consistent with the other file. Hopefully everything will become much more magical in v2... 😈

PR #94, opened 2016-04-20
Bug 1261910: use listWorkerTypeSummaries() and state()

djmitche:
The `listWorkerTypeSummaries()` endpoint does not require
authentication, so this means that the site is visible even to those who
are not logged in.

As added bonuses:

 * the table renders immediately when `listWorkerTypeSummaries()`
   returns, much sooner than in older versions; pending counts are
   filled in as they arrive
 * the table's summary is updated whenver a workerType's details are
   loaded.

### DIFF
--- a/aws-provisioner/workertypetable.js
+++ b/aws-provisioner/workertypetable.js
@@ -28,11 +28,13 @@ var WorkerTypeRow = React.createClass({
 
   propTypes: {
     provisionerId: React.PropTypes.string.isRequired,
-    workerType: React.PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


jhford:
landed ea94be89dc8378d1286c2741de194a08e6718e88.  I just noticed after deployment that you're using '...' as a place holder.  We usually use the renderWaitFor thing, if you're touching this again, please switch to that.

djmitche:
I didn't use renderWaitFor since it's larger than a single line and thus adjusts line-height when it renders.

PR #95, opened 2016-04-20
Fixing a few papercuts in the group inspector UI

eliperelman:
@gregarndt maybe this would be an improvement?


jonasfj:
Nice

jonasfj:
Maybe we should invite some automatic staging site system for each PR.. Reviews like this are more fun that way :)

eliperelman:
That sounds awesome.

jonasfj:
`r+` btw

eliperelman:
Thanks!

PR #97, opened 2016-04-28
Render the worker type detail and create button at the top

jhford:
This small change moves the worker type detail to the top of the page.  This makes links to the worker types useful, since you don't need to scroll the entire length of the page to see the thing you requested.  It also makes the create worker type button infinitely more discoverable.  I would almost argue that we should rather show only a single worker if that's what's selected and show the table only when there's no selected worker.

### DIFF
--- a/aws-provisioner/workertypetable.js
+++ b/aws-provisioner/workertypetable.js
@@ -242,29 +242,33 @@ var WorkerTypeTable = React.createClass({
   },
 
   render() {
-    return <span>{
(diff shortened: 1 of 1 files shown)


jhford:
(I have no clue why this was ever put on the bottom... i didn't do it :) )

djmitche:
I'm going to bump this one to @eliperelman, since I don't really know if this is better or not.

I suspect that, with a wide enough viewport, the provisioner view would do the same thing most of the rest do: list on the left, details on the right.

eliperelman:
@jhford r+

PR #99, opened 2016-05-18
Bug 1272110 - Run taskcluster-interactive-shell.sh from $HOME

ahal:
Currently there is a Permission Denied error when trying to run /etc/taskcluster-interactive-shell.sh. Moving it to the $HOME directory instead seems like the simplest solution.

### DIFF
--- a/shell/app.jsx
+++ b/shell/app.jsx
@@ -34,7 +34,7 @@ $(function () {
           'if [ -z `which "$SHELL"` ]; then export SHELL="/.taskclusterutils/busybox sh"; fi;',
           'SPAWN="$SHELL";',
           'if [ "$SHELL" = "bash" ]; then SPAWN="bash -li"; fi;',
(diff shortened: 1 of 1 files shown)


jonasfj:
hmm... i guess it totally make sense that files in `/etc/` aren't executable...
Maybe they can't be?

@ahal, If `/etc/` files can't be executable... Maybe we're better off putting this into `/bin/taskcluster-interactive-shell`.
I think using `$HOME` is a bit scary, because that could change under us... Absolute paths are more predictable. And when building the image there is no reason we couldn't just put it in `/bin/`

After all, it is supposed to be a binary, specifically a shell.
Typically, a good implementation would be:
```sh
#!/bin/bash
export ...=...;
...
cd ...;
# do somethings...

exec bash --login # Continue as an interactive shell
```

jonasfj:
I propose we do: https://github.com/taskcluster/taskcluster-tools/pull/100

Which is almost the same I guess... I just prefer absolute paths.

PR #101, opened 2016-05-23
show task dependencies

djmitche:


lib/ui/taskinfo.jsx
```diff
@@ -172,6 +172,29 @@ var TaskInfo = React.createClass({
             <dd><code>{task.taskGroupId}</code></dd>
           )
         }
+        {
+          task.schedulerId === 'task-graph-scheduler' && task.dependencies.length > 0 ? (
```
eliperelman: Using `> 0` here is redundant; just use `task.dependencies.length`.

lib/ui/taskinfo.jsx
```diff
@@ -172,6 +172,29 @@ var TaskInfo = React.createClass({
...
+            <dt>Dependencies</dt>
+          ) : (
+            undefined
```
eliperelman: Using an empty string here is preferred over `undefined`.

lib/ui/taskinfo.jsx
```diff
@@ -172,6 +172,29 @@ var TaskInfo = React.createClass({
...
+        }
+        {
+          task.schedulerId === 'task-graph-scheduler' && task.dependencies.length > 0 ? (
+            <dd>
+              This task will run when all of the following tasks are
+              { task.requires == "all-completed" ? " completed successfully:" : " resolved in any way:" }
+              <ul>
+                {
+                  task.dependencies.sort().map((dep, index) =>
+                    <li key={index}><a href={'../task-inspector/#' + dep}>{dep}</a></li>)
```
eliperelman: You can avoid the concatenation if you move the path to outside of the evaluation:

```jsx
<li key={index}><a href="../task-inspector/#{dep}">{dep}</a></li>)
```

lib/ui/taskinfo.jsx
```diff
@@ -172,6 +172,29 @@ var TaskInfo = React.createClass({
...
+                }
+              </ul>
+            </dd>
+          ) : (
+            undefined
```
eliperelman: Use empty string.

lib/ui/taskinfo.jsx
```diff
@@ -172,6 +172,29 @@ var TaskInfo = React.createClass({
...
         }
+        {
+          task.schedulerId === 'task-graph-scheduler' && task.dependencies.length > 0 ? (
+            <dt>Dependencies</dt>
+          ) : (
+            undefined
+          )
+        }
+        {
+          task.schedulerId === 'task-graph-scheduler' && task.dependencies.length > 0 ? (
```
eliperelman: Remove `> 0`.

lib/ui/taskinfo.jsx
```diff
@@ -172,6 +172,29 @@ var TaskInfo = React.createClass({
...
+            <dd>
+              This task will run when all of the following tasks are
+              { task.requires == "all-completed" ? " completed successfully:" : " resolved in any way:" }
+              <ul>
+                {
+                  task.dependencies.sort().map((dep, index) =>
```
gregarndt: Is there a way to have this be a list that can be collapsed? Maybe it is, I don't know enough about the syntax used in these templates.

lib/ui/taskinfo.jsx
gregarndt: doesn't an empty list evaluate true?

eliperelman: The array does, not its length.

gregarndt: Ah, sorry, I stripped off the length too when thinking about it.  Thanks for the clarification.

eliperelman: :thumbs up:

lib/ui/taskinfo.jsx
djmitche: That would mean some extra widget-building.  I don't think the *dependencies* will be too large most of the time.  It's the reverse dependencies (build -> test) that will be huge.

djmitche:
updated

eliperelman:
R+

lib/ui/taskinfo.jsx
gregarndt: Oh you're right, since reverse dependencies was still fresh in my head, I for some reason though is what we were adding here.  Clearly I can't review today.

PR #102, opened 2016-05-31
Custom routes are always prefixed route.

jonasfj:


djmitche:
Ah, I thought that was just me :)

PR #103, opened 2016-06-03
Bug 1147963 - [Fix] Log viewer (Inspect Task) has very poor UX 

andreadelrio:
Builds on top of PR #73 
Replaced stringification approach for log-fetcher webworker to use webworkify instead.
Added ANSI color support.

### DIFF
--- a/lib/ui/log-fetcher.js
+++ b/lib/ui/log-fetcher.js
@@ -2,48 +2,45 @@ module.exports = function (self) {
   var request;
   var dataOffset = 0;
 
-  function divide(data){
-    var lines = [];
-    data.split('\n').forEach(function(line){
-      lines.push(line);
-    });
-    return lines;
+  function divide(data) {
+    return data.split('\n');
   }
 
-  function onData(e){
+  function onData(e) {
     var resp = {};
-    if (request.responseText !== null ||
-      request.responseText !== undefined){
+    if (request.responseText != null) {
       // Check if we have new data
       var length = request.responseText.length;
-      if (length > dataOffset){
+      if (length > dataOffset) {
         resp.data = divide(request.responseText);
         resp.done = false;
       }
     }
     // When request is done
-    if (request.readyState === request.DONE){
+    if (request.readyState === request.DONE) {
       resp.done = true;
       // Write an error, if request failed
-      if (request.status !== 200){
+      if (request.status !== 200) {
         resp.data = ["\r\n[task-inspector] Failed to fetch log!\r\n"];
       } else {
         resp.data = divide(request.responseText);
       }
-      }
+    }
     postMessage(resp);
-    if(resp.done) close();
+    if (resp.done) {
+      close();
+    } 
   }
 
-  function abort(){
+  function abort() {
     request.removeEventListener('progress', onData);
     request.removeEventListener('load', onData);
     request.abort();
     close();
   }
 
-  self.addEventListener('message', function(e){
-    if(e.data.url){
+  self.addEventListener('message', function(e) {
+    if (e.data.url) {
       request = new XMLHttpRequest();
       request.open('get', e.data.url, true);
       request.addEventListener('loadstart', onData);
@@ -51,6 +48,8 @@ module.exports = function (self) {
       request.addEventListener('load', onData);
       request.send();
     };
-    if(e.data.abort) abort();
+    if (e.data.abort) {
+      abort();
+    } 
   });
  };
--- a/lib/ui/log-fetcher.js
+++ b/lib/ui/log-fetcher.js
@@ -3,36 +3,38 @@ var request;
 var dataOffset = 0;
 var cols;
 
+function divide(data){
+    var lines = [];
+    data.split('\n').forEach(function(line){
+        for(var i = 0; i < line.length; i += cols)
+            lines.push(line.substr(i, cols));
+    });
+    return lines;
+}
+
 function onData(e){
     var resp = {};
     if (request.responseText !== null ||
-        request.responseText !== undefined) {
-      // Check if we have new data
-      var length = request.responseText.length;
-      if (length > dataOffset) {
-        var data = "Loading log: " + e.loaded + ' / ' + e.total;
-        resp.data = [data];
-        resp.done = false;
-      }
+        request.responseText !== undefined){
+        // Check if we have new data
+        var length = request.responseText.length;
+        if (length > dataOffset){
+            resp.data = divide(request.responseText);
+            resp.done = false;
+        }
     }
     // When request is done
-    if (request.readyState === request.DONE) {
-      resp.done = true;
-      // Write an error, if request failed
-      if (request.status !== 200){
-        resp.data = ["\r\n[task-inspector] Failed to fetch log!\r\n"];
-      } else {
-        var temp = request.responseText.split('\n');
-        var lines = [];
-        temp.forEach(function(line){
-            for(var i = 0; i < line.length; i += cols)
-                lines.push(line.substr(i, cols));
-        });
-        resp.data = lines;
+    if (request.readyState === request.DONE){
+        resp.done = true;
+        // Write an error, if request failed
+        if (request.status !== 200){
+            resp.data = ["\r\n[task-inspector] Failed to fetch log!\r\n"];
+        } else {
+            resp.data = divide(request.responseText);
+        }
       }
-    }
-  postMessage(resp);
-  if(resp.done) close();
+    postMessage(resp);
+    if(resp.done) close();
 }
 
 function abort(){
--- a/lib/ui/log-fetcher.js
+++ b/lib/ui/log-fetcher.js
@@ -1,0 +1,57 @@
+var source = function(){
+var request;
+var dataOffset = 0;
+var cols;
+
+function onData(e){
+    var resp = {};
+    if (request.responseText !== null ||
+        request.responseText !== undefined) {
+      // Check if we have new data
+      var length = request.responseText.length;
+      if (length > dataOffset) {
+        var data = "Loading log: " + e.loaded + ' / ' + e.total;
+        resp.data = [data];
+        resp.done = false;
+      }
+    }
+    // When request is done
+    if (request.readyState === request.DONE) {
+      resp.done = true;
+      // Write an error, if request failed
+      if (request.status !== 200){
+        resp.data = ["\r\n[task-inspector] Failed to fetch log!\r\n"];
+      } else {
+        var temp = request.responseText.split('\n');
+        var lines = [];
+        temp.forEach(function(line){
+            for(var i = 0; i < line.length; i += cols)
+                lines.push(line.substr(i, cols));
+        });
+        resp.data = lines;
+      }
+    }
+  postMessage(resp);
+  if(resp.done) close();
+}
+
+function abort(){
+    request.removeEventListener('progress', onData);
+    request.removeEventListener('load', onData);
+    request.abort();
+    close();
+}
+
+onmessage = function(e){
+    if(e.data.url){
+        cols = e.data.cols;
+        request = new XMLHttpRequest();
+        request.open('get', e.data.url, true);
+        request.addEventListener('progress', onData);
+        request.addEventListener('load', onData);
+        request.send();
+    };
+    if(e.data.abort) abort();
+}
+}.toString().split('\n').slice(1,-1).join('\n');
+module.exports = window.URL.createObjectURL(new Blob([source]));
--- a/lib/ui/log-fetcher.js
+++ b/lib/ui/log-fetcher.js
@@ -1,59 +1,56 @@
-var source = function(){
-var request;
-var dataOffset = 0;
-var cols;
+module.exports = function (self) {
+  var request;
+  var dataOffset = 0;
 
-function divide(data){
+  function divide(data){
     var lines = [];
     data.split('\n').forEach(function(line){
-        for(var i = 0; i < line.length; i += cols)
-            lines.push(line.substr(i, cols));
+      lines.push(line);
     });
     return lines;
-}
+  }
 
-function onData(e){
+  function onData(e){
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -12,6 +12,8 @@
     "url": "https://github.com/taskcluster/taskcluster-tools.git"
   },
   "dependencies": {
+    "ansi-regex": "^2.0.0",
+    "ansi_up": "^1.3.0",
```
walac: I think you forgot to add `workify` here

andreadelrio: Thanks for spotting that, I just added it.

eliperelman:
Please use spaces between function parens and the opening brace, e.g. `myFunction() {`

eliperelman:
Same for conditions:

```js
if (condition) {
   ...
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+        height: this.scrollbarHeight(),
+        marginTop: this.scrollbarMargin()
+        }} ref="scrollbar"/>
+      </div>;
   }
 });
 
 // Export TerminalView
-module.exports = TerminalView;
+module.exports = TerminalView;
```
eliperelman: Add newline to end of file.

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+      var frame = this.state.lines.slice(start, start + this.props.rows);
+    } else {
+      var frame = this.state.lines.slice(start + 15, start + this.props.rows + 15);
+    }
+    return <div className="viewer" onWheel={this.onMouseWheel}>
+      <div className="buffer" ref="buffer">
+      {
+        frame.map(function(line){
+          // Check if there are any ansi colors/styles
+          if (ansiRegex().test(line) === true) {
```
eliperelman: `=== true` is redundant:

```js
if (ansiRegex().test(line)) {
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+    if(e.button == 0){
+      this.dragging = true;
+      this.startY = e.pageY;
+      this.margin = this.scrollbarMargin();
+      this.startOffset = this.state.fromBottom;
+    }
+  },
+
+  onMouseUp(e){
+    if(e.button == 0)
```
eliperelman: - Let's keep conditionals consistent
- Use a value's truthiness or falsiness to evaluate condition
- Prefer `===` when value coercion isn't sufficient

```js
if (e.button) {
  this.dragging = false;
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+    if(ratio > 1) ratio = 1;
+    var height = ratio * this.refs.buffer.offsetHeight;
+    return Math.max(height, 10);
   },
 
-  componentWillUnmount: function() {
-    if (this.request) {
-      this.abortRequest();
+  scrollbarMargin(){
+    if(!this.refs.buffer) return 0;
```
eliperelman: Keep conditionals consistent:

```js
if (!this.refs.buffer) {
  return 0;
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
-        this.term.write(data);
+    if (response.data){
+      var newFromBottom = 0;
+      if(!this.props.scrollDown){
+        // we don't expect the data to get shrunk
+        // since it's a log, it can only grow
+        newFromBottom += response.data.length - this.state.lines.length;
       }
+      this.setState({
+        lines : response.data,
```
eliperelman: nit: remove extra space before colon:

```js
lines: response.data,
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
-    this.dataOffset = 0;
-    this.request = new XMLHttpRequest();
-    this.request.open('get', this.props.url, true);
-    this.request.addEventListener('progress', this.onData);
-    this.request.addEventListener('load', this.onData);
-    this.request.send();
+    this.dataOffset = 0;  
+    this.worker = work(require('./log-fetcher.js'));
+    this.worker.addEventListener('message', this.onData);
+    this.worker.postMessage({url: this.props.url});
```
eliperelman: Use spacing with object braces:

```diff
- ({url: this.props.url});
+ ({ url: this.props.url });
```

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,56 @@
...
+      request = new XMLHttpRequest();
+      request.open('get', e.data.url, true);
+      request.addEventListener('loadstart', onData);
+      request.addEventListener('progress', onData);
+      request.addEventListener('load', onData);
+      request.send();
+    };
+    if(e.data.abort) abort();
+  });
+ };
```
eliperelman: Insert newline at end of file.

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,56 @@
+module.exports = function (self) {
+  var request;
+  var dataOffset = 0;
+
+  function divide(data){
```
eliperelman: This function can be simplified to:

```js
function divide(data) {
  return data.split('\n');
}
```

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,56 @@
...
+    var lines = [];
+    data.split('\n').forEach(function(line){
+      lines.push(line);
+    });
+    return lines;
+  }
+
+  function onData(e){
+    var resp = {};
+    if (request.responseText !== null ||
```
eliperelman: This condition can be simplified to:

```js
if (request.responseText != null) {
  ...
}
```

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,56 @@
...
+    // When request is done
+    if (request.readyState === request.DONE){
+      resp.done = true;
+      // Write an error, if request failed
+      if (request.status !== 200){
+        resp.data = ["\r\n[task-inspector] Failed to fetch log!\r\n"];
+      } else {
+        resp.data = divide(request.responseText);
+      }
+      }
```
eliperelman: Dedent this curly brace 2 spaces to the left.

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,56 @@
...
+    postMessage(resp);
+    if(resp.done) close();
```
eliperelman: Make this conditional consistent:

```js
if (resp.done) {
  close();
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
-      this.term.showCursor();
+  },
 
-      // Write an error, if request failed
-      if (this.request.status !== 200) {
-        this.term.write("\r\n[task-inspector] Failed to fetch log!\r\n");
-      }
+  abortRequest(){
+    if(this.worker){
+      this.worker.postMessage({abort: true});
```
eliperelman: Object spacing:

```js
- ({abort: true})
+ ({ abort: true })
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+    if(!this.refs.buffer) return 0;
+    var ratio = (this.state.lines.length - this.state.fromBottom - this.props.rows) / this.state.lines.length;
+    return ratio * (this.refs.buffer.offsetHeight - this.scrollbarHeight());
+  },
+
+  scrollbarSet(newState){
+    newState = Math.floor(newState);
+    newState = Math.max(0, newState);
+    newState = Math.min(this.state.lines.length - this.props.rows, newState);
+    if(newState != this.state.fromBottom)
```
eliperelman: Use `!==`

lib/ui/log-fetcher.js
```diff
@@ -0,0 +1,56 @@
...
+  self.addEventListener('message', function(e){
+    if(e.data.url){
+      request = new XMLHttpRequest();
+      request.open('get', e.data.url, true);
+      request.addEventListener('loadstart', onData);
+      request.addEventListener('progress', onData);
+      request.addEventListener('load', onData);
+      request.send();
+    };
+    if(e.data.abort) abort();
```
walac:
```javascript
if (e.data.abort) {
  abort();
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
-  open: function() {
-    // Destroy existing terminal if there is one
-    if (this.term) {
-      this.term.destroy();
-      this.term = null;
-    }
+  open(){
     // Abort previous request if any
-    if (this.request) {
+    if (this.request){
```
walac: nit: undo this change

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
 
-  abortRequest: function() {
-    this.request.removeEventListener('progress', this.onData);
-    this.request.removeEventListener('load', this.onData);
-    this.request.abort();
-    this.request = null;
+  /* Some methods to work with the scrollbar. * 
+   * Maybe should become a separate class?    */
+  scrollbarHeight(){
+    if(!this.refs.buffer) return 0;
```
walac:
```javascript
if (!this.refs.buffer) {
  return 0;
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+    var ratio = this.props.rows / this.state.lines.length;
+    if(ratio > 1) ratio = 1;
```
walac:
```javascript
if (!this.refs.buffer) {
  return 0;
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+
+  componentDidMount(){
+    this.refs.scrollbar.addEventListener('mousedown', this.onMouseDown);
+    window.addEventListener('mouseup', this.onMouseUp);
+    window.addEventListener('mousemove', this.onMouseMove);
+  },
+
+  render(){
+    var start = this.state.lines.length - this.state.fromBottom - this.props.rows;
+    if(start < 0) start = 0;
```
walac:
```javascript
if (start < 0) {
  start = 0;
}
```

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+    //Check if the log has less lines than the number of rows or if we are displaying the beggining of the log
+    if ((this.state.lines.length < this.props.rows) || (this.state.lines[start] == this.state.lines[0])) {
```
walac: use `===` instead of `==`

lib/ui/terminalview.jsx
```diff
@@ -14,125 +16,174 @@ var TerminalView = React.createClass({
...
+      var frame = this.state.lines.slice(start, start + this.props.rows);
+    } else {
+      var frame = this.state.lines.slice(start + 15, start + this.props.rows + 15);
+    }
```
walac: I think this is a bit better:

```javascript
var displacement = 0;
if (this.props.rows <= this.state.lines.length && this.state.lines[start] !== this.state.lines[0]) {
  displacement = 15;
}
var frame = this.state.lines.slice(start + displacement, start + this.props.rows + displacement);
```

walac:
generally advice: s/`var`/`let`/

I tested the implementation and it looks really good!

@jonasfj  Do you have any additional comments?

jonasfj:
So this is definitely better than what we have... I might not be the best to evaluate performance considerations, so maybe someone else test for that...

As for interaction, we probably have to:
A) Implement selection, scroll bar, page up/down, arrow keyboard, searching, window  resize, full screen mode, etc.
B) Find an existing viewport implementation..

(A) certainly isn't trivial... but it could be the most fun thing to do... Basically, it involves rebuilding a textbox display in javascript... There is lots of particularities that isn't implemented in the current view:
 * control + a, to select all,
 * scroll by dragging selection
 * Move selection with arrow keys + shift ( + control to jump by word boundaries) 
 * page up/down, arrow keys
 * text searching (control + f)
 * full screen mode
 * Click on scroll bar (dragging works)
 * Clipboard handling (stripping colors and only copying plain text)

I'm sure there is more issues... There are so many things we take for given in a  textview.
Reimplementing them all isn't trivial... We don't have to do them all at once, but probably consider it.

Final note:
I see that the this now scrolls through the log as it is downloaded... That's super nice :)
But maybe we can do something so that animation doesn't happen when we load a log that is fully present (check for content-length header, or reduce animation speed with requestAnimationFrame).

Actually, using `window.requestAnimationFrame` might be a smart trick for smoother operation...

jonasfj:
I think that it also needs some scroll speed adjustment... Usually one can make things scroll faster by spinning the wheel faster :) he he...

Another option worth considering is calculating full log height using line_height * number_of_lines... Javascript can tell you the computed size... Using this it might be possible to make the animation more smooth by support scrolling half a line... Granted this requires more math :)

jonasfj:
Note: again, this is probably better than what we have... So we totally could land this...
Just saying we need to either find a real viewport implementation, or look to fix the short comings of this implementation. In which case we might have to look at qt for inspiration on how to actually implement a viewport correctly. Codemirror and ace could also be used for inspiration as both of them pretty much does this in HTML/javascript.

walac:
I filed a [follow up Bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1278288) for @jonasfj suggestions

PR #105, opened 2016-06-06
Create interactive tasks with high priority

ahal:


lib/ui/loaner-button.jsx
```diff
@@ -112,6 +113,9 @@ let LoanerButton = React.createClass({
     // Set task,retries to 0
     task.retries = 0;
 
+    // Set high priority as a developer is waiting for it in real time
+    task.priority = 'high'
```
eliperelman: nit: missing semicolon.

ahal:
Thanks. I'm going to close this as it doesn't actually work (see bug 1271677). I mostly just needed a diff up, and in hindsight, should have done a "compare" on my fork, rather than a pull request. Sorry.

jonasfj:
Too late I think we should consider taking this...

Well, depending on what we do with `queue:priority:high`, if we give to everybody we should just set it here...

Otherwise, maybe we could deduce a way to determine if people have the `queue:priority:high` scope.

jonasfj:
Discussion on [bug 1271677](https://bugzilla.mozilla.org/show_bug.cgi?id=1271677) to see if we open the scope or not...

djmitche:
I'll close this since discussion in the bug has stalled.

PR #106, opened 2016-06-08
Fix link for dependencies in task info view

acmiyaguchi:
The referral link under Dependencies in a task info view will direct you to an invalid url. This should fix the broken link.


djmitche:
Thanks!  I kept meaning to fix that :)

PR #107, opened 2016-06-09
Added notification for loaner being ready...

jonasfj:
Sort of a hack... but should give us a neat little system notification like irccloud does...

Note: I had some issues with this on localhost + firefox, but it works in chrome, and sometimes in FF... I suspect rate-limiting or something is playing a trick on me... Or maybe just libnotify on linux..


imbstack:
I'm curious what the issues would be that are causing this to not work sometimes. I think the best thing is to just deploy it and see how often it works in the real world since it's a non-essential thing anyway.

r+

jonasfj:
Firefox had some rate limiting for notifications... That might be it..

Or maybe it was just bugs on nightly...

PR #108, opened 2016-06-10
Fixed the task-group link

jonasfj:


lib/ui/taskinfo.jsx
```diff
@@ -161,40 +161,42 @@ var TaskInfo = React.createClass({
...
+            {task.taskGroupId}
+          </a>
+        </dd>
+        <dt>Dependencies</dt>
+        <dd>
+          {
+            task.dependencies.length > 0 ? (
               <ul>
                 {
-                  task.dependencies.sort().map((dep, index) =>
```
imbstack: I won't pretend to have a great idea of what's going on here, but if we drop this `sort()` won't this list be in arbitrary orders on ever refresh?

/me has no idea how any of this actually works.

jonasfj: I think it'll be the same arbitrary list...

Queue doesn't reorder them for fun... hmm... I wonder if there is a good reason to change their ordering from when queue.createTask is created...

At first thought I'm tempted to sort... on second thought not doing so preserves the order given in queue.createTask... I suspect that order might often have say the docker-image build dependency at the top, or in some specific order as listed in-tree...

So it's tempting not to sort by slugid... after all sort by uuid carries little value.

imbstack:
## r+

We can always add a sort in later if we want to.

PR #110, opened 2016-06-14
Diagnostics

ckousik:


diagnostics/app.jsx
```diff
@@ -0,0 +1,12 @@
+var ReactDOM		= require('react-dom');
+var Diagnostics	= require('./diagnostics');
+var $ 					= require('jquery');
+var bs 					=	require('react-bootstrap');
+var React 			=	require('react');
```
djmitche: (nit) whitespace

djmitche:
I merged this and ran "grunt --force", but the diagnostics page doesn't show anything for me :(

diagnostics/diagnostics.jsx
```diff
@@ -0,0 +1,66 @@
...
+import Retrieve from './retrieve';
+import { Grid, Row } from 'react-bootstrap';
+
+export default class Diagnostics extends React.Component {
+
+  constructor (props) {
+    super(props);
+    
+    this.state = {
+      jsonLog:  '',
```
djmitche: This is a string here, but in `componentWillMount` it looks like it's parsed JSON?

diagnostics/diagnostics.jsx
```diff
@@ -0,0 +1,66 @@
...
+    return <Grid> 
+      <h4>Date: {this.state.testDate}</h4>
+      <h4>TestId: {this.state.testId}</h4>
+      <Row>
+        {
+          _.keys(this.state.jsonLog).map(s => <Service key={s} title={s} test={this.state.jsonLog[s]} />)
+        }
+      </Row>
+      <h3>Log: </h3>
+      <RawLog text={this.state.rawLog} />
```
djmitche: Would it make more sense to hide this behind a click?  Perhaps just linking to a raw text file?  It feels duplicative to have the broken-out JSON results (which are awesome!) with the logs below.

ckousik: This log contains both stdout and stderr, so it does seem duplicative when all tests are passing but it should prove useful in case a test fails. Hiding it behind a click will be trivial though

djmitche:
:+1: from me!

djmitche:
@jonasfj is away, so I'll merge this for now.  Jonas should feel free to make additional comments when he's back :)

PR #112, opened 2016-06-20
Fixed https error in diagnostics

ckousik:
The logs stored in s3 were being requested using http instead of https. This caused the browser to block the request. This pr fixes that.


djmitche:
Any idea why this PR has so many commits in it?  Can you rebase it on the master branch?

PR #115, opened 2016-06-28
Added support for lines spanning multiple rows in terminal

andreadelrio:
This is to fix the bug reported by @gregarndt on IRC.
You can test it with the same TaskId EHkFHgEYSPeGqMmJt43JBg

### DIFF
--- a/lib/ui/log-fetcher.js
+++ b/lib/ui/log-fetcher.js
@@ -1,9 +1,24 @@
 module.exports = function (self) {
   var request;
   var dataOffset = 0;
+  var ansiRegex = require('ansi-regex');
+
 
   function divide(data) {
-    return data.split('\n');
+    let tester = ansiRegex();
+    let matcher = /.{1,139}/g;
+
+    return data
+      .split('\n')
+      .reduce((arr, line) => {
+        if (line.length < 140 || tester.test(line)) {
+          arr.push(line);
+        } else {
+          arr.push(...line.match(matcher));
+        }
+
+        return arr;
+      }, []);
   }
 
   function onData(e) {


djmitche:
looks good to me

lib/ui/log-fetcher.js
```diff
@@ -1,9 +1,20 @@
 module.exports = function (self) {
   var request;
   var dataOffset = 0;
+  var ansiRegex  = require('ansi-regex');
```
eliperelman: nit: remove extra space before assignment

lib/ui/log-fetcher.js
```diff
@@ -1,9 +1,20 @@
...
 
-  function divide(data) {
-    return data.split('\n');
+  function divide(data){
+    var lines = [];
```
eliperelman: Could this be simplified to the following?

```js
data
  .split('\n')
  .map(line => line.substr(0, 139));
```

walac: hrm, does this replace the current implementation? When line is larger than 140 chars, it is broken in 140 of 140. Also, if it is larger than 140 chars, but matches the regex, it is added all together.

One solution I see is this:

```javascript
let newData = data
  .split('\n')
  .map(line => ( line.length < 140 || ansiRegex().test(line) ) ? line : line.match(/.{1,139}/g));

return [].concat.apply(...newData);
```

But I am not sure it is a better solution than the original one and `match` is pretty slow.

eliperelman: I guess it would help to know what the actual intent is. Is the intent to truncate lines longer than 139 chars? Is it to split long lines into multiple lines?

andreadelrio: The intent is to split long lines that don't have ANSI codes into lines of 139 chars.

eliperelman: @andreadelrio so one line can become many?

andreadelrio: @eliperelman yes

eliperelman: Ah ok, that makes things clearer. I think @walac is on the right track, and I wouldn't worry too much about `match` being slow for a short string. You could even cache a couple things to make the process a bit faster:

```js
let tester = ansiRegex();
let matcher = /.{1,139}/g;

return data
  .split('\n')
  .reduce((arr, line) => {
    if (line.length < 140 || tester.test(line)) {
      arr.push(line);
    } else {
      arr.push(...line.match(matcher));
    }
    
    return arr;
  }, []);
```

I think that would do it.

andreadelrio: @eliperelman Yes, that works thanks. Just pushed the changes.

PR #117, opened 2016-07-05
Bug 1274624 - Fixes floating zero in taskview UI

eliperelman:
Note that this is a silly solution thanks to JSX syntax. 😐


eliperelman:
@gregarndt this should solve your issue. r?

gregarndt:
r+

PR #119, opened 2016-07-07
Bug 1284617 - new live log viewer doesn't break lines as well as the old one

andreadelrio:
Added support for progress tables and more ways to break lines in log viewer.
Tested with:

https://tools.taskcluster.net/task-graph-inspector/#K0Fihhx4SNurUEEQrx8zZw/Hc0RyZgsRNGA2hQ2kle4Gw/0
https://tools.taskcluster.net/task-graph-inspector/#K0Fihhx4SNurUEEQrx8zZw/dWW6AGkFRa-oYpgYezfYMg/0
https://tools.taskcluster.net/task-inspector/#bNtU135dR3SqROzO-ZBa3Q/0

### DIFF
--- a/lib/ui/log-fetcher.js
+++ b/lib/ui/log-fetcher.js
@@ -9,14 +9,14 @@ module.exports = function (self) {
     let matcher = /.{1,139}/g;
 
     return data
-      .split('\n')
+      .split(/\n|\[K|\[[0-9]B|\[[1-9][0-9]B/)
       .reduce((arr, line) => {
-        if (line.length < 140 || tester.test(line)) {
+        if (line.length < 140 || line.match(tester)) {
           arr.push(line);
         } else {
           arr.push(...line.match(matcher));
         }
-
+        
         return arr;
       }, []);
   }
@@ -68,3 +68,4 @@ module.exports = function (self) {
     } 
   });
  };
+ 
--- a/lib/ui/terminalview.jsx
+++ b/lib/ui/terminalview.jsx
@@ -19,7 +19,7 @@ var TerminalView = React.createClass({
   getDefaultProps() {
     return {
       url:            undefined,  // No URL to display at this point
-      rows: 40,
+      rows: 32,
       scrollDown: false,
     };
   },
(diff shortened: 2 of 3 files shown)


lib/ui/terminalview.jsx
```diff
@@ -172,15 +173,20 @@ var TerminalView = React.createClass({
     if (this.props.rows <= this.state.lines.length && this.state.lines[start] !== this.state.lines[0]) {
       paddingRows = 15;
     }
-    var frame = this.state.lines.slice(start + paddingRows, start + this.props.rows + paddingRows);
+    var frame = this.state.lines.slice(start + paddingRows, start + this.props.rows + paddingRows + 5);
     return <div className="viewer" onWheel={this.onMouseWheel}>
       <div className="buffer" ref="buffer">
       {
         frame.map(function(line) {
+          var divStyle = {
```
eliperelman: Prefer CSS classes over inline styles.

lib/ui/terminalview.jsx
```diff
@@ -163,6 +163,7 @@ var TerminalView = React.createClass({
   },
 
   render() {
+    console.log(this.state.lines.length)
```
eliperelman: Remove `console.log`.

lib/ui/log-fetcher.js
```diff
@@ -2,21 +2,24 @@ module.exports = function (self) {
...
     let tester = ansiRegex();
     let matcher = /.{1,139}/g;
+    const separators = _
+      .range(0,90)
+      .map(i => '\\[' + i + 'B')
+      .concat(['\\\n', '\\[K'])
 
     return data
-      .split('\n')
+      .split(new RegExp(separators.join('|'), 'g'))
```
eliperelman: Having a range from 0-90 is going to produce a *long* regex. I haven't tested this but what about using something like this instead?

```js
/*
  The goal being to match:
    \n
    [K
    [ with numbers 0-9
    [ with numbers 10-89
    [ with number 90
.split(/\n|\[K|\[[0-9]|\[[1-8][0-9]|\[90/)
```

Thoughts? Might want to test it. 😆

lib/ui/log-fetcher.js
```diff
@@ -2,21 +2,24 @@ module.exports = function (self) {
...
       .reduce((arr, line) => {
         if (line.length < 140 || tester.test(line)) {
           arr.push(line);
         } else {
           arr.push(...line.match(matcher));
         }
-
```
eliperelman: Leave this blank line.

lib/ui/log-fetcher.js
andreadelrio: @eliperelman I've seen cases up to [29B so if we cover 0B to 99B it should be good for now. I just pushed changes with a shorter regex based on your suggestion.

lib/ui/log-fetcher.js
walac: The problem is the trailing spaces in the diff below.

lib/ui/terminalview.jsx
```diff
@@ -167,20 +167,21 @@ var TerminalView = React.createClass({
...
+    var frame = this.state.lines.slice(start, start + this.props.rows);
     return <div className="viewer" onWheel={this.onMouseWheel}>
       <div className="buffer" ref="buffer">
       {
         frame.map(function(line) {
           // Check if there are any ansi colors/styles
           if (ansiRegex().test(line)) {
             var new_line = ansi_up.ansi_to_html(line);
-            return <div key={start++} dangerouslySetInnerHTML={{__html: new_line}}></div>;
+            if (line.match(/  /g) && line.match(/  /g).length > 0) {
```
walac: We could save `line.match` result in a temporary variable to avoid the double call.

lib/ui/terminalview.jsx
```diff
@@ -167,20 +167,21 @@ var TerminalView = React.createClass({
...
+              return <div className="pre-white" key={start++} dangerouslySetInnerHTML={{__html: new_line}}></div>;
+            } else {
+              return <div key={start++} dangerouslySetInnerHTML={{__html: new_line}}></div>;
+            }
+          } else if (line.match(/  /g) && line.match(/  /g).length > 0) {
```
walac: Same here.

lib/ui/log-fetcher.js
```diff
@@ -67,4 +67,4 @@ module.exports = function (self) {
       abort();
     } 
   });
- };
+ };
```
eliperelman: Keep a blank line at the end of the file.

lib/ui/terminalview.less
```diff
@@ -30,3 +30,7 @@
   width: 20px;
   background-color: #7f7f7f;
 }
+
+.pre-white {
+  white-space: pre-wrap;
+}
```
eliperelman: Add a blank line to the end of the file.

lib/ui/terminalview.jsx
```diff
@@ -196,4 +198,4 @@ var TerminalView = React.createClass({
 });
 
 // Export TerminalView
-module.exports = TerminalView;
+module.exports = TerminalView;
```
eliperelman: Add a blank line to the end of the file.

PR #120, opened 2016-07-13
Swapping markdown library

eliperelman:
Swapped the Markdown library. I can confirm that Markdown containing data URIs renders as plain text, and safe URLs render rich.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -18,7 +18,7 @@ var COMMON_MODULES = [
   'slugid',
   'debug',
   'lodash',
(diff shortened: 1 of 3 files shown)


eliperelman:
@jonasfj r?

jonasfj:
# `r+`

there is a bug for this somewhere too... if you want to resolve something on bugzilla :)

eliperelman:
@jonasfj yep! Just being secret agent... 😎

claudijd:
👍

PR #122, opened 2016-07-23
Adding sign in prompt when a feature needs it

andreadelrio:
This is to fix [Bug 1248977](https://bugzilla.mozilla.org/show_bug.cgi?id=1248977) - [Prompt user to login if a feature needs it]

### DIFF
--- a/lib/layout.jsx
+++ b/lib/layout.jsx
@@ -9,13 +9,6 @@ let format      = require('./format');
 // time before expiration at which we warn
 let EXPIRY_WARNING = 5 * 60 * 1000;
 
-/** Log-in open a authentication URL */
-function signIn() {
-  window.open(auth.buildLoginURL(), '_blank');
-}
-
-exports.signIn = signIn;
-
 /** Navigation bar for layout.jade */
 let Navigation = React.createClass({
   /** Get initial state */
@@ -27,6 +20,10 @@ let Navigation = React.createClass({
     };
   },
 
+  /** Log-in open a authentication URL */
+  signIn() {
+    window.open(auth.buildLoginURL(), '_blank');
+  },
 
   /** Log out (clear credentials) */
   signOut() {
@@ -187,7 +184,7 @@ let Navigation = React.createClass({
     // if there are no credentials at all, then there is no menu -- just a sign-in link
     if (!this.state.credentials) {
       return (
-        <bs.NavItem onSelect={signIn} ref="credentials">
+        <bs.NavItem onSelect={this.signIn} ref="credentials">
           <bs.Glyphicon glyph="log-in"/>&nbsp;Sign in
         </bs.NavItem>
       );
@@ -204,7 +201,7 @@ let Navigation = React.createClass({
           <format.Icon name="key"/>&nbsp;Manage Credentials
         </bs.MenuItem>
         <bs.MenuItem divider />
-        <bs.NavItem onSelect={signIn}>
+        <bs.NavItem onSelect={this.signIn}>
           <bs.Glyphicon glyph="log-in"/>&nbsp;
           Sign In
         </bs.NavItem>
@@ -225,13 +222,17 @@ let Navigation = React.createClass({
         <bs.Overlay
           show={true}
           rootClose={true}
-          onHide={ () => this.setState({credentialsMessage: undefined}) }
+          onHide={this.overlayHideHandler}
           placement="bottom"
-          target={ props => ReactDOM.findDOMNode(this.refs.credentials) }>
+          target={props => ReactDOM.findDOMNode(this.refs.credentials)}>
           {popover}
         </bs.Overlay>
       );
     }
+  },
+
+  overlayHideHandler() {
+    this.setState({ credentialsMessage: '' });
   },
 });
 
--- a/lib/layout.jsx
+++ b/lib/layout.jsx
@@ -9,6 +9,13 @@ let format      = require('./format');
 // time before expiration at which we warn
 let EXPIRY_WARNING = 5 * 60 * 1000;
 
+/** Log-in open a authentication URL */
+function signIn() {
+  window.open(auth.buildLoginURL(), '_blank');
+}
+
+exports.signIn = signIn;
+
 /** Navigation bar for layout.jade */
 let Navigation = React.createClass({
   /** Get initial state */
@@ -20,10 +27,6 @@ let Navigation = React.createClass({
     };
   },
 
-  /** Log-in open a authentication URL */
-  signIn() {
-    window.open(auth.buildLoginURL(), '_blank');
-  },
 
   /** Log out (clear credentials) */
   signOut() {
@@ -61,7 +64,7 @@ let Navigation = React.createClass({
 
     // Reload credentials
     this.setState({
-      credentials: credentials,
+      credentials,
       credentialsExpiringSoon: false,
       credentialsMessage: credentials? {
         title: "Signed In",
@@ -183,9 +186,11 @@ let Navigation = React.createClass({
   renderCredentialsMenu() {
     // if there are no credentials at all, then there is no menu -- just a sign-in link
     if (!this.state.credentials) {
-      return <bs.NavItem onSelect={this.signIn} ref="credentials">
-        <bs.Glyphicon glyph="log-in"/>&nbsp;Sign in
-      </bs.NavItem>
+      return (
+        <bs.NavItem onSelect={signIn} ref="credentials">
+          <bs.Glyphicon glyph="log-in"/>&nbsp;Sign in
+        </bs.NavItem>
+      );
     }
 
     // TODO: color this according to time until expiry
@@ -193,20 +198,22 @@ let Navigation = React.createClass({
     let className = this.state.credentialsExpiringSoon? 'text-warning' : '';
     let menuHeading = <span><bs.Glyphicon className={className} glyph={glyph}/>&nbsp;
                           {this.state.credentials.clientId}</span>;
-    return <bs.NavDropdown key={2} title={menuHeading} ref="credentials" id="credentials">
-      <bs.MenuItem href="/credentials">
-        <format.Icon name="key"/>&nbsp;Manage Credentials
-      </bs.MenuItem>
-      <bs.MenuItem divider />
-      <bs.NavItem onSelect={this.signIn}>
-        <bs.Glyphicon glyph="log-in"/>&nbsp;
-        Sign In
-      </bs.NavItem>
-      <bs.NavItem onSelect={this.signOut}>
-        <bs.Glyphicon glyph="log-out"/>&nbsp;
-        Sign Out
-      </bs.NavItem>
-    </bs.NavDropdown>;
+    return (
+      <bs.NavDropdown key={2} title={menuHeading} ref="credentials" id="credentials">
+        <bs.MenuItem href="/credentials">
+          <format.Icon name="key"/>&nbsp;Manage Credentials
+        </bs.MenuItem>
+        <bs.MenuItem divider />
+        <bs.NavItem onSelect={signIn}>
+          <bs.Glyphicon glyph="log-in"/>&nbsp;
+          Sign In
+        </bs.NavItem>
+        <bs.NavItem onSelect={this.signOut}>
+          <bs.Glyphicon glyph="log-out"/>&nbsp;
+          Sign Out
+        </bs.NavItem>
+      </bs.NavDropdown>
+    );
   },
 
   renderCredentialsPopover() {
@@ -214,14 +221,16 @@ let Navigation = React.createClass({
       let popover = <bs.Popover placement="bottom" id="signin-alert" title={this.state.credentialsMessage.title}>
         {this.state.credentialsMessage.body}
       </bs.Popover>;
-      return <bs.Overlay
-              show={true}
-              rootClose={true}
-              onHide={ () => this.setState({credentialsMessage: undefined}) }
-              placement="bottom"
-              target={ props => ReactDOM.findDOMNode(this.refs.credentials) }>
-        {popover}
-      </bs.Overlay>
+      return (
+        <bs.Overlay
+          show={true}
+          rootClose={true}
+          onHide={ () => this.setState({credentialsMessage: undefined}) }
+          placement="bottom"
+          target={ props => ReactDOM.findDOMNode(this.refs.credentials) }>
+          {popover}
+        </bs.Overlay>
+      );
     }
   },
 });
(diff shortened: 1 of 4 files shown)


djmitche:
Conceptually, r+ from me (I didn't run it or look line-by-line at the code)

andreadelrio:
@eliperelman do you have any suggestions as to how to hide the message box (or change its content) once the user has logged in? I was thinking about using layout's handleCredentialsChanged.

lib/layout.jsx
```diff
@@ -9,6 +9,13 @@ let format      = require('./format');
 // time before expiration at which we warn
 let EXPIRY_WARNING = 5 * 60 * 1000;
 
+/** Log-in open a authentication URL */
+function signIn() {
+  window.open(auth.buildLoginURL(), '_blank');
+}
+
+exports.signIn = signIn.bind(this);
```
eliperelman: `signIn` doesn't use `this`, so not sure why it needs to be bound.

lib/layout.jsx
```diff
@@ -58,7 +61,6 @@ let Navigation = React.createClass({
   /** Credentials changed */
   handleCredentialsChanged(e) {
     let credentials = auth.loadCredentials();
-
```
eliperelman: Keep this line, don't remove.

lib/layout.jsx
```diff
@@ -58,7 +61,6 @@ let Navigation = React.createClass({
...
     // Reload credentials
     this.setState({
       credentials: credentials,
```
eliperelman:
```js
this.setState({
  credentials,
```

lib/layout.jsx
```diff
@@ -233,4 +235,4 @@ $(function() {
     <Navigation/>,
     $('#navbar')[0]
   );
-});
+});
```
eliperelman: Keep a blank line at the end of the file.

lib/layout.less
```diff
@@ -16,3 +16,8 @@
 #navbar nav.navbar-inverse .navbar-text {
   color: #fff;
 }
+
+.pointer {
+	cursor: pointer;
```
eliperelman: Indent using 2 spaces.

lib/layout.less
```diff
@@ -16,3 +16,8 @@
...
+	text-decoration: underline;
+}
```
eliperelman: Keep blank line at end of file.

lib/utils.jsx
```diff
@@ -280,23 +282,36 @@ var createTaskClusterMixin = (options) => {
...
+
     /**
      * Error object, assumed to have message and possible properties from
      * taskcluster-client
      */
     renderError(err) {
       // Find some sort of summary or error code
-      let code = err.code + ' Error!';
+      let code = changeCase.titleCase(err.code.replace('Error','')) + ' Error!';
       if (!err.code && err.statusCode) {
```
eliperelman: Here we say `!err.code` but the previous line uses it regardless. It is either there or isn't and should be guarded against.

lib/utils.jsx
```diff
@@ -280,23 +282,36 @@ var createTaskClusterMixin = (options) => {
...
+    },
+
     /**
      * Error object, assumed to have message and possible properties from
      * taskcluster-client
      */
     renderError(err) {
       // Find some sort of summary or error code
-      let code = err.code + ' Error!';
+      let code = changeCase.titleCase(err.code.replace('Error','')) + ' Error!';
```
eliperelman: Insert line between last declaration and conditional.

lib/utils.jsx
```diff
@@ -280,23 +282,36 @@ var createTaskClusterMixin = (options) => {
...
       if (!err.code && err.statusCode) {
         code = 'HTTP ' + err.statusCode;
       }
       code = code || 'Unknown Error';
-
+      var action = err.body['requestInfo']['method'];
```
eliperelman: `err.body.requestInfo.method`

lib/utils.jsx
```diff
@@ -280,23 +282,36 @@ var createTaskClusterMixin = (options) => {
...
+      var loggedOut403 = !auth.hasCredentials() && err.statusCode == "403";
```
eliperelman: `err.statusCode === '403'

lib/utils.jsx
```diff
@@ -280,23 +282,36 @@ var createTaskClusterMixin = (options) => {
...
       // Find some sort of message
-      let message = err.message || '```\n' + err.stack + '\n```';
+      let message = err.message.slice(0, err.message.search('----')) || '```\n' + err.stack + '\n```';
+
+      let title = <bs.Button>Additional details...</bs.Button>;
```
eliperelman: Prefer `const`.

lib/layout.jsx
```diff
@@ -58,10 +61,10 @@ let Navigation = React.createClass({
   /** Credentials changed */
   handleCredentialsChanged(e) {
     let credentials = auth.loadCredentials();
-
+
```
eliperelman: This should be a blank line with no indentation.

lib/utils.jsx
```diff
@@ -280,23 +282,41 @@ var createTaskClusterMixin = (options) => {
...
+
     /**
      * Error object, assumed to have message and possible properties from
      * taskcluster-client
      */
     renderError(err) {
       // Find some sort of summary or error code
-      let code = err.code + ' Error!';
-      if (!err.code && err.statusCode) {
+      let code;
```
eliperelman:
```js
const code = err.code ?
  `${changeCase.titleCase(err.code.replace('Error', ''))} Error!` :
  `HTTP ${err.statusCode}`;
```

lib/utils.jsx
```diff
@@ -280,23 +282,41 @@ var createTaskClusterMixin = (options) => {
...
+
+      if (err.code) {
+        code = changeCase.titleCase(err.code.replace('Error','')) + ' Error!';
+      } else if (!err.code && err.statusCode) {
         code = 'HTTP ' + err.statusCode;
       }
+      
       code = code || 'Unknown Error';
```
eliperelman: This line is impossible to evalulate falsy. Remove this line.

lib/utils.jsx
```diff
@@ -5,10 +5,12 @@ var Promise       = require('promise');
 var debug         = require('debug')('lib:utils');
 var assert        = require('assert');
 var format        = require('./format');
+var layout        = require('./layout');
```
eliperelman: Let's make this:

```js
 var { signIn } = require('./layout');
```

Then below you won't need to define another `signIn`, method, just call it directly.

lib/utils.jsx
```diff
@@ -280,23 +282,40 @@ var createTaskClusterMixin = (options) => {
...
-      let message = err.message || '```\n' + err.stack + '\n```';
+      const message = err.message.slice(0, err.message.search('----')) || '```\n' + err.stack + '\n```';
+      const title = <bs.Button>Additional details...</bs.Button>;
 
-      let title = <bs.Button bsStyle="link">Additional details...</bs.Button>;
       return (
+        loggedOut403 ? (
+        <bs.Alert bsStyle="info"> 
+          <strong>{code}&nbsp;</strong>
+          <p>Please <a className="pointer" onClick={this.signIn}>sign in</a> and <strong>{changeCase.titleCase(action)}</strong> again</p>
```
eliperelman: Once you bring in just the layout `signIn`, this can just be `onClick={signIn}`.

lib/utils.jsx
```diff
@@ -280,23 +282,40 @@ var createTaskClusterMixin = (options) => {
       );
     },
 
+    signIn() {
```
eliperelman: Bringing in the `signIn` method, you can remove this declaration.

lib/utils.jsx
```diff
@@ -280,23 +282,40 @@ var createTaskClusterMixin = (options) => {
...
-      code = code || 'Unknown Error';
+
+      var action = err.body.requestInfo.method;
+
+      // Find if user is logged out and error code is 403
+      var loggedOut403 = !auth.hasCredentials() && err.statusCode === 403;
 
       // Find some sort of message
-      let message = err.message || '```\n' + err.stack + '\n```';
+      const message = err.message.slice(0, err.message.search('----')) || '```\n' + err.stack + '\n```';
```
eliperelman: In what case would the first expression evaluate falsy? If there is no `message` property, then this will throw. If it's an empty string, then we are just wasting cycles on this. Maybe try this:

```js
const message = err.message ?
  err.message.slice(0, err.message.search('----')) :
  '```\n' + err.stack + '\n```';
```

andreadelrio:
I've noticed something is happening. When I'm logged out and try to create a task I get the message box asking me to log in, when I do so I'm taken back to the 'create task' page still showing the message box. Then, if I type anything else on the 'create task' box (for example, change "workerType" from "tutorial" to "tutorials") I immediately get a new error message even before clicking the 'Create Task' button. How can I prevent this from happening?

djmitche:
What's the new error message?

eliperelman:
I'm currently trying to track down the issue in your branch. I see the same issue as well:

screenshot

eliperelman:
I think when the task attempts to create, and the sign in is shown, the error state of the taskcreator component is never reset when the sign in is completed. I think that if you reset the error state upon successful sign in that this error will go away when editing.

lib/layout.jsx
```diff
@@ -183,7 +186,7 @@ let Navigation = React.createClass({
   renderCredentialsMenu() {
     // if there are no credentials at all, then there is no menu -- just a sign-in link
     if (!this.state.credentials) {
-      return <bs.NavItem onSelect={this.signIn} ref="credentials">
+      return <bs.NavItem onSelect={signIn} ref="credentials">
```
eliperelman: Wrap multiline JSX in parens:

```jsx
return (
  <bs.NavItem onSelect={signIn} ref="credentials">
    <bs.Glyphicon glyph="log-in" /> Sign in
  </bs.NavItem>
);
```

walac:
@andreadelrio travis is failing because a package `change-case` is *required* but package.json doesn't have it.

eliperelman:
Hmm odd, since the change to package.json was in the commits previously.

andreadelrio:
@walac something went wrong when I squashed my commits but I already fixed package.json

andreadelrio:
I added some code to cover the case of One-Click Loaner, as it was just brought up on the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1248977).

lib/layout.jsx
```diff
@@ -183,45 +186,51 @@ let Navigation = React.createClass({
...
-              onHide={ () => this.setState({credentialsMessage: undefined}) }
-              placement="bottom"
-              target={ props => ReactDOM.findDOMNode(this.refs.credentials) }>
-        {popover}
-      </bs.Overlay>
+      return (
+        <bs.Overlay
+          show={true}
+          rootClose={true}
+          onHide={ () => this.setState({credentialsMessage: undefined}) }
```
eliperelman: This should probably be moved out into a handler, e.g. `overlayHideHandler`. So something like the following:

```jsx
onHide={this.overlayHideHandler}
```

```js
overlayHideHandler() {
  this.setState({ credentialsMessage: '' });
}
```

Avoid using `undefined` directly, prefer `null` or probably an empty string since this is a message.

lib/layout.jsx
```diff
@@ -183,45 +186,51 @@ let Navigation = React.createClass({
...
+          placement="bottom"
+          target={ props => ReactDOM.findDOMNode(this.refs.credentials) }>
```
eliperelman: Remove spaces from React interpolation braces.

```jsx
target={props => ReactDOM.findDOMNode(this.refs.credentials)}
```

lib/utils.jsx
```diff
@@ -141,7 +143,8 @@ var createTaskClusterMixin = (options) => {
     handleCredentialsChanged(e) {
       // Update clients with new credentials
       this._createClients(e.detail);
-
+      
+      this.setState({createdTaskIdError: undefined});
```
eliperelman:
```js
this.setState({ createdTaskIdError: null });
```

lib/utils.jsx
```diff
@@ -286,17 +289,36 @@ var createTaskClusterMixin = (options) => {
...
-        code = 'HTTP ' + err.statusCode;
+      let code = 'Unknown Error';
+
+      if (err.code) {
+        code = `${changeCase.titleCase(err.code.replace('Error', ''))} Error!`;
+      } else if (err.statusCode) {
+        code = `HTTP ${err.statusCode}`;
+      }
+
+      var action = changeCase.titleCase(err.body.requestInfo.method);
```
eliperelman: This sure seems like a lot of logic just to get a specific message for getting the user to log in. Let's simplify this.

If it's a 403, just show the message: `Please sign in and try again`. Otherwise this method will continue to grow as specific actions need to be handled. Also remove the anchor. They are being told to sign in, they can use the existing link in the navigation.

lib/layout.less
```diff
@@ -16,3 +16,8 @@
 #navbar nav.navbar-inverse .navbar-text {
   color: #fff;
 }
+
+.pointer {
```
eliperelman: Remove this since the error message is going to be simplified and will not need an anchor.

eliperelman:
I'm going to backpedal on this one. I'd rather keep the error message for the un-authenticated case simple by just telling them there was an error involving authentication, they should sign in, and try again. This will negate the need for exposing sign in everywhere, and probably has repercussions with this PR. See if you can boil this down to just showing the proper message on authentication-failed requests.

lib/utils.jsx
```diff
@@ -286,17 +288,29 @@ var createTaskClusterMixin = (options) => {
...
+      const message = err.message ?
+        err.message.slice(0, err.message.search('----')) :
+        '```\n' + err.stack + '\n```';
+      const title = <bs.Button>Additional details...</bs.Button>;
 
-      let title = <bs.Button bsStyle="link">Additional details...</bs.Button>;
       return (
+        loggedOut403 ? (
+        <bs.Alert bsStyle="info"> 
+          <p>We need to authenticate you first. Please sign in and try again.</p>
```
eliperelman: How about "You are not authorized to perform the requested action. Please sign in with an authorized user and try again."?

eliperelman: I say this because it's possible that the user is already signed in, but just doesn't have valid scopes for performing the action.

andreadelrio: @eliperelman but we are only showing that message to users that are logged out.

eliperelman: Ah right, `!auth.hasCredentials()`.

eliperelman: OK, I'd change the wording to "You are not authorized to perform the requested action. Please sign in and try again."

PR #123, opened 2016-07-26
analytics

jonasfj:
Okay, I didn't test this... just saw it was missing...

If you don't have access to google analystics, hit me up and I'll find the bug and we can file for you to get access too...

We should totally send events for a few of buttons etc... I for one would like to know how often people using things like the one-click-loaner feature..


eliperelman:
r+

PR #125, opened 2016-07-27
Fixed some weird docker-worker specific validation

jonasfj:
I don't think any validation makes sense... Except maybe validate that there is a task definition, and that the property isn't `null`...

@imbstack, git says you wrote this... I have to ask, any idea why it does what it does?

### DIFF
--- a/lib/ui/retrigger-button.jsx
+++ b/lib/ui/retrigger-button.jsx
@@ -32,7 +32,7 @@ let RetriggerButton = React.createClass({
 
   valid() {
     // Simple sanity check
-    return this.props.task && this.props.task.payload;
+    return !!(this.props.task && this.props.task.payload);
   },
 
   render() {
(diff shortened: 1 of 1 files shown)


jonasfj:
@imbstack, mostly I'm curious if we've perhaps repeated this "validation" pattern more than just here?

lib/ui/retrigger-button.jsx
```diff
@@ -31,17 +31,8 @@ let RetriggerButton = React.createClass({
...
-    }
-    if (!(payload.command instanceof Array)) {
-      return false;
-    }
-    if (typeof(payload.maxRunTime) !== 'number') {
-      return false;
-    }
-    return true;
+    // Simple sanity check
+    return this.props.task && this.props.task.payload;
```
eliperelman: Not sure if there are any boolean dependents on this, but let's coerce to be safe:

```js
return !!(this.props.task && this.props.task.payload);
```

eliperelman:
Granted with nit fixed.

PR #127, opened 2016-08-05
make progressbar onclick work on text i.e. the label e.g, C(129)

helfi92:
Clicking on the label of a progress bar was not filtering the list of tasks. This is important to have when you have a very small section in the progress bar and the label is occupying most of the space.

### DIFF
--- a/push-inspector/components/progressBar.jsx
+++ b/push-inspector/components/progressBar.jsx
@@ -126,7 +126,7 @@ class ProgressBar extends Component {
     };
 
     return (
-      <span onClick={this.progressBarClicked}>
+      <div onClick={this.progressBarClicked}>
         <bs.ProgressBar className="progressBar">
           {groups.map((group, index) => {
             const subtasks = this[group];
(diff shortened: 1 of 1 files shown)


push-inspector/components/progressBar.jsx
```diff
@@ -126,23 +126,25 @@ class ProgressBar extends Component {
...
-            <bs.ProgressBar
-              key={index}
-              title={title}
-              className={className}
-              now={getWeightedPercent(percents[index])}
-              label={label} />
-          );
-        })}
-      </bs.ProgressBar>
+      <span onClick={this.progressBarClicked}>
```
eliperelman: I try not to wrap block-level elements with inline elements. I'm pretty sure there's a div somewhere in this progress bar, so let's switch from `<span>` to `<div>`.

eliperelman:
Just one thing to change, then I'll merge it.

helfi92:
good call. i agree :)

PR #128, opened 2016-08-10
Push Inspector - fix bug when large number of tasks in a taskgroupid

helfi92:
This bug was spotted when handling a large amount of tasks in a taskGroupId. You can see the problem by inspecting the taskGroupId: **cfMDvTzETHagYx-uSPCOxQ**

The loaded list of tasks were behaving weirdly, shifting the progress bar states left and right.


eliperelman:
I'll land once the build is green.

PR #129, opened 2016-08-15
Bug 1295025: merge uptimerobot status into tools/status, make it better

djmitche:
The main goal here is to put the output of status.taskcluster.net on tools.taskcluster.net/status.  However, this does a bunch of other cleanup of the status page as well.


status/status.jsx
```diff
@@ -23,10 +24,16 @@ function makeRequest(options, allowHeaders = []) {
...
 }
 
-function pollTaskclusterService(url, cb) {
-  $.getJSON(url)
-    .done(data => cb(data.alive))
-    .fail(err => cb(false));
+function pollTaskclusterService(key, cb) {
+  return $.getJSON(`http://api.uptimerobot.com/getMonitors?apiKey=${key}&format=json&noJsonCallback=1`)
+    .done(res => {
+      var monitor = res.monitors.monitor[0];
```
eliperelman:
```js
const [ monitor ] = res.monitors.monitor;
```

status/status.jsx
```diff
@@ -23,10 +24,16 @@ function makeRequest(options, allowHeaders = []) {
...
+      cb((monitor.status === "2") ? "up" : "down");  // 2 is "up"
```
eliperelman: Remove parens around ternary condition. Switch all double quotes to single quotes.

eliperelman: Also, add a line between variable declaration and callback expression.

status/status.jsx
```diff
@@ -23,10 +24,16 @@ function makeRequest(options, allowHeaders = []) {
...
+    }).fail(err => {
+      console.log("Error fetching data from uptimerobot");
```
eliperelman: Let's also make this a `console.error`. Switch to single quotes.

status/status.jsx
```diff
@@ -23,10 +24,16 @@ function makeRequest(options, allowHeaders = []) {
...
+      console.error(err);
+      cb("error");
```
eliperelman: Any reason why the callback takes a string and not the error itself.

status/status.jsx
```diff
@@ -73,15 +98,15 @@ let otherServices = [
...
         let items = data.getElementsByTagName('item');
         if (!items.length) {
-          cb(true);
+          cb("up");
           return;
         }
 
         let title = items[0].getElementsByTagName('title');
-        cb(title[0].innerHTML.startsWith('Service is operating normally'));
+        cb((title[0].innerHTML.startsWith('Service is operating normally'))? "up" : "down");
```
eliperelman: Remove parens from around ternary condition. Insert space before `?`.

status/status.jsx
```diff
@@ -73,15 +98,15 @@ let otherServices = [
...
       } catch (err) {
         console.log(err.stack || err);
-        cb(false);
+        cb("down");
```
eliperelman: Single quotes.

status/status.jsx
```diff
@@ -141,23 +166,36 @@ let otherServices = [
...
       .catch(err => {
         console.log(err || err.stack);
-        cb(false);
+        cb("err");
       });
     }
   }
 ];
 
+var STATUS_DISPLAY = {
```
eliperelman: s/var/const/

status/status.jsx
```diff
@@ -141,23 +166,36 @@ let otherServices = [
...
+  'loading': {'icon': 'spinner', 'spin': true, 'color': 'gray'},
```
eliperelman: Remove quotes from property names.

eliperelman: Insert spaces after opening curly brace and before closing curly brace.

status/status.jsx
```diff
@@ -141,23 +166,36 @@ let otherServices = [
...
+};
+
 export let StatusChecker = React.createClass({
   propTypes: {
-    up: React.PropTypes.bool.isRequired
+    status: React.PropTypes.string.isRequired
   },
   render: function() {
-    let image = this.props.up ? "green-check.png" : "red-check.png";
+    let {icon, spin, color} = STATUS_DISPLAY[this.props.status] || STATUS_DISPLAY['err'];
```
eliperelman: Is `this.props.status` potentially `null` or `undefined`, or rather that the key may be unknown again `STATUS_DISPLAY`?

eliperelman: s/let/const/

eliperelman: Insert spaces inside destructuring braces.

status/status.jsx
```diff
@@ -141,23 +166,36 @@ let otherServices = [
...
     return (
-      <img className="img-check img-fluid center-block" src={image}/>
+        <format.Icon
+          name={icon}
+          size="lg"
+          spin={spin}
+          className="pull-left"
+          style={{'color': color}} />
```
eliperelman:
```jsx
style={{ color }} />
```

status/status.jsx
```diff
@@ -169,22 +207,36 @@ export let Service = React.createClass({
...
+
+  componentWillUnmount() {
+    if (this.timer) {
+      clearTimeout(this.timer);
+    }
+  },
+
+  poll() {
+    this.props.poll(status => {
+      this.setState({status});
```
eliperelman:
```js
this setState({ status });
```

status/status.jsx
```diff
@@ -169,22 +207,36 @@ export let Service = React.createClass({
...
+      this.timer = setTimeout(this.poll, 5000);
+    });
+  },
+
+  getInitialState() {
     return {
-      up: true,
-      intervalId: intervalId
+      status: "loading",
```
eliperelman: Single quotes.

status/status.jsx
```diff
@@ -169,22 +207,36 @@ export let Service = React.createClass({
...
     };
   },
+
   render: function() {
     return (
       <div className="form-horizontal service-status-container">
         <a target="_blank" href={this.props.link}>
           <div data-tip data-for={this.props.name}>
-            <label className="service-status"><StatusChecker up={this.state.up}/></label>
+            <label className="service-status"><StatusChecker status={this.state.status}/></label>
```
eliperelman: nit: space before tag closing of self-closing tags.

status/status.jsx
```diff
@@ -243,11 +280,12 @@ export let TaskclusterStatus = React.createClass({
...
   render: function() {
     return (
-      <bs.Row>
-        <TaskclusterStatus/>
-        <ServiceGroup name="Taskcluster" services={taskclusterServices} description="Taskcluster services"/>
-        <ServiceGroup name="Services" services={otherServices} description="External services Taskcluster depends on"/>
-      </bs.Row>
+      <div>
+        <bs.Row>
+          <ServiceGroup name="Taskcluster Services" services={taskclusterServices} description="Taskcluster services"/>
```
eliperelman: nit: insert space before tag closing on self-closing tags.

status/status.jsx
```diff
@@ -36,27 +43,45 @@ function dummyPoll(cb) {
 let taskclusterServices = [
   {
     name: "Queue",
-    poll: pollTaskclusterService.bind(null, "https://queue.taskcluster.net/v1/ping"),
+    poll: pollTaskclusterService.bind(null, "m776323830-a170e7abc854f94cc2f4c078"),
```
eliperelman: Is this an API key? Any reason why we are hard-coding this in source?

eliperelman: If it is, let's move this to the `envify` section of `Gruntfile.js`.

imbstack:
I'll leave the nitty gritty to @eliperelman who clearly has it covered, but I was running this locally and noticed that I think whatever logic we're using to determine the status of the external services isn't correct. All three of them appear to be up right now but are red on my local copy of this. Not sure if that's something wrong with me or running it locally or if the logic is just weird somehow. It doesn't necessarily stop this from shipping as it's not the purpose of this change, but definitely something we should look into.

status/status.jsx
djmitche: It translates to the error appearance for the status check

status/status.jsx
djmitche: either one -- I'm just trying to be safe

status/status.jsx
djmitche: Too many python dictionaries for bug 1286075..

status/status.jsx
djmitche: Because of the way the data structure works -- the poll method doesn't get
a reference to the data structure, so there's nowhere else to stick it.

djmitche: What's envify??

djmitche:
The three external services appear to be down because your localhost isn't
whitelisted by the cors proxy.  But it's a nice chance to see the "down"
and "error" results :)

status/status.jsx
eliperelman: Something like this:

https://github.com/taskcluster/taskcluster-tools/blob/master/Gruntfile.js#L216

https://github.com/taskcluster/taskcluster-tools/blob/master/status/status.jsx#L18

status/status.jsx
djmitche: But, it should have been `cb('err')`.

status/status.jsx
djmitche: I don't think this makes sense -- those API keys correspond to the sites they're monitoring, so they should travel together.  Tools isn't a config-driven site, so I don't think we should move the entire list of things to monitor into the config file (is there even a config file? how do env vars work for a static site on S3?).  Typical usage of uptimerobot is to include the API keys directly in the list of monitors to be displayed, more-or-less as I've done here.  In fact, that's what https://github.com/taskcluster/taskcluster-status-deprecated does.

eliperelman: > Tools isn't a config-driven site

Yet. It will be more so as things evolve. I don't prefer having env vars in a config file at all, as I'd prefer them to be in the environment (and not in source control), but when that isn't possible, we should be using the config file, which I posted in the previous comment:

https://github.com/taskcluster/taskcluster-tools/blob/master/Gruntfile.js#L216

If several env/config vars need to travel together, perhaps they should be wrapped in some sort of component that will handle that travel. As I see it now, I still think they should be in config.

status/status.jsx
```diff
@@ -23,10 +24,17 @@ function makeRequest(options, allowHeaders = []) {
...
-    .done(data => cb(data.alive))
-    .fail(err => cb(false));
+function pollTaskclusterService(key, cb) {
+  return $.getJSON(`http://api.uptimerobot.com/getMonitors?apiKey=${key}&format=json&noJsonCallback=1`)
+    .done(res => {
+      const [ monitor ] = res.monitors.monitor;
+
+      cb(monitor.status === '2' ? 'up' : 'down');  // 2 is 'up'
+    }).fail(err => {
+      console.error("Error fetching data from uptimerobot");
```
eliperelman: Single quotes.

status/status.jsx
```diff
@@ -73,15 +99,15 @@ let otherServices = [
...
         let items = data.getElementsByTagName('item');
         if (!items.length) {
-          cb(true);
+          cb('up');
           return;
         }
 
         let title = items[0].getElementsByTagName('title');
-        cb(title[0].innerHTML.startsWith('Service is operating normally'));
+        cb(title[0].innerHTML.startsWith('Service is operating normally') ? "up" : "down");
```
eliperelman: Single quotes.

status/status.jsx
```diff
@@ -141,23 +167,36 @@ let otherServices = [
       Promise.resolve(makeRequest({
         url: 'https://status.heroku.com/feed'
       }))
-      .then(data => cb(!data.length || data[0].title.startsWith('Resolved')))
+      .then(data => cb((!data.length || data[0].title.startsWith('Resolved')) ? "up" : "down"))
```
eliperelman: Single quotes.

status/status.jsx
```diff
@@ -141,23 +167,36 @@ let otherServices = [
...
         console.log(err || err.stack);
-        cb(false);
+        cb('err');
       });
     }
   }
 ];
 
+const STATUS_DISPLAY = {
+  'loading': {icon: 'spinner', spin: true, color: 'gray'},
```
eliperelman: Insert spaces inside curly braces. Remove quotes around property names.

status/status.jsx
```diff
@@ -141,23 +167,36 @@ let otherServices = [
...
+};
+
 export let StatusChecker = React.createClass({
   propTypes: {
-    up: React.PropTypes.bool.isRequired
+    status: React.PropTypes.string.isRequired
   },
   render: function() {
-    let image = this.props.up ? "green-check.png" : "red-check.png";
+    let { icon, spin, color } = STATUS_DISPLAY[this.props.status] || STATUS_DISPLAY['err'];
```
eliperelman: s/let/const/

status/status.jsx
```diff
@@ -141,23 +167,36 @@ let otherServices = [
...
     return (
-      <img className="img-check img-fluid center-block" src={image}/>
+        <format.Icon
+          name={icon}
+          size="lg"
+          spin={spin}
+          className="pull-left"
+          style={{color}} />
```
eliperelman: Insert spaces inside object curly braces.

```jsx
style={{ color }} />
```

status/status.jsx
eliperelman: Still need single quotes here.

status/status.jsx
eliperelman: Still pending this change.

eliperelman:
Just a few more changes and should be ready to land.

PR #131, opened 2016-08-18
Added scope-inspector for exploring scopes, reverse lookup

jonasfj:
This is ugly code and maybe it could be better structured... But it solves a problem, and is confined to a sub-folder. If we one day decide this code is unmaintainable, we just delete the subfolder.

Please don't run eli-lint on this :)

### DIFF
--- a/auth/scopes/app.jsx
+++ b/auth/scopes/app.jsx
@@ -1,0 +1,18 @@
+let React               = require('react');
+let ReactDOM            = require('react-dom');
+let ScopeInspector      = require('./scopeinspector');
+let $                   = require('jquery');
+
+var utils               = require('../../lib/utils');
+
+var hashManager = utils.createHashManager({
+  separator:            '/'
+});
+
+// Render component
+$(function() {
+  ReactDOM.render(
+    <ScopeInspector hashEntry={hashManager.root()}/>,
+    $('#container')[0]
+  );
+});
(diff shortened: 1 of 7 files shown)


jonasfj:
Eli, being PTO, I move this to bstack...

imbstack:
This seems... Good to me. r+

I leave the eli-lint for later!

djmitche:
I just ran it without really looking at the code :)

I agree this is a useful tool.  In terms of UI, I don't like that one list is replaced with a very similar-looking list.  I think I'd rather have the search bar stay at the top and maybe use some kind of autocomplete.  Also, I have no idea what the "Match" options mean, and if I don't understand it then most people won't!

djmitche:
(I understand what the Match options are there for -- I just couldn't figure out what each choice means)

djmitche:
Whoops, I was five minutes late!  Please do consider those comments.

I do think this is really cool.  It also suggests the ability to make assertions about the results of some of these queries -- something I'd like to see someday.

jonasfj:
He he, yeah, I'm not entirely sure how this tool should work...

Definitely we need some explanations in there... Perhaps, some tooltips
will help...

Note: I see this as a generic tool that we probably use to develop some
more process oriented tools.
The goal I'm moving towards is people being able to review scopes that they
manage...

Den 18. aug. 2016 2.49 PM skrev "Dustin J. Mitchell" <
notifications@github.com>:

PR #134, opened 2016-09-01
Add purge-cache viewer

imbstack:
This is a quick hack to view the currently active purge-cache requests. At some point I'd like to add a create-purge-cache-request function on this page, but don't really want to spend too much time on this right now.


purge-caches/cachemanager.jsx
```diff
@@ -0,0 +1,90 @@
...
+      clients: {
+        purgeCaches:       taskcluster.PurgeCache
+      }
+    }),
+  ],
+
+  /** Create an initial state */
+  getInitialState() {
+    return {
+      caches: undefined,
```
eliperelman: I typically avoid all usage of `undefined` in JS. It's more succinct to imply caches contains an empty value, or `null`.

purge-caches/cachemanager.jsx
```diff
@@ -0,0 +1,90 @@
...
+      cachesLoaded: false,
+    };
+  },
+
+  load() {
+    return {
+      caches: this.purgeCaches.allPurgeRequests().then((resp) => resp.requests)
```
eliperelman: So `caches` is supposed to contain a Promise? Or should this be `await`ed?

purge-caches/cachemanager.jsx
```diff
@@ -0,0 +1,90 @@
...
+    };
+  },
+
+  /** Render the main layout of the cache manager page */
+  render() {
+    try {
+      return (
+        <bs.Row>
+          <bs.Col md={8}>
+            <p >All currently ative cache purges are displayed below. 24 hours after creation, requests expire and are no longer displayed here. The <strong>before</strong> column is the time at which any caches that match the previous three classifiers are considered invalid. Any caches created after that time are fine.</p>
```
eliperelman: Remove extra space in `p` tag.

purge-caches/cachemanager.jsx
```diff
@@ -0,0 +1,90 @@
...
+        <thead>
+          <tr>
+            <th>Provisioner ID</th>
+            <th>Worker Type</th>
+            <th>Cache Name</th>
+            <th>Before</th>
+          </tr>
+        </thead>
+        <tbody>
+             {this.state.caches.map(this.renderCacheRow)}
```
eliperelman: Dedent this line 2 spaces.

purge-caches/cachemanager.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+  load() {
+    return {
+      caches: this.purgeCaches.allPurgeRequests().then((resp) => resp.requests)
+    };
+  },
+
+  /** Render the main layout of the cache manager page */
+  render() {
+    try {
```
eliperelman: This whole section is in a `try`? If there is going to be a problem, maybe just wrap the `renderCachesTable()` section in a try so its not completely blank.

purge-caches/cachemanager.jsx
eliperelman: If `caches` should be a promise, maybe the initial value should be `Promise.resolve()` or `Promise.reject()`.

purge-caches/cachemanager.jsx
jonasfj: this is horrible and intended... :)

purge-caches/cachemanager.jsx
jonasfj: this is now the horrible mixin works...

purge-caches/cachemanager.jsx
eliperelman: 😆

jonasfj:
# `r+`

Ship it, if it works...

purge-caches/cachemanager.jsx
eliperelman: Ah ok, mixin requirement, fair enough. I just couldn't determine if this was supposed to contain an actual value or an eventual one.

eliperelman:
Everything seems to be in order here, just a few nits.

purge-caches/cachemanager.jsx
jonasfj: It is supposed to contain a promise... load() returns a map from keys to
promises.. the taskcluster mixin will convert these to state properties and
take care of the awaiting...

Den 1. sep. 2016 5.32 PM skrev "Eli Perelman" <notifications@github.com>:

PR #135, opened 2016-09-06
Fix bug 1299599 - race condition between listing artifacts and listen…

jonasfj:
…ing for events

Code is ugly, just r+ and pretend you didn't see this --- it works for me :)


eliperelman:
*closes eyes*

PR #140, opened 2016-09-24
Mega refactor for new build process

eliperelman:


djmitche:
Wow, github won't even show the diff.

Do you have some guidance for us on reviewing this?

eliperelman:
Yep, I will soon, have to make a couple updates to this and then I'll open for review. 

This is mostly changes to the build process, and with that comes ESLint. So I moved all the source to src/ and fixed all lint errors. There shouldn't be much in the way of functional changes, just linting updates.

walac:
Files changed: 504
I will never ever complain about @jonasfj's PRs size again

eliperelman:
@walac 507 now. Just for you.

walac:
Poor reviewer.

jonasfj:
@eliperelman, please change to use `'^x.y.z'` for versions in `package.json`..
Then let's use an `npm-shrinkwrap.json` to lock versions.

That way all indirect dependencies are locked too, and upgrading versions is simple.
Additionally, if we decide to not upgrade a version then it's a conscience decision to not use `^` for that specific dependency.

jonasfj:
Artifact list is broken: http://localhost:9000/task-inspector/#YD8TUY3fTVqEPWBkJxqVlQ/0

jonasfj:
Not sure if it's new, but the text has whitespace issues: http://localhost:9000/pulse-inspector/
Right before the link, there is no space between the link and text.

jonasfj:
task-creator shows error even though there isn't any: http://localhost:9000/task-creator/

eliperelman:
@jonasfj OK, I think all your concerns are now fixed. All dependencies should now be ranges along with a shrinkwrap, except for durable-json-lint, which has a long-standing issue created by you. :laughing:

jonasfj:
Awesome. I think it's nice that we're explicit about node upgrading `durable-json-lint`, though we probably could fork and fix it at some point.

--

r+

A huge patch deserves a huge approval stamp!

djmitche:
# r+

This looks great.  I think the only thing I'd like to see improved is the commenting and organization of `tools-preset/index.js` -- especially without any documentation of the contents of presets (I looked at the neutrino readme and the neutrino-preset-base readme..) it's pretty impenetrable and I would find myself unable to modify it if that became necessary.

eliperelman:
I'll update the tools preset with some organization and comments before merging. I'll try to add some comments to neutrino and its presets soon.

djmitche:
Awesome!  That doesn't necessarily need to block landing -- if it's more than an hour or two today, I'd rather see this merged and the comments added when they are done.

I'm being selfish, of course: I have work today that will occur in the tools repo and I'd rather not rebase over this refactor :)

PR #143, opened 2016-09-29
fix bug 1295949 for public artifacts

jonasfj:
https://bugzilla.mozilla.org/show_bug.cgi?id=1295949

### DIFF
--- a/src/lib/ui/artifactlist.jsx
+++ b/src/lib/ui/artifactlist.jsx
@@ -139,27 +139,20 @@ export default React.createClass({
       let icon;
 
       if (/^public\//.test(artifact.name)) {
+        url = 'https://queue.taskcluster.net/v1/task/';
         if (this.props.runId != null) {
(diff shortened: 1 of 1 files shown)


jonasfj:
Maybe at some point we extend the reference format to specify if a URL parameter needs to have slashes encoded... It's kind of an exception that artifacts don't.

src/lib/ui/artifactlist.jsx
```diff
@@ -140,18 +140,26 @@ export default React.createClass({
...
         if (this.props.runId != null) {
+          url = `https://queue.taskcluster.net/v1/task/${this.props.taskId}` +
+                `/runs/${this.props.runId}/artifacts/${artifact.name}`;
+          /*
           url = this.queue.buildUrl(
             this.queue.getArtifact,
             this.props.taskId,
             this.props.runId,
             artifact.name
           );
```
djmitche: *dislike* commented-out code

perhaps just a text comment here indicating that buildUrl generates URLs containing quoted slashes, and we don't want that here?

PR #148, opened 2016-10-03
Make help link open in new tab

imbstack:
I haven't tested this locally, but I don't think it should break anything?


djmitche:
nice branch name, heh

PR #150, opened 2016-10-03
Push Inspector - UI styling

helfi92:
The new theme of taskcluster-tools broke small parts of push-inspector. This is a fix

### DIFF
--- a/src/push-inspector/app.less
+++ b/src/push-inspector/app.less
@@ -23,7 +23,7 @@
 }
 
 .select-wrapper {
-  width: 70%;
+  width: 75%;
   display: inline-flex;
   margin-bottom: 5px;
 }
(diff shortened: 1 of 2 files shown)


eliperelman:
Can you post some screenshots of the changes?

eliperelman:
So, this change was a conscious decision by myself in order to transition away from the UI we are currently presenting here. Right now the push inspector has the table and the task information side-by-side, but it will not stay that way.

Consider the changes in UX to the task inspector, where each concern is placed on a different tab: Task Details, Actions, Runs, where a Run has its own Details, Artifacts, and Logs. The Push Inspector should and will eventually follow suit by replicating the task inspector view, but adding an initial tab for results. Clicking on any of the results will load the Task Details and switch to that tab.

helfi92:
OK, no problem. I will be closing the PR

PR #152, opened 2016-10-05
Editing worker types in the provisioner UI

helfi92:
I don't seem to have sufficient scopes to edit workertypes to fully test it.

Problem:
`this.workerType.workerType` is undefined

Solution:
the map function gives access to `workerType`. No need to use `this`


helfi92:
@imbstack

imbstack:
Looks reasonable to me. We can merge this and see if it fixes @gregarndt's problem

gregarndt:
Yup, appears to have fixed it.  Thanks!

PR #153, opened 2016-10-06
Do all sign-in on the tools site

djmitche:
* Always show a sign-in menu
 * Remove sign in/out from the manage-credentials page, as they are now
   more complex
 * Parameterize the sign-in methods:
   * Okta
   * Persona
   * Manual -- copy/paste creds
   * Devel -- act as 3rd party to the production instance (development only)
 * Implement Persona with the login API's promise-resolution method
 * Implement Okta logins with a GET in a new window that eventually
   updates LocalStorage and closes the window

---

This isn't ready, but I've been fighting with it for a long time, so I need some help:

 * The style changes don't seem to be working; the manual form should look like that on https://login.taskcluster.net, but I don't even see the styles applied in devtools.  it looks like the styles get bundled into common.js, so I can't really tell if Neutrino is even looking at that file.  I tried saving the file a few times to see if Neutrino would pick it up, and it didn't, but that's not much indication -- it often does the same for Layout.js.  What am I missing?

 * I had to work around a React-Bootstrap bug (see comments) where the sign-in menu didn't close after signing in with Persona or Okta.  Is there a better way than to manually control `MenuItem.open`?

 * The manual Modal disappears as soon as I hit a key.  I think this is because the sign-in menu closes, removing the Manual component, which removes the Modal.  I'm trying to nicely encapsulate these sign-in methods, but it's looking like the Modal needs to be stuck somewhere higher up in the HTML hierarchy and thus have its state managed by Navigation instead.  What's the best way to structure this?

 * Console complains about an array of things without keys, but I think I've given a key to everything at the level of `{ signinEntities }`.  What am I missing?

@eliperelman is on PTO, so I'm hoping @garbas or @helfi92 can offer me some advice on at least one of those items.

### DIFF
--- a/src/lib/Layout.js
+++ b/src/lib/Layout.js
@@ -1,139 +1,25 @@
 import React from 'react';
 import { findDOMNode } from 'react-dom';
 import {
-  Navbar, Nav, NavItem, NavDropdown, MenuItem,
-  Glyphicon, Popover, Overlay, OverlayTrigger,
-  Tooltip, Modal, Button, FormControl, FormGroup,
-  ControlLabel
+  Navbar, Nav, NavItem, NavDropdown, MenuItem, Glyphicon, Popover, Overlay, Modal, Button,
+  FormControl, FormGroup, ControlLabel
 } from 'react-bootstrap';
 import menu from '../menu';
 import * as auth from './auth';
 import { Icon } from './format';
+import OktaMenuItem from './ui/OktaMenuItem';
+import PersonaMenuItem from './ui/PersonaMenuItem';
+import DevelopmentMenuItem from './ui/DevelopmentMenuItem';
+import ManualMenuItem from './ui/ManualMenuItem';
 import './base-layout.less';
-import taskcluster from 'taskcluster-client';
 
 // time before expiration at which we warn
 const EXPIRY_WARNING = 5 * 60 * 1000;
-
-const OktaMenuItem = React.createClass({
-  // This authenticates to Okta by opening a new Window where Okta will do its thing,
-  // then closing that window once creds are acquired.
-  signIn() {
-    // Just loading this page is sufficient.  It eventually redirects back to
-    // https://tools.t.n/login, which updates LocalStorage.  Auth listens for
-    // such updates and loads the creds out of storage.
-    window.open('https://login.taskcluster.net/sso/login', '_blank');
-  },
-
-  render() {
-    return (
-      <OverlayTrigger placement="left" delay={600} overlay={
-        <Tooltip id="okta-signin">If you have a <b>Mozilla LDAP Account</b>, sign
-          in with Okta to maximize your permissions. You can use this option even
-          if you are not an employee! If you also have a Mozillians account, make
-          sure your LDAP email is included in your Mozillians
-          profile.</Tooltip>
-        }>
-        <NavItem onSelect={this.signIn}>
-          <Glyphicon glyph="log-in"/> Sign In with Okta
-        </NavItem>
-      </OverlayTrigger>
-    );
-  }
-});
-
-const PersonaMenuItem = React.createClass({
-  // This interfaces directly with Persona, then uses the Login service API to
-  // convert the resulting promise into TaskCluster credentials.  NOTE: when
-  // Persona is gone, also remove the include.js reference in template.ejs
-  propTypes: {
-    setCredentials: React.PropTypes.func.isRequired,
-    showMessage: React.PropTypes.func.isRequired
-  },
-
-  signIn() {
-    navigator.id.get(assertion => {
-      const port = location.port || location.protocol === 'https:' ? '443' : '80';
-      const audience = `${location.protocol}//${location.host}:${port}`;
-      if (assertion) {
-        const login = new taskcluster.Login();
-        login.credentialsFromPersonaAssertion({ assertion, audience })
-          .then(this.props.setCredentials)
-          .catch(err => {
-            this.propx.showMessage({
-              title: 'Sign-In Error',
-              body: err.details ? `${err.body.details.code}: ${err.body.details.message}` : err.code
-            });
-          });
-      } else {
-        this.propx.showMessage({
-          title: 'Sign-In Cancelled',
-          body: 'Sign-In was cancelled'
-        });
-      }
-    });
-  },
-
-  render() {
-    return (
-      <OverlayTrigger placement="left" delay={600} overlay={
-        <Tooltip id="persona-signin">If you are a Mozillian, but do not have an LDAP
-          account, sign in with Persona. If you do not have a Mozillians profile,
-          set one up now. Get vouched to gain access to additional
-          scopes.</Tooltip>
-        }>
-        <NavItem onSelect={this.signIn}>
-          <Glyphicon glyph="user"/> Sign In with Persona
-        </NavItem>
-      </OverlayTrigger>
-    );
-  }
-});
-
-const DevelMenuItem = React.createClass({
-  // this uses `auth.buildLoginURL()` to generate a URL to the production
-  // login service, so this development instance of tools acts as a third-
-  // party to the produciton instance
-  signIn() {
-    window.open(auth.buildLoginURL(), '_blank');
-  },
-
-  render() {
-    return (
-      <OverlayTrigger placement="left" delay={600} overlay={
-        <Tooltip id="devel-signin">When running tools on a devel server, use this
-          option to get credentials from the production tools</Tooltip>
-        }>
-        <NavItem onSelect={this.signIn}>
-          <Glyphicon glyph="console"/> Development Sign-In
-        </NavItem>
-      </OverlayTrigger>
-    );
-  }
-});
-
-const ManualMenuItem = React.createClass({
-  propTypes: {
-    showManualModal: React.PropTypes.func.isRequired
-  },
-
-  render() {
-    return <OverlayTrigger placement="left" delay={600} overlay={
-      <Tooltip id="manual-signin">Use this option to provide a clientId,
-        accessToken, and certificate manually.</Tooltip>
-      }>
-      <NavItem onSelect={this.props.showManualModal}>
-        <Glyphicon glyph="paste"/> Sign In Manually
-      </NavItem>
-    </OverlayTrigger>;
-  }
-});
-
 const SIGNIN_MENU_ITEMS = {
-  okta: React.createFactory(OktaMenuItem),
-  persona: React.createFactory(PersonaMenuItem),
-  devel: React.createFactory(DevelMenuItem),
-  manual: React.createFactory(ManualMenuItem)
+  okta: OktaMenuItem,
+  persona: PersonaMenuItem,
+  development: DevelopmentMenuItem,
+  manual: ManualMenuItem
 };
 
 /** Navigation bar for layout.jade */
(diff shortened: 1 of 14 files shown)


src/lib/Layout.js
```diff
@@ -177,34 +326,62 @@ const Navigation = React.createClass({
...
+    };
 
+    const showMessage = ({ title, body }) => {
+      this.setState({
+        credentialsMessage: { title, body },
+        signinMenuOpen: false
+      });
+    };
+
+    const signinElements = signinMethods.map(
```
helfi92: Keys should always be supplied directly to the components inside the map function, not to the container HTML child. Hopefully this will fix it.
https://facebook.github.io/react/docs/multiple-components.html#dynamic-children

helfi92: This is probably the reason behind the error at `{ signinElements }`

djmitche:
Hm, that makes sense, but I suppose means I can no longer mix function
calls with react classes.  Thanks for the pointer!

djmitche:
OK, this is as far as I can go.  The only bit remaining un-done is the styling of the modal form, as described above.  Honestly, it's not too bad so maybe I should just drop the style hunk and call it a day?

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
 import menu from '../menu';
 import * as auth from './auth';
 import { Icon } from './format';
 import './base-layout.less';
+import taskcluster from 'taskcluster-client';
 
 // time before expiration at which we warn
 const EXPIRY_WARNING = 5 * 60 * 1000;
 
+const OktaMenuItem = React.createClass({
```
eliperelman: Each of these classes should live in its own file separate from Layout.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+  // This authenticates to Okta by opening a new Window where Okta will do its thing,
+  // then closing that window once creds are acquired.
+  signIn() {
+    // Just loading this page is sufficient.  It eventually redirects back to
+    // https://tools.t.n/login, which updates LocalStorage.  Auth listens for
+    // such updates and loads the creds out of storage.
+    window.open('https://login.taskcluster.net/sso/login', '_blank');
```
eliperelman: Is there no user interaction on this page?

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+  },
+
+  render() {
+    return (
+      <OverlayTrigger placement="left" delay={600} overlay={
+        <Tooltip id="okta-signin">If you have a <b>Mozilla LDAP Account</b>, sign
```
eliperelman: For tags that overflow line length, please place the contents of the tag off of the lines with the tags themselves, e.g.:

```jsx
<Tooltip id="okta-signin">
   Content...
</Tooltip>
```

Also, use `<strong>` tags over `<b>`.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+        </NavItem>
+      </OverlayTrigger>
+    );
+  }
+});
+
+const PersonaMenuItem = React.createClass({
+  // This interfaces directly with Persona, then uses the Login service API to
+  // convert the resulting promise into TaskCluster credentials.  NOTE: when
+  // Persona is gone, also remove the include.js reference in template.ejs
```
eliperelman: You should be able to import this resource in this file rather than modifying `template.ejs`.

```js
import 'https://login.persona.org/include.js';
```

src/template.ejs
```diff
@@ -18,5 +18,6 @@
     ga('create', 'UA-49796218-12', 'auto');
     ga('send', 'pageview');
   </script>
+  <script type="text/javascript" src="https://login.persona.org/include.js"></script>
```
eliperelman: This can be removed since you will import it elsewhere.

Also just FYI, `script type` is not necessary since HTML5.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+  propTypes: {
+    setCredentials: React.PropTypes.func.isRequired,
+    showMessage: React.PropTypes.func.isRequired
+  },
+
+  signIn() {
+    navigator.id.get(assertion => {
+      const port = location.port || location.protocol === 'https:' ? '443' : '80';
+      const audience = `${location.protocol}//${location.host}:${port}`;
+      if (assertion) {
```
eliperelman: Insert blank line between declaration and conditional.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+        const login = new taskcluster.Login();
+        login.credentialsFromPersonaAssertion({ assertion, audience })
```
eliperelman: Insert blank line after declarations. (Gotta get this fixed in the linter.)

eliperelman: If using method chaining, chain all calls on new lines:

```js
login
  .credentialsFromPersonaAssertion({ assertion, audience })
  .then(this.props.setCredentials)
  .catch(err => {
```

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+          body: 'Sign-In was cancelled'
+        });
+      }
+    });
+  },
+
+  render() {
+    return (
+      <OverlayTrigger placement="left" delay={600} overlay={
+        <Tooltip id="persona-signin">If you are a Mozillian, but do not have an LDAP
```
eliperelman: Place tag contents on lines separate from parent tags.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+        </NavItem>
+      </OverlayTrigger>
+    );
+  }
+});
+
+const DevelMenuItem = React.createClass({
+  // this uses `auth.buildLoginURL()` to generate a URL to the production
+  // login service, so this development instance of tools acts as a third-
+  // party to the produciton instance
```
eliperelman: s/produciton/production/

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+  signIn() {
+    window.open(auth.buildLoginURL(), '_blank');
+  },
+
+  render() {
+    return (
+      <OverlayTrigger placement="left" delay={600} overlay={
+        <Tooltip id="devel-signin">When running tools on a devel server, use this
```
eliperelman: Place tag contents on lines separate from parent tags.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+});
+
+const ManualMenuItem = React.createClass({
+  propTypes: {
+    showManualModal: React.PropTypes.func.isRequired
+  },
+
+  render() {
+    return <OverlayTrigger placement="left" delay={600} overlay={
+      <Tooltip id="manual-signin">Use this option to provide a clientId,
```
eliperelman: Place tag contents on lines separate from parent tags.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+        <Glyphicon glyph="paste"/> Sign In Manually
+      </NavItem>
+    </OverlayTrigger>;
+  }
+});
+
+const SIGNIN_MENU_ITEMS = {
+  okta: React.createFactory(OktaMenuItem),
+  persona: React.createFactory(PersonaMenuItem),
+  devel: React.createFactory(DevelMenuItem),
```
eliperelman: Any reason to not use `development`/`DevelopmentMenuItem` vs. the shorthand?

src/lib/Layout.js
```diff
@@ -172,39 +298,72 @@ const Navigation = React.createClass({
...
   renderCredentialsMenu() {
-    // if there are no credentials at all, then there is no menu -- just a sign-in link
-    if (!this.state.credentials) {
-      return (
-        <NavItem onSelect={this.signIn} ref="credentials">
-          <Glyphicon glyph="log-in" /> Sign in
-        </NavItem>
+    let menuHeading;
+    let disabledIfNotSignedIn;
+    if (this.state.credentials) {
```
eliperelman: Insert line after declarations (before this conditional).

src/lib/Layout.js
```diff
@@ -172,39 +298,72 @@ const Navigation = React.createClass({
...
     }
 
-    // TODO: color this according to time until expiry
-    const glyph = this.state.credentialsExpiringSoon ? 'exclamation-sign' : 'user';
-    const className = this.state.credentialsExpiringSoon ? 'text-warning' : '';
-    const menuHeading = (
-      <span>
-        <Glyphicon className={className} glyph={glyph}/> {this.state.credentials.clientId}
-      </span>
+    const setCredentials = credentials => {
```
eliperelman: Move these function declarations outside of this rendering function as it will be called many times, but the functions won't change.

src/lib/Layout.js
```diff
@@ -2,31 +2,155 @@ import React from 'react';
...
+      }>
+      <NavItem onSelect={this.props.showManualModal}>
+        <Glyphicon glyph="paste"/> Sign In Manually
+      </NavItem>
+    </OverlayTrigger>;
+  }
+});
+
+const SIGNIN_MENU_ITEMS = {
+  okta: React.createFactory(OktaMenuItem),
```
eliperelman: Any reason why we need these `createFactory`s vs. just inlining `<OktaMenuItem />`s in the JSX?

src/lib/Layout.js
```diff
@@ -234,6 +393,87 @@ const Navigation = React.createClass({
     );
   },
 
+  renderManualModal() {
+    // This modal must be outside of the "Sign In Manually" menu option as that
+    // menu option is removed from the DOM as soon as it loses focus.
+    const closeModal = () => {
```
eliperelman: Move these functions outside of this rendering function.

src/lib/Layout.js
```diff
@@ -234,6 +393,87 @@ const Navigation = React.createClass({
...
+      e.preventDefault();
+
+      if (this.state.manualCertificate !== '') {
+        try {
+          certificate = JSON.parse(this.state.manualCertificate);
+        } catch (err) {
+          return;
+        }
+      }
+      auth.saveCredentials({
```
eliperelman: Insert line between end of conditional and this function call.

src/lib/Layout.js
```diff
@@ -234,6 +393,87 @@ const Navigation = React.createClass({
...
+      });
+      this.setState({
+        signinMenuOpen: false,
+        showManualModal: false
+      });
+    };
+
+    const certificateIsValid = () => {
+      const manualCertificate = this.state.manualCertificate;
+      if (manualCertificate !== '') {
```
eliperelman: Insert line between declaration and conditional.

src/lib/Layout.js
```diff
@@ -234,6 +393,87 @@ const Navigation = React.createClass({
...
+        try {
+          JSON.parse(manualCertificate);
+        } catch (err) {
+          return false;
+        }
+      }
+      return true;
```
eliperelman: Insert line before return.

src/lib/Layout.js
```diff
@@ -234,6 +393,87 @@ const Navigation = React.createClass({
...
+    };
+
+    const formIsValid = () =>
```
eliperelman:
```js
const formIsValid = () => this.state.manualClientId &&
  this.state.manualAccessToken &&
  certificateIsValid();
```

src/lib/Layout.js
```diff
@@ -234,6 +393,87 @@ const Navigation = React.createClass({
...
+      this.state.manualClientId && this.state.manualAccessToken && certificateIsValid();
+
+    return this.state.showManualModal ? (
```
eliperelman:
```jsx
if (!this.state.showManualModal) {
  return;
}

return (
  <Modal show={true}>
    ...
  </Modal>
);
```

src/lib/auth.js
```diff
@@ -48,6 +54,7 @@ export const loadCredentials = () => {
     return creds;
   } catch (err) {
     debug('Failed to parse credentials, err: %s', err, err.stack);
+    return null;
```
eliperelman: I agree, this linter part is annoying. I'll see about fixing this in the future.

src/lib/auth.js
```diff
@@ -4,11 +4,17 @@ import url from 'url';
 const debug = createDebugger('tools:lib:credentials');
 
 /** Save credentials from localStorage (removed them if null is given) */
-export const saveCredentials = credentials => {
+export const saveCredentials = credentialsParam => {
+  const credentials = credentialsParam; // work around lint error
```
eliperelman: This isn't a workaround, it's an intended feature to avoid problems with functions modifying objects passed to them, since the stateful object could be relied upon elsewhere. Clone the object first:

```js
export const saveCredentials = _credentials => {
  const credentials = { ..._credentials };
}
```

src/lib/base-layout.less
```diff
@@ -55,3 +55,34 @@ pre {
     margin-top: -3px;
   }
 }
+
+.login-form .form-control {
+  position: relative;
+  height: auto;
+  -webkit-box-sizing: border-box;
```
eliperelman: Don't line up properties as it can causes noisy diffs. Go ahead and leave them indented 2 spaces as normal:

```css
-webkit-box-sizing: border-box;
-moz-box-sizing: border-box;
box-sizing: border-box;
```

src/lib/base-layout.less
```diff
@@ -55,3 +55,34 @@ pre {
...
+     -moz-box-sizing: border-box;
+          box-sizing: border-box;
+  padding: 10px;
+  font-size: 14px;
+}
+.login-form .form-control:focus {
```
eliperelman: Insert line between CSS rules.

tools-preset/env.js
```diff
@@ -23,5 +23,8 @@ module.exports = {
     process.env.UPTIMEROBOT_API_KEY_SCHEDULER || 'm776120202-44923d8660c2a1bd1a5de440'),
 
   UPTIMEROBOT_API_KEY_SECRETS: JSON.stringify(
-    process.env.UPTIMEROBOT_API_KEY_SECRETS || 'm777577313-6d58b81186c4064cf7a8d1e1')
+    process.env.UPTIMEROBOT_API_KEY_SECRETS || 'm777577313-6d58b81186c4064cf7a8d1e1'),
+
+  SIGN_IN_METHODS: JSON.stringify(
```
eliperelman: Would this work rather than modifying the startup script?

```js
SIGN_IN_METHODS: JSON.stringify(process.env.SIGN_IN_METHODS ||
  process.env.NODE_ENV === 'development' ? 'devel' : 'okta persona manual')
```

eliperelman:
I'll run this locally and take a look at styling, etc.

src/lib/Layout.js
djmitche: That page itself, no -- it redirects directly to Okta, which does its thing and redirects back to login, which generates credentials and redirects back to tools.  Whew!

tools-preset/env.js
djmitche: I'm looking forward to the configurable future -- recall you wanted to make the uptimerobot API keys configurable, and I see the login options as much more likely to differ from one install to another.

eliperelman: Notice I kept `process.env.SIGN_IN_METHODS`, but provided a default value during development.

djmitche: oh, that makes sense

djmitche:
OK, Thanks!!

PR fixed -- I rebased instead of merging.  I saw at least one thing I want to fix here (manual has an unused signIn function) and I will also double-check that all of the other review comments are addressed, then ask for another round.

src/lib/Layout.js
djmitche: This is un-wrapped in your update, but I think that's OK (it's <98 characters)

PR #154, opened 2016-10-07
[WIP] Cache Purge Request Form

owlishDeveloper:
The form is not functional yet. Just wanted some feedback on its looks, whether I didn't forget to include any fields, and whether I used the correct element to enter the date. I rendered the form in the same file as the rest of the manager, since there is no editor (I noticed all editors go into separate JSX files).


owlishDeveloper:
@imbstack Seems like I did something wrong here... Can you please look? Thanks!

eliperelman:
This commit was based on your previous commit instead of master. You should pull the latest master, create a new branch from it, then rebate your commit on top of that.

owlishDeveloper:
Oh, I see! Thank you, I'll do that!

PR #159, opened 2016-10-12
Added ws-shell support -- interactive shell version 2

jonasfj:
There is probably still a few bugs in ws-shell, but they should be solved in that module..
This basically adds support for a new protocol for interactive shell.. This protocol uses ack messages, instead of requiring pause/resume messages, which you can't send when network is under pressure.

background: tc-worker will be using a new protocol for interactive shell, because implementing the old (bad) protocol in golang seemed like a bad idea... Eventually, we might migrate docker-worker to use this protocol too.

### DIFF
--- a/src/shell/app.jsx
+++ b/src/shell/app.jsx
@@ -43,11 +43,14 @@ term.onTerminalReady = async () => {
     ]
   };
 
+  // Create a shell client, with interface similar to child_process
+  // With an additional method client.resize(cols, rows) for TTY sizing.
   let client;
   if (args.v === '1') {
     client = new DockerExecClient(options);
     await client.execute();
 
+    // Wrap client.resize to switch argument ordering
     const resize = client.resize;
     client.resize = (c, r) => resize.call(client, r, c);
   } else if (args.v === '2') {
--- a/src/shell/app.jsx
+++ b/src/shell/app.jsx
@@ -42,10 +42,12 @@ term.onTerminalReady = async () => {
       ].join('')
     ]
   };
+
   let client;
   if (args.v === '1') {
     client = new DockerExecClient(options);
     await client.execute();
+
     const resize = client.resize;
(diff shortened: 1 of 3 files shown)


src/shell/app.jsx
```diff
@@ -41,8 +41,18 @@ term.onTerminalReady = async () => {
         'exec $SPAWN;'
       ].join('')
     ]
-  });
+  };
+  let client;
+  if (args.v === '1') {
```
eliperelman: Insert line between declaration and conditionals.

src/shell/app.jsx
```diff
@@ -41,8 +41,18 @@ term.onTerminalReady = async () => {
...
+    client = new DockerExecClient(options);
+    await client.execute();
+    const resize = client.resize;
```
eliperelman: Insert line before and after this declaration.

src/shell/app.jsx
```diff
@@ -41,8 +41,18 @@ term.onTerminalReady = async () => {
...
+    client.resize = (c, r) => resize.call(client, r, c);
+  } else if (args.v === '2') {
+    client = await wsshell.dial(options);
+  }
 
+  const io = term.io.push();
```
eliperelman: Insert line after this declaration.

src/shell/app.jsx
jonasfj: before `let client` I can understand

src/shell/app.jsx
eliperelman: Keeps declarations readable from other statements.

jonasfj: but, but.... const resize is only used inside the `client.resize = (...) => ` statement on the next line... I'm effectively wrapping the resize method to switch argument ordering :)

jonasfj:
Hmm... Can't figure out what's up with npm... will regenerate shrinkwrap and update.

PR #162, opened 2016-10-17
Integrating unified-logviewer to replace custom implementation

eliperelman:


eliperelman:
Would either of you be interested in reviewing this change to replace the logviewer with the one from unified-logviewer?

package.json
```diff
@@ -10,60 +10,59 @@
...
+    "redux": "3.6.0",
+    "redux-thunk": "2.1.0",
+    "rison": "0.1.1",
+    "shell-escape": "0.2.0",
+    "slugid": "1.1.0",
+    "taskcluster-client": "1.4.0",
+    "term.js": "0.0.7",
+    "ws-shell": "0.1.5",
+    "www-authenticate": "0.6.2",
+    "xml2js": "0.4.17"
```
jonasfj: We have one or two packages that can't be upgraded... everything else should have a `^` and be locked with npm-shrinkwrap.json..

Otherwise, we'll 100% upgrade thing we can't upgrade later!

eliperelman: Every PR that is opened seems to suffer from the shrinkwrap needing to be regenerated. For the moment I am locking dependencies for tools and removing the shrinkwrap until I can come up with a better solution to having this creep up for every PR.

Note that I am in favor of using the semver ranges and locking sub-dependencies; it's just that at the moment this is a hinderance.

eliperelman:
@imbstack it's possible, but right now we don't deep link to the log in the task inspector. Line highlighting is a feature of the logviewer, not the task inspector, so that would be a separate implementation.

And yes, line wrapping as a stored setting in tools is totally possible, but outside the scope of this, since the core feature belongs to the logviewer. Basically we would just need to set pass `wrapLines=true` to the logviewer URL.

PR #163, opened 2016-10-17
Adding display/include files to the output build

eliperelman:


tools-preset/index.js
```diff
@@ -42,4 +43,13 @@ preset.eslint.ignorePattern = [
   'src/lib/persona.js'
 ];
 
+if (process.env.NODE_ENV === 'production') {
+  // Neutrino copies non-JS files, but we also need the JS files from src/display/include that are
+  // dynamically imported
+  preset.plugins.push(new CopyPlugin([{
+    context: path.resolve(__dirname, '../src'),
+    from: `display/include/**/*`,
```
jonasfj: I wish we copied in all `.js` files... so that when you view sources, the sourcemaps can find the original sources.
Our users are likely to open the console when some error happens...

In particularly we might debug some minor things in production, because things like CORS and https issues can't be debugged on localhost.

PR #165, opened 2016-10-24
Switching to taskcluster preset, using yarn, fixing all linter errors

eliperelman:


eliperelman:
@imbstack @jonasfj starting a conversation around this PR.

I've been using yarn pretty successfully since the last update. I've tried it out in this PR by consuming the shiny new [TaskCluster Web Preset](https://github.com/taskcluster/neutrino-preset-taskcluster-web) by moving the ESLint changes out of this repo and into there. Yarn also brings back "shrinkwrapping", but by using a smarter lockfile which worked pretty well out-of-the-box here. I haven't switched to using version ranges yet, but I should be able to with no ill effects.

With the new preset in place, I've tried to go through and remove all linter errors, which should now fail the build properly when a linting error exists. This should give us a clean Travis build going forward.

I'd appreciate if you two could give a look and let me know your thoughts. Don't worry about all the file changes since those are all whitespace linting errors. I'm more concerned about the usage of Yarn, the lockfile as a concept, and switching to the new preset.

jonasfj:
Linter rules should be:
https://github.com/taskcluster/eslint-config-taskcluster/blob/master/index.js

They should be the same across all projects. Sure web projects might need minor tweaks to accommodate JSX. But otherwise they should be consistent as much as possible.

If you want to argue for changes to the linter rules we use for TC that's fine. But let's do that separately, and after we have adopted neutrino.
Ideally all such changes is made in small increments and with broad consensus.

I notice changes in `'comma-dangle': ['error', 'never'],`, make such changes as a separate PR, so we can keep the ensuing flamewar and associated hostilities to a dedicated thread that's easier to delete :)

Note: I suggest that you check that linter config works for `taskcluster-queue` without modifications.
We also use `eslint-config-taskcluster` in a few other repos.. If one or two of those passes without modifications then the lint config is acceptable.

jonasfj:
Note. I don't oppose yarn and neutrino. But we need a config that is consistent with what we have now, so we can migrate easily. Then we can discuss rule changes later.

eliperelman:
@jonasfj agh, that's my mistake, I thought we could discuss the rule validity here, but I'm totally cool in taking it in phases. In looking at the current taskcluster rules, I notice I made a couple mistakes in copying those over, so I will get that fixed.

I will also try out neutrino and our preset in a few projects before even merging this. I'll try `taskcluster-queue`, `docker-worker` (since @gregarndt has been updating it), and `taskcluster-client` with my branch.

Sorry for the confusion!

PR #167, opened 2016-10-26
Bug 1310035: display 'sign in' when credentials are expired at page load

djmitche:
The issue was that `loadCredentials` was, when invalidating the credentials, still returning the expired credentials.  I think that in the old implementation we were lucky enough to not call `loadCredentials` until Layout started listening for `credentials-changed`, but that has since changed and anyway is a very fragile thing to rely on.

### DIFF
--- a/src/lib/auth.js
+++ b/src/lib/auth.js
@@ -45,13 +45,14 @@ export const loadCredentials = () => {
 
     if (creds.certificate && creds.certificate.expiry < Date.now()) {
       saveCredentials(null); // clear credentials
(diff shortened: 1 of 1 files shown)


djmitche:
Thanks! How's that?

eliperelman:
@djmitche does the `clearTimeout` need to go inside the `if (creds.certificate)`?

djmitche:
I moved it out because it was the first statement in either branch of the conditional.

eliperelman:
Sounds good!

PR #168, opened 2016-11-03
Ignore trailing `index.html` when determining current pages.

tomprince:
It probably makes more sense to redirect to the page without `/index.html`, but if somebody ends up on a page with `/index.html`, it makes sense to correctly detect where in the application they are.

### DIFF
--- a/src/lib/Layout.js
+++ b/src/lib/Layout.js
@@ -123,11 +123,12 @@ const Navigation = React.createClass({
   // Render navigation bar
   render() {
     // Find active menu entry
-    let activeEntry = menu.find(entry => entry.link === location.pathname) ||
+    const pathname = location.pathname.replace(/\/index.html$/, '/');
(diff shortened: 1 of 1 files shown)


eliperelman:
@tomprince in what scenarios would this problem occur?

tomprince:
The only way I know to cuase "Unknown Page" to occur is directly navigating to https://tools.taskcluster.net/index.html or https://tools.taskcluster.net/auth/clients/index.html for example. I ran into this when trying to diagnose what might have been a caching issue while setting up a independent instance (I was trying to access `/index.html' to bypass any potential caching of `/`).

djmitche:
I'm not totally convinced by the use-case, but at any rate I'd rather rewrite the location than silently accept it.

tomprince:
I don't think I'd call it a use case, just a minor UI nicety, in case something weird happens. I've opened #170 to do the redirect instead.

PR #170, opened 2016-11-04
Redirect pages with `/index.html` to the corresponding page without.

tomprince:
This is an alternative to #168.

I'm not familiar enough with react to know if `componentWillMount` is the right place for this logic; the documentation suggests that state shouldn't be manipulated in it, but since this entirely replaces the page, I don't think that is a big issue.

### DIFF
--- a/src/lib/Layout.js
+++ b/src/lib/Layout.js
@@ -51,6 +51,13 @@ const Navigation = React.createClass({
     this.startExpirationTimer();
   },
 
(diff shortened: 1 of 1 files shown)


eliperelman:
I think the consensus here is that we would prefer to have bad URLs fail than silently accept them. Using `index.html` is not a valid URL for the tools UI since we use browser history over file mapping.

PR #171, opened 2016-11-04
Upgrade `neutrino` so deploy doesn't upload on failed builds.

tomprince:
This pulls in the fix from mozilla-neutrino/neutrino#3, which bit me several times trying to deploy an instance of this.


eliperelman:
This is fixed in #165, which is on a temporary hold.

https://github.com/taskcluster/taskcluster-tools/pull/165/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R61

PR #175, opened 2016-11-10
Fixing reference error to namespace in index artifacts

eliperelman:


eliperelman:
Actually, I'm just going to crash-land this.

PR #176, opened 2016-11-11
Guard window.Notification on unsupported devices

eliperelman:
Opening the task group inspector on iOS blows up from lack of `Notification`.

### DIFF
--- a/src/task-group-inspector/lib/utils.jsx
+++ b/src/task-group-inspector/lib/utils.jsx
@@ -115,13 +115,17 @@ export const labels = {
 */
 export const notifications = {
   notifyUser: message => {
-    /* eslint-disable no-new */
-    new Notification(message, {
-      icon: '/lib/assets/taskcluster-36.png',
-    });
-    /* eslint-enable no-new */
+    if ('Notification' in window) {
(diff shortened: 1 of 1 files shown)


eliperelman:
Just looking for a quick review from anyone that can give it. :)

src/task-group-inspector/lib/utils.jsx
```diff
@@ -115,13 +115,17 @@ export const labels = {
 */
 export const notifications = {
   notifyUser: message => {
-    /* eslint-disable no-new */
-    new Notification(message, {
-      icon: '/lib/assets/taskcluster-36.png',
-    });
-    /* eslint-enable no-new */
+    if ('Notification' in window) {
```
jonasfj: I think the trick is `if (window.Notification)`

And `new window.Notification(` to make it clear that we're reaching for a global... Hmm, maybe... not sure -- you probably know better than me :)

jonasfj: I've generally tried to be explicit about prefixing with `window` when it's browser specific...
ie `window.location` but not `window.setTimeout` as that is available in both node and browser.

PR #178, opened 2016-11-15
Decompress zst when running locally

jonasfj:


Callek:
For anyone passing by, these commands worked fine for me locally.

PR #179, opened 2016-11-16
Add an email signin option

djmitche:
This works a lot like Okta: open a new window, which will eventually
redirect to the tools site, which will update LocalStorage and close the
window.

(Note that pending https://github.com/taskcluster/taskcluster-login/pull/38 this will not automatically redirect back; I'll land that first)

### DIFF
--- a/src/lib/ui/EmailMenuItem.js
+++ b/src/lib/ui/EmailMenuItem.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import {OverlayTrigger, Tooltip, NavItem, Glyphicon} from 'react-bootstrap';
+
+export default class extends React.Component {
+  // This authenticates with Auth0's passwordless lock by opening a new Window
+  // where Auth0 will do its thing, then closing that window once creds are
+  // acquired.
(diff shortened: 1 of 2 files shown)


src/lib/ui/EmailMenuItem.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import {OverlayTrigger, Tooltip, NavItem, Glyphicon} from 'react-bootstrap';
+
+const EmailMenuItem = React.createClass({
```
eliperelman: Can you switch the class to be an ES class instead of `React.createClass`?

Also, just export this by default and add a display name:

```js
export default class extends React.Component {
  static displayName = 'EmailMenuItem';
```

djmitche: I prefer `React.createClass` since it correctly binds all of the methods.  It doesn't happen to be an issue for this class, but in general it saves a lot of fiddly `this.foo.bind(this)`.

djmitche:
(updated as requested)

PR #182, opened 2016-12-05
Pulse exchange ui

Naillik1:
UI to list rabbitMQ exchanges from tc-pulse. Lists all exchanges retrieved from the tc-pulse api and allows someone to view the exchange properties.

### DIFF
--- a/src/exchange-inspector/exchangedetails.jsx
+++ b/src/exchange-inspector/exchangedetails.jsx
@@ -1,0 +1,107 @@
+import React from 'react';
+import {Row, Col} from 'react-bootstrap';
+import * as utils from '../lib/utils';
+import * as auth from '../lib/auth';
+import taskcluster from 'taskcluster-client';
+import _ from 'lodash';
+import './exchangeinspector.less';
+
+const ExchangeDetails = React.createClass({
+  
+  mixins: [
+    utils.createTaskClusterMixin({
+      clients: {
+        pulse: taskcluster.Pulse,
+      },
+    }),
+  ],
+
+  getDefaultProps() {
+    return {
+      currentExchange: {},
+    };
+  }, 
+
+  getInitialState() {
+    return {
+      exchangeLoaded: false,
+      exchangeError: null,
+    };
+  },
+
+  load() {
+    return {
+      exchangeError: null,
+    };
+  },
+
+  render() {
+    const args = _.values(this.props.currentExchange.arguments);
+    const argsExist = args.length != 0;
+    return (
+      <div>
+        <Row style={{marginTop: 0}}>
+          <label className="control-label col-md-3">Name</label>
+          <div className="col-md-9">
+            {this.props.currentExchange.name}
+          </div>
+        </Row> 
+        <Row>
+          <label className="control-label col-md-3">Virtual Host</label>
+          <div className="col-md-9">
+            {this.props.currentExchange.vhost}
+          </div>
+        </Row>
+        <Row>
+          <label className="control-label col-md-3">Type</label>
+          <div className="col-md-9">
+            {this.props.currentExchange.type}
+          </div>
+        </Row>
+        <Row>
+          <label className="control-label col-md-3">Durable</label>
+          <div className="col-md-9">
+            {`${this.props.currentExchange.durable}`}
+          </div>
+        </Row>
+        <Row>
+          <label className="control-label col-md-3">Auto-Delete</label>
+          <div className="col-md-9">
+            {`${this.props.currentExchange.auto_delete}`}
+          </div>
+        </Row>
+        <Row>
+          <label className="control-label col-md-3">Internal</label>
+          <div className="col-md-9">
+            {`${this.props.currentExchange.internal}`}
+          </div>
+        </Row>
+        <Row>
+          <label className="control-label col-md-3">Arguments</label>
+          <div className="col-md-9">
+            {
+              argsExist ? (
+                <ul>
+                  {args.map(this.renderArguments)}
+                </ul>
+              ) : (
+                <span>No arguments</span>
+              )
+            }
+          </div>
+        </Row>
+      </div>
+      ); 
+  }, 
+  
+  renderArguments(arg, index) {
+    return (
+      <li key={index}>
+        <code>{`${arg}`}</code>
+      </li>
+    ); 
+  },
+
+});
+
+export default ExchangeDetails;
--- a/src/exchange-inspector/exchangedetails.jsx
+++ b/src/exchange-inspector/exchangedetails.jsx
@@ -3,11 +3,9 @@ import {Row, Col} from 'react-bootstrap';
 import * as utils from '../lib/utils';
 import * as auth from '../lib/auth';
 import taskcluster from 'taskcluster-client';
-import _ from 'lodash';
 import './exchangeinspector.less';
 
 const ExchangeDetails = React.createClass({
-  
   mixins: [
     utils.createTaskClusterMixin({
       clients: {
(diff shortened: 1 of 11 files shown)


src/exchange-inspector/exchangedetails.jsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import {Row, Col} from 'react-bootstrap';
+import * as utils from '../lib/utils';
+import * as auth from '../lib/auth';
+import taskcluster from 'taskcluster-client';
+import _ from 'lodash';
+import './exchangeinspector.less';
+
+const ExchangeDetails = React.createClass({
+
```
eliperelman: Remove this extra line.

src/exchange-inspector/exchangedetails.jsx
```diff
@@ -0,0 +1,107 @@
...
+  load() {
+    return {
+      exchangeError: null,
+    };
+  },
+
+  render() {
+    const args = _.values(this.props.currentExchange.arguments);
+    const argsExist = args.length != 0;
```
eliperelman: This variable is only used once, so it can be removed and made inline. I'll comment there for that.

eliperelman: Insert a blank line between variables and return for readability.

src/exchange-inspector/exchangedetails.jsx
```diff
@@ -0,0 +1,107 @@
...
+    return (
+      <div>
+        <Row style={{marginTop: 0}}>
+          <label className="control-label col-md-3">Name</label>
+          <div className="col-md-9">
```
eliperelman: We have `Col` imported in this file, so replace these with:

```jsx
<Col md={9}></Col>
```

src/exchange-inspector/exchangedetails.jsx
```diff
@@ -0,0 +1,107 @@
...
+        <Row>
+          <label className="control-label col-md-3">Internal</label>
+          <div className="col-md-9">
+            {`${this.props.currentExchange.internal}`}
+          </div>
+        </Row>
+        <Row>
+          <label className="control-label col-md-3">Arguments</label>
+          <div className="col-md-9">
+            {
```
eliperelman: Simplify to:

```jsx
{
  args.length ? (
    <ul>{args.map((arg, key) => <li key={key}><code>{arg}</code></li>)}</ul>
  ) : (
    <span>No arguments</span>
  )
}
```

src/exchange-inspector/exchangedetails.jsx
```diff
@@ -0,0 +1,107 @@
...
+                <span>No arguments</span>
+              )
+            }
+          </div>
+        </Row>
+      </div>
+      ); 
+  }, 
+  
+  renderArguments(arg, index) {
```
eliperelman: Remove this method and inline. See other comment.

src/exchange-inspector/exchangedetails.jsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+
+  load() {
+    return {
+      exchangeError: null,
+    };
+  },
+
+  render() {
+    const args = _.values(this.props.currentExchange.arguments);
```
eliperelman: Remove the lodash dependency and use `Object.values()`.

src/exchange-inspector/exchangeinspector.jsx
```diff
@@ -0,0 +1,99 @@
...
+  displayName: 'ExchangeInspector',
+
+  /** Initialize mixins */
+  mixins: [
+    utils.createTaskClusterMixin({
+      clients: {
+        pulse: taskcluster.Pulse,
+      },
+    }),
+  // Serialize state.selectedExchange to location.hash as string
```
eliperelman: Indent this comment 2 more spaces.

src/exchange-inspector/exchangeinspector.jsx
```diff
@@ -0,0 +1,99 @@
...
+import * as utils from '../lib/utils';
+import * as auth from '../lib/auth';
+import taskcluster from 'taskcluster-client';
+import _ from 'lodash';
+import './exchangeinspector.less';
+
+export default React.createClass({
+  displayName: 'ExchangeInspector',
+
+  /** Initialize mixins */
```
eliperelman: Remove this comment.

src/exchange-inspector/exchangeinspector.jsx
```diff
@@ -0,0 +1,99 @@
...
+    };
+  },
+  
+  load() {
+    return {
+      exchanges: this.pulse.exchanges(),
+    };
+  },
+
+  /** Render user-interface */
```
eliperelman: Remove this comment.

src/exchange-inspector/exchangeinspector.jsx
```diff
@@ -0,0 +1,99 @@
...
+  render() {
+    const exchangeSelected = this.state.selectedExchange != null && typeof this.state.selectedExchange != 'undefined';
```
eliperelman: Change this to:

```js
const exchangeSelected = this.state.selectedExchange != null;
```

Insert a line between the variable declaration and the `return`.

src/exchange-inspector/exchangeinspector.jsx
```diff
@@ -0,0 +1,99 @@
...
+            ) : (
+              <div>Select an exchange to view details</div>
+            )
+          }
+        </Col>
+      </Row>
+    ); 
+  },
+  
+  /** Render table of all exchanges */
```
eliperelman: Remove this comment.

src/exchange-inspector/exchangeinspector.jsx
```diff
@@ -0,0 +1,99 @@
...
+  renderExchangesTable() {
+    return this.renderWaitFor('exchanges') || (
+      <Table condensed={true} hover={true} className="exchange-inspector-exchanges-table">
+        <tbody>
+          {this.state.exchanges.map(this.renderExchangesRow)}
```
eliperelman: Change this to:

```jsx
{this.state.exchanges.map((exchange, key) => (
  <tr
    key={key}
    className={this.state.selectedExchange === exchange ? 'info' : null}
    onClick={() => this.selectExchange(exchange)}
    <td><code>{exchange.name}</code></td>
  </tr>
)}
```

src/exchange-inspector/exchangeinspector.jsx
```diff
@@ -0,0 +1,99 @@
...
+        </tbody>
+      </Table>
+    );
+  },
+
+  /** Render a row with an exchange*/
+  renderExchangesRow(exchange, index) {
```
eliperelman: Remove this method and inline it. See previous comment.

djmitche:
I'll land this after http://bugzil.la/1335954 is done.  Right now the service isn't actually pointing to the pulse server, but to its own cloudamqp instance, which would cause a lot of confusion.

djmitche:
I had another look at this -- it still works great!  The one thing I would fix is to put just the exchange name in the location, rather than the whole exchange data structure.

However, the list of exchanges is .. pretty huge.  And most of them are users' test exchanges -- not an especially interesting list.

I think that the more useful UI here would be to use the exchanges list to autocomplete the list in the pulse inspector.

We could potentially also create a view of pulse namespaces (which will not encompass all of the things users might want to listen to, since e.g., hgpushes doesn't use tc-pulse), and provide a way to list the exchanges and queues associated with each namespace.  But such a tool would only very occasionally be useful for debugging purposes, so I'm not sure it's worth the effort.  For example, we don't have such a tool for signalfx or sentry credentials.

All of which is to say, I think we should close this without merging.  @eliperelman do you agree?

eliperelman:
Yes, I agree. Thank you for the effort, and we will consider how best to approach this in the future, if this even belongs in the web UI.

PR #183, opened 2016-12-07
Goodbye two-click loaner

eliperelman:
Right now the one-click loaner page actually requires two clicks in order to launch a loaner, which makes it a bit of a misnomer, as well as degrading UX by complicating the flow of intent. Instead of keeping the relevant information hidden in a modal, this PR puts that data on the page and jumps immediately to launching the one-click loaner when clicking the button.

### DIFF
--- a/src/one-click-loaner/one-click-loaner.jsx
+++ b/src/one-click-loaner/one-click-loaner.jsx
@@ -3,7 +3,7 @@ import {findDOMNode} from 'react-dom';
 import {FormGroup, FormControl, ControlLabel, InputGroup, Button} from 'react-bootstrap';
 import * as utils from '../lib/utils';
 import taskcluster from 'taskcluster-client';
-import LoanerButton from '../lib/ui/loaner-button';
+import LoanerButton from '../lib/ui/LoanerButton';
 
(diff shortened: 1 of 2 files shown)


Callek:
As a random fly-by, can you provide a bug # or other information that justifies/explains this change?

eliperelman:
@Callek there's no bug but more of a UX improvement, but I will update the description appropriately. (Should have taken gps's advice here lol).

src/one-click-loaner/one-click-loaner.jsx
```diff
@@ -3,7 +3,7 @@ import {findDOMNode} from 'react-dom';
 import {FormGroup, FormControl, ControlLabel, InputGroup, Button} from 'react-bootstrap';
 import * as utils from '../lib/utils';
 import taskcluster from 'taskcluster-client';
-import LoanerButton from '../lib/ui/loaner-button';
```
djmitche: missing a `git rm` of this?

eliperelman: If I redirect task inspector to one-click loaner, then yes, we can remove this.

PR #185, opened 2016-12-22
Clarity for the format of the routing key pattern

srfraser:
As the documentation mentions at https://docs.taskcluster.net/reference/platform/queue/exchanges the routes must be prefixed by 'route.'.  As an aid to new users, let's mention that on the UI for Pulse Inspector.


src/pulse-inspector/pulseinspector.jsx
```diff
@@ -214,7 +216,7 @@ export default React.createClass({
             <FormControl
               type="text"
               ref="routingKeyPattern"
-              placeholder="#.some-interesting-key.#"
+              placeholder="#route.some-interesting-key.#"
```
djmitche: `#` alone as a character in a routing pattern has a specific meaning ([zero or more words](https://www.rabbitmq.com/tutorials/tutorial-five-python.html)).  `#route` does not.  So aside from the queue-specificity issue, this would need to be `route.#.some-interesting-key.#` or something like that.

PR #187, opened 2017-01-10
Yaml in task creator

jonasfj:
YAML is a strict subset of JSON, so this should work nicely even when copy/pasting JSON into the form...


jonasfj:
This also fixes annoying initial font-size bug...

PR #189, opened 2017-01-18
Strip trailing and leading spaces from scope names

djmitche:
This has bitten us a few times, when copy/pasting scope lists.  The
spaces are invisible when pasted.

This still allows spaces *in* a scope, just not leading or trailing
spaces.  Those are allowed in the API, just not in the UI.

@jonasfj I know we talked about this once.. but at this point I think this will prevent issues and that's more important than preventing the typing of some hypothetical scopes.  It caused a treeclosure today :(


jonasfj:
A warning would have done too... but I don't mind stripping them :)

PR #191, opened 2017-01-23
use splice correctly to remove array elements

### LINKED ISSUE #190: deleting hook schedules is buggy
@callek and I have noticed the following behavior:

- When we try to delete the only schedule from a hook, the trash button does nothing
- When we create a 2nd schedule on a hook, the trash button for the first item deletes the newly created 2nd item
 - @callek noticed that the trash button for the 2nd item deletes the first

I was thinking this might be either
- permissions: I didn't create the hook that I was trying to delete the schedule for, or
- index-starts-at-0 bug, so deleting the first item (bug: index 1) deletes the 2nd item
 - the final point, the trash button for the 2nd item deleting the 1st schedule, doesn't match up with this unless the indexes wrap

My workaround for this has been to copy the hook's task definition, save it elsewhere, and then delete it completely.  We can then recreate the hook with the right schedule(s).  This isn't the ideal workflow, so I'm hoping this is a relatively easy fix.

djmitche:
Fixes #190

### DIFF
--- a/src/hooks/hookeditor.jsx
+++ b/src/hooks/hookeditor.jsx
@@ -478,9 +478,8 @@ const HookEditor = React.createClass({
   },
 
   removeScheduleItem(index) {
-    const schedule = _
-      .cloneDeep(this.state.schedule)
-      .splice(index, 1);
+    const schedule = [...this.state.schedule];
+    schedule.splice(index, 1);
 
     this.setState({schedule});
   },


Callek:
fwiw, doing this manually in the browser debugger seemed to work!

djmitche:
Thanks for the confirmation.  Sorry I (still) suck at JS :)

src/hooks/hookeditor.jsx
```diff
@@ -478,9 +478,8 @@ const HookEditor = React.createClass({
   },
 
   removeScheduleItem(index) {
-    const schedule = _
-      .cloneDeep(this.state.schedule)
-      .splice(index, 1);
+    const schedule = _.cloneDeep(this.state.schedule);
```
eliperelman: A few different ways we could do this without lodash:

```js
const schedule = [...this.state.schedule];

schedule.splice(index, 1);
this.setState({ schedule });
```

```js
const schedule = [
  ...this.state.schedule.slice(0, index),
  ...this.state.schedule.slice(index + 1)
];

this.setState({ schedule });
```

djmitche:
I merged *and* updated :)

PR #194, opened 2017-01-31
`.taskcluster.yml` creator UI

owlishDeveloper:
The page looks and works as I imagined, but I feel that the code is a mess. Please review and give your feedback!

@imbstack 
@djmitche 
@eliperelman


src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+      PRoCheckbox: false,
+      PRcCheckbox: false,
+      PRsCheckbox: false,
+      PRrCheckbox: false,
+      pushCheckbox: false,
+      relCheckbox: false,
+    });
+  }
+
+  renderEditor() {
```
eliperelman: I know it's a little more verbose, but I think this helps break down the structure of the object being generated, without relying on mutation. Also if `state.events` becomes a Set, that will need to be converted before being stringified:

```jsx
renderEditor() {
  const newYml = {
    ...initialYml,
    metadata: {
      ...initialYml.metadata,
      name: this.state.rootName,
      description: this.state.rootDescription,
    },
    tasks: {
      ...initialYml.tasks[0],
      ...{
        metadata: {
          name: this.state.taskName,
          description: this.state.taskDescription,
        },
        extra: {
          github: {
            events: [...this.state.events],
          },
        },
        payload: {
          command: this.state.commands,
          image: this.state.image,
        },
      },
    },
  };
}
```

src/yml-creator/ymlcreator.less
```diff
@@ -0,0 +1,25 @@
+@import "~codemirror/lib/codemirror.css";
+@import "~codemirror/addon/lint/lint.css";
+@import "~codemirror/theme/ambiance.css";
+
+.btn-toolbar {
```
eliperelman: Indent this file using 2 spaces per indentation.

src/yml-creator/ymlcreator.less
```diff
@@ -0,0 +1,25 @@
...
+
+.CodeMirror {
+   height: auto;
+   margin-bottom: 20px;
+}
+
+.CodeMirror-scroll {
+  overflow-x: auto;
+  overflow-y: hidden;
+}
```
eliperelman: Add blank line to end of file.

src/menu.js
```diff
@@ -28,6 +28,14 @@ module.exports = [
       in \`localStorage\` so you can try again with new parameters`,
   },
   {
+    title: 'YML Creator',
```
eliperelman: You'll see me use `YML`, `yml`, or `Yml` in my review comments, but the actual phrase is `YAML`. When being viewed for display, I'd like to change every instance to be the full spelled out form of YAML, except for file extensions where it's appropriate.

So to summarize, except for `.yml` file extensions, please spell as YAML. Examples:

- `YAML Creator`, all caps when users see the YAML phrase.
- `/yaml-creator/`, all lowercase in URL.
- `YamlCreator`, Pascal cased when used as a component or constructor
- `initialYaml` or `yamlInitial`, or `newYaml`, camel cased when used in a variable
- `.taskcluster.yml`, only `yml` when used in a file extension

Hope that makes sense. Ping me if you have questions.

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+    source: '{{ event.head.repo.url }}',
+  },
+  tasks: [
+    {
+      provisionerId: '{{ taskcluster.docker.provisionerId }}',
+      workerType: '{{ taskcluster.docker.workerType }}',
+      extra: {
+        github: {
+          env: true,
+          events: [],
```
eliperelman: I make a comment later about using a Set here instead of an array. If that happens, you'll want to change this here to be a Set as well:

```js
events: new Set(),
```

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+];
+
+export default class YmlCreator extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      rootName: '',
+      rootDescription: '',
+      tasks: [],
+      events: [],
```
eliperelman: I make a comment later about using a Set here instead of an array. If that happens, you'll want to change this here to be a Set as well:

```js
events: new Set(),
```

owlishDeveloper: oh how interesting! Thank you, I didn't know about those! _owlish goes to read MDN_

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+      PRcCheckbox: false,
+      PRsCheckbox: false,
+      PRrCheckbox: false,
+      pushCheckbox: false,
+      relCheckbox: false,
+
+      displayEditor: false,
+      file: initialYML,
+    };
+    this.saveTextInput = this.saveTextInput.bind(this);
```
owlishDeveloper: I tried using arrow functions, but I kept getting syntax errors. What's the correct syntax? I tried:
```
const saveTextInput = (event) => {
    this.setState({
      [event.target.name]: event.target.value,
    });
  };
```
instead of 
```
saveTextInput(event) {
    this.setState({
      [event.target.name]: event.target.value,
    });
  }
```

eliperelman: Apologies, I mean arrow functions where they are called from. So instead of:

```jsx
onChange={this.saveTextInput} />
```

it would be 

```jsx
onChange={e => this.saveTextInput(e)} />
```

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+      <Grid fluid={true}>
+        <Row>
+          <Col sm={12}>
+            <h4>Configuration File Creator</h4>
+            <p>
+              This tool lets you easily generate a simple generic <code>.taskcluster.yml</code> file, 
+              which should live at the root of your repository. It defines
+              tasks that you want TaskCluster to run for you. The tasks will run when certain 
+              GitHub events happen — you will choose the events you're interested in while 
+              creating the file. After pressing <b>Create file</b> button, you will see the contents of
```
owlishDeveloper: But the name of the button doesn't have any special importance, it just needs to be bold, that's why I chose `<b>`.

src/yml-creator/ymlcreator.jsx
owlishDeveloper: Thank you so much!

src/yml-creator/ymlcreator.jsx
eliperelman: We don't use the `<b>` tag to convey bold in HTML since HTML4:

> Historically, the <b> element was meant to make text boldface. Styling information has been deprecated since HTML4, so the meaning of the <b> element has been changed.

Semantically, the `<strong>` tag now represents bolded typeface. This is the same reason we don't use `<i>` for italics, as it no longer carries that semantic meaning, being replaced by `<em>`. This is why you'll see frameworks like Font Awesome repurpose the semantics of `<i>` for icons.

tldr; Yes, change it to `<strong>`. :smile:

src/yml-creator/ymlcreator.jsx
owlishDeveloper: Oh how cool! Thank you!

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+              your pushes, releases and pull requests, you have to install 
+              <a href="https://github.com/integration/taskcluster-staging"> TaskCluster-GitHub integration</a>.
+            </p>
+            <hr />
+          </Col>
+        </Row>
+
+        <Row>
+          <Col sm={12}>
+            <h5>Enter the name and desciption of your project or these tasks:</h5>
```
djmitche: It would be great to have some explainer sentences for each form element or group, maybe in <em> tags.  Here, something like "<em>These strings will appear at the top of `.taskcluster.yml` and help the reader understand what they are seeing.</em>"

(OK, I dislike that there's this top-level metadata in `.taskcluster.yml`, since it isn't actually used.. in another bug, let's make the fields optional and remove the form elements..)

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+              <FormControl
+                type="text"
+                placeholder="Description of the task"
+                name="taskDescription"
+                value={this.state.taskDescription}
+                onChange={this.saveTextInput} />
+            </FormGroup>
+
+            <FormGroup id="checkboxGroup">
+              <ControlLabel>Events:</ControlLabel>
```
djmitche: "<em>Select the Github events for which you want this task to run.</em>"

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+                id="relCheckbox"
+                class="data_checkboxes"
+                checked={this.state.relCheckbox}
+                onChange={this.eventsSelection}>
+                Release or tag created
+              </Checkbox>
+            </FormGroup>
+
+            <FormGroup>
+              <ControlLabel>Language:</ControlLabel>
```
djmitche: "<em>Select the language your project uses.  This will select a corresponding docker image and some basic commands for checking out and testing a typical project.</em>"

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+    );
+  }
+
+  saveTextInput(event) {
+    this.setState({
+      [event.target.name]: event.target.value,
+    });
+  }
+
+  eventsSelection(event) {
```
owlishDeveloper: I like the idea of sets because they're <s>new and exci</s> convenient here, but even with sets, aren't we mutating the object that belongs to the state anyway? I mean, we essentially do the same thing, only we use set instead of an array...

eliperelman: No, we aren't mutating the state.

```js
// This creates a copy of the Set that is in state
const events = new Set(this.state.events);

// modifies the Set we just copied
events.delete();

// modifies the Set we copied
events.add();
```

With React state (and even more so with Redux state), we avoid mutating state at all costs. Mutation can cause side effects in other place that maintain a reference to the state, but can cause problems because they didn't follow the normal updating flow put in place by React (via `setState`). It is OK to create a copy and modify it within the scope of the function which will make a call to update, which is what we are doing here, but we should never directly modify an object belonging to `state`.

```js
// This would modify the events in state in place, causing potential issues
this.state.events.splice();
```

eliperelman: Like I said, using a Set is optional, it's just makes for using arrays more verbose/imperative, something similar to this:

```js
const events = [...this.state.events];
const includes = events.includes(event.target.name);

if (!includes) {
  events.push(event.target.name);
}

this.setState({
    [event.target.id]: !this.state[event.target.id],
    events: includes ?
      events.filter(e => e !== event.target.name) :
      events
  });
```

owlishDeveloper:
@djmitche 

- Yeah, what a great idea about "Github Quick-Start"! I like that!
- I thought if the editor will be there from the start, it might cause some distraction and confusion. A form seemed cleaner and more straightforward to me. And you can press the button to see the YAML at any point. As for dynamic updating - I thought if I should get rid of it, but decided against it, because that way the user can experiment with input to see what goes where into the file. Sort of an interactive experience instead of just reading through documentation and comments in an example file. "Reset" - I thought if somebody would start experimenting with the file and editing and re-editing and gets carried away, or accidentally deletes something or does whatever else users usually do - "Reset" is kinda more civilized than having to reload the page. What do you think?
- I plan to add that feature after these reviews, don't worry! :)
- I'll try that, thanks!
- Great point about the commands! I totally forgot! Thank you so much!

As for the last idea - I actually had thought about that (I had two wireframes, one was this one, and another one was very similar to your idea), but decided that this option is sort of quicker. I don't know though :) I can make a branch and implement your idea, so we could actually use both and compare and decide which one we like better.

src/yml-creator/ymlcreator.jsx
```diff
@@ -0,0 +1,360 @@
...
+              your repository, create a file at its root, paste and save 
+              as <code>.taskcluster.yml</code>. Optionally, after you create your file, you can edit 
+              it here or in you favorite editor to add more bells and wistles - please refer to 
+              the <a href="https://docs.taskcluster.net/manual/integrations/github#a-simple-taskcluster-yml-file">
+              full documentation on our configuration files</a>.
+            </p>
+            <p>
+              <em>Attention!</em> TaskCluster begins accepting jobs as soon as a <code>.taskcluster.yml</code> exists 
+              in your repository. However, if you want to see the status of TaskCluster jobs in
+              your pushes, releases and pull requests, you have to install
```
owlishDeveloper: You're right! Even with oAuth app, nothing will happen except the robot will come and comment. Yeah, I'll change that!

djmitche:
I agree about the dynamic update being a good teaching tool, which is why I want to see the editor earlier.  What if the editor was dimmed out with an overlay that says "Begin filling out the project details to see your `.taskcluster.yml` developed here" or something like that?

I can be convinced to keep the "Reset" button -- but maybe name it "Reset Form"?

@eliperelman do you have thoughts on whether the "unfolding" interface makes sense?

owlishDeveloper:
You can press the button any time! I can suggest that in the instructions explicitly and rename the button to "Reveal the file". The "Reset editor" button actually resets everything - both the file and the form. I can rename it to "Reset all" or simply "Reset".

djmitche:
Sure, but I didn't see the button until I filled out the rest of the form.
It's really cool and informative to see the .taskcluster.yml updating as
you type things into the form, so let's do that as soon as possible.  Only
showing that after the first thing is typed is fine, though.

src/yml-creator/ymlcreator.jsx
owlishDeveloper: Ah, I see now! Thank you for the explanation! I'll use sets

owlishDeveloper:
@djmitche @imbstack @eliperelman review please!

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+      image: event.target.value,
+      currentCmd: cmdDirectory[event.target.value],
+      resetActive: true,
+    });
+    this.state.displayCmds ? this.setState({commands: cmdDirectory[event.target.value]}) :
+      this.setState({commands: []});
+  }
+
+  handleCommandsSelection(event) {
+    this.setState({
```
eliperelman: Consolidate `setState` calls:

```js
this.setState({
  displayCmds: event.target.value === 'standard',
  currentCmd: this.state.commands,
  commands: event.target.value === 'standard' ?
    this.state.commands :
    [],
});
```

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+      this.setState({commands: []});
+  }
+
+  resetAll() {
+    this.setState({
+      resetActive: false,
+      rootName: '',
+      rootDescription: '',
+      tasks: [],
+      events: [],
```
eliperelman:
```js
events: new Set(),
```

src/quickstart/yamlcreator.less
```diff
@@ -0,0 +1,30 @@
...
+  margin-bottom: 20px;
+}
+
+.CodeMirror-scroll {
+  overflow-x: auto;
+  overflow-y: hidden;
+}
+
+#anotherTask {
+	display: none;
```
eliperelman: Indent these CSS properties only 2 spaces.

src/quickstart/yamlcreator.less
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.infoText {
+	color: rgb(77, 185, 217);
```
eliperelman: Indent 2 spaces.

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+      [event.target.name]: event.target.value,
+      resetActive: true,
+    });
+  }
+
+  handleEventsSelection(event) {
+    const events = new Set(this.state.events);
+    events.has(event.target.name) ? events.delete(event.target.name) : events.add(event.target.name);
+    this.setState({
+      events: [...events],
```
eliperelman: Found it, you are already converting the Set to an array on line 374, so it doesn't need to be done here.

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+            </p>
+            <hr />
+          </Col>
+        </Row>
+
+        <Row>
+          <Col md={5}>
+            <h5>Enter the name and description of your project or these tasks:</h5>
+            <p className="infoText">
+              <Glyphicon glyph="info-sign" />
```
eliperelman: Insert a space after here:

```js
<Glyphicon glyph="info-sign" />&nbsp;
```

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+                Release or tag created
+              </Checkbox>
+            </FormGroup>
+
+            <FormGroup>
+              <ControlLabel>
+                Language your project uses:
+              </ControlLabel>
+              <p className="infoText">
+                <Glyphicon glyph="info-sign" />
```
eliperelman: Insert a space after here:

```js
<Glyphicon glyph="info-sign" />&nbsp;
```

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+              which should live at the root of your repository. It defines
+              tasks that you want TaskCluster to run for you. The tasks will run when certain 
+              GitHub events happen — you will choose the events you're interested in while 
+              creating the file.
+            </p>
+            <h5>How to set up your repository with TaskCluster:</h5>
+            <ul>
+              <li>
+                Fill out the form below. Note that you can press the 
+                <strong>Reveal file in editor</strong> button at any time to see your file. All
```
eliperelman: Need a space before the word "Reveal". Also, I don't see this button anywhere.

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+                changes in the form will be instantly show up in the editor.
```
eliperelman: Remove word "be".

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+              </li>
+              <li>
+                When you are done editing, copy the contents of the editor. Go to
```
eliperelman: When you are done editing, copy the contents of the editor and paste it into a file named `.taskcluster.yml` in the root of your repository.

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+          <Col sm={12}>
+            <h4>GitHub Quick-Start</h4>
+            <p>
+              This tool lets you easily generate a simple generic <code>.taskcluster.yml</code> file, 
+              which should live at the root of your repository. It defines
+              tasks that you want TaskCluster to run for you. The tasks will run when certain 
+              GitHub events happen — you will choose the events you're interested in while 
+              creating the file.
+            </p>
+            <h5>How to set up your repository with TaskCluster:</h5>
```
eliperelman: Insert `<hr />` before this `<h5>`.

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+    '/bin/bash',
+    '--login',
+    '-c',
+    'git clone {{event.head.repo.url}} repo && cd repo && git checkout {{event.head.sha}} && python setup.py test',
+  ],
+  'jimmycuadra/rust:latest': [
+    '/bin/bash',
+    '--login',
+    '-c',
+    'git clone {{event.head.repo.url}} repo && cd repo && git checkout {{event.head.sha}} && rustc --test unit_test.rs && ./unit_test',
```
eliperelman: The linter errors from this line, so use an `eslint-disable` directive in order to appease the linting gods.

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+  ],
+  'golang:1.8': [
+    '/bin/bash',
+    '--login',
+    '-c',
+    'go get -t github.com/taskcluster/taskcluster-cli/... && cd  /go/src/github.com/taskcluster/taskcluster-cli make && go test ./...',
```
eliperelman: The linter errors from this line, so use an `eslint-disable` directive in order to appease the linting gods.

src/menu.js
```diff
@@ -3,6 +3,14 @@
 // Note, only entries with a description will be show on the landing page.
 module.exports = [
   {
+    title: 'GitHub Quick-Start',
```
eliperelman: Also, move this menu object to after "Task Creator".

src/quickstart/YamlCreator.jsx
```diff
@@ -0,0 +1,403 @@
...
+                changes in the form will be instantly show up in the editor.
+              </li>
+              <li>
+                When you are done editing, copy the contents of the editor. Go to 
+                your repository, create a file at its root. Paste. Save 
+                as <code>.taskcluster.yml</code>.
+              </li>
+              <li>
+                Make sure to install 
+                the <a title="bobobob" href="https://github.com/integration/taskcluster-staging"> TaskCluster-GitHub
```
owlishDeveloper: magical taskcluster spirit I guess 👹

src/quickstart/YamlCreator.jsx
owlishDeveloper: (I put the directive but I'm not sure if I did this correctly though)

PR #195, opened 2017-02-09
add maxRunTime

djmitche:
docker-worker requires this parameter

@owlishDeveloper fyi :)

### DIFF
--- a/src/quickstart/YamlCreator.jsx
+++ b/src/quickstart/YamlCreator.jsx
@@ -379,6 +379,7 @@ export default class YamlCreator extends React.Component {
             },
           },
           payload: {
+            ...initialYaml.tasks[0].payload,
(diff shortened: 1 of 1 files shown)


src/quickstart/YamlCreator.jsx
```diff
@@ -381,6 +381,7 @@ export default class YamlCreator extends React.Component {
           payload: {
             command: this.state.commands,
             image: this.state.image,
+            maxRunTime: 3600,
```
owlishDeveloper: I think it should be:

payload: {

    ...initialYaml.tasks[0].payload,

   command: blahblahblah

djmitche: Yep, that is the better fix -- thanks!

PR #199, opened 2017-02-23
Error displayed, when resetting access tokens, is clear. Bug fixed.

Sidd0107:


src/auth/clients/clienteditor.jsx
```diff
@@ -429,7 +429,7 @@ const ClientEditor = React.createClass({
     } catch (err) {
       this.setState({
         working: false,
-        error: err,
+        error: ': Sorry, you are not logged in. You need to be logged in to reset access tokens. \n ' + err,
```
djmitche: What if I *am* logged in, but don't have permission to reset access tokens?  The this message would be confusing!

Sidd0107: In that case, I will modify it to say "You are either not logged in or do not have permission to reset access tokens"

eliperelman: How about "You do not have permission to reset access tokens. Please ensure you are logged in and have sufficient privileges."

Sidd0107: Sounds good. I will change it accordingly. Thanks

Sidd0107: Done. The latest version has the change stated above.

djmitche:
Thanks!  But the code has access to determine whether or not the user is logged in.  I would like to see a *different* error message for those two conditions.

Sidd0107:
I will change it to so that the code determines if the user is logged in and displays the error message accordingly.

src/auth/clients/clienteditor.jsx
```diff
@@ -427,10 +427,24 @@ const ClientEditor = React.createClass({
         working: false,
       });
     } catch (err) {
-      this.setState({
```
eliperelman: At the top of the page we import `auth`, a helper for working with credentials:

https://github.com/Sidd0107/taskcluster-tools/blob/1068ab5fa1da8521791e404438918594a5b1671f/src/auth/clients/clienteditor.jsx#L11

It is already in use in this file, and you can see an example of that here:

https://github.com/Sidd0107/taskcluster-tools/blob/1068ab5fa1da8521791e404438918594a5b1671f/src/auth/clients/clienteditor.jsx#L115-L116

So the value of whether there are credentials or not is driving the value. The same technique can be applied here as well.

```js
this.setState({
  working: false,
  error: auth.loadCredentials() ?
    'You do not have sufficient permission to reset access tokens for this user.' :
    'You must be logged in and have permission to reset access tokens for this user.'
});
```

src/auth/clients/clienteditor.jsx
```diff
@@ -427,10 +427,24 @@ const ClientEditor = React.createClass({
...
-        working: false,
-        error: err,
-      });
+      try {
+        if (this.auth._options.credentials.clientId!=null) {
+          this.setState({
+            working: false,
+            error: ': Sorry, you do not have permission to reset access tokens for this user. '  + err,
```
eliperelman: I put some example messages in my previous comment, but let's avoid apologizing to the user. Could you also post some screenshots of what these messages look like in action?

src/auth/clients/clienteditor.jsx
```diff
@@ -427,10 +427,24 @@ const ClientEditor = React.createClass({
         working: false,
       });
     } catch (err) {
-      this.setState({
-        working: false,
-        error: err,
-      });
+      try {
```
eliperelman: Remove surrounding the `setState` from a `try/catch` as there are no errors that will be thrown here.

src/auth/clients/clienteditor.jsx
Sidd0107: Wow, I never knew you could just do 

`auth.loacCredentials()`

within the error. This just made everything a lot easier. 

 ? within

src/auth/clients/clienteditor.jsx
Sidd0107: Just did that. The earlier method I was using threw and error each time I was checking for undefined.

src/auth/clients/clienteditor.jsx
Sidd0107: screen shot 2017-03-09 at 10 14 04 am
screen shot 2017-03-09 at 10 13 33 am

Sidd0107: And when the user is not logged in:

screen shot 2017-03-09 at 10 13 33 am

Sidd0107: Thank you very much for all the detailed guidance Eli.

djmitche:
I like - thanks for the reviews Eli!

We'll talk about next steps back on the bug :)

Sidd0107:
Thank you very much Dustin and Eli for all your help and guidance. This was my first bug fix. Dustin, I look forward to discussing other bugs with you on the same Bugzilla page.

PR #202, opened 2017-03-08
Improve hooks page ui a bit

imbstack:
This does 3 things:

1. Updates the text on the "email if failed" section to help indicate that this is not for failed tasks.
2. Make it so that triggering the hook adds a new link to the new task.
3. Updates react-treeview because chenglou/react-treeview#35 was annoying me

Of the three, (2) is the one I am the least sure about. It seems like I want the UI to react that way, but perhaps the way I've gone about it is un-react-y or something.


package.json
```diff
@@ -28,7 +28,6 @@
     "font-awesome-svg-png": "^1.1.5",
     "highlight.js": "^9.6.0",
     "hterm-umd": "^1.0.1",
-    "js-yaml": "^3.7.0",
```
eliperelman: ._.

src/hooks/hookeditor.jsx
```diff
@@ -57,37 +57,22 @@ const HookStatusDisplay = React.createClass({
...
+    hookStatus: React.PropTypes.object,
+    refreshHookStatus: React.PropTypes.func.isRequired,
   },
 
   render() {
-    const waitFor = this.renderWaitFor('hookStatus');
-
-    if (waitFor) {
-      return waitFor;
+    const stat = this.props.hookStatus;
```
eliperelman: Insert line after declaration.

src/hooks/hookeditor.jsx
```diff
@@ -645,12 +637,24 @@ const HookEditView = React.createClass({
...
       .triggerHook(this.props.currentHookGroupId, this.props.currentHookId, {})
+      .then(this.refreshHookStatus)
       .catch(error => this.setState({error}));
   },
 
+  refreshHookStatus() {
+    this.setState({hookStatus: null});
+    return this.hooks
+      .getHookStatus(this.props.currentHookGroupId, this.props.currentHookId)
+    .then(stat => {
```
eliperelman: Indent this group 2 more spaces.

package.json
imbstack: So it turns out this was in `package.json` twice? I think when I `yarn add`'ed the new version of treeview, it removed the duplicate. It's down on line 33 below.

eliperelman: Yeah, just funny that this was here and we didn't notice.

PR #203, opened 2017-03-09
Add tag display to task inspector

wlach:
The tag information is now useful for debugging some types of things
(e.g. action tasks), so let's include it in the task inspector. We'll
use the same code display that we do for the environment.

### DIFF
--- a/src/lib/ui/taskinfo.jsx
+++ b/src/lib/ui/taskinfo.jsx
@@ -132,6 +132,26 @@ const TaskInfo = React.createClass({
             </tr>
 
             <tr>
+              <td>Tags</td>
+              <td>
+                <table className="tag-table">
+                  <tr>
+                    <th>Tag</th><th>Value</th>
+                  </tr>
+                  {
+                    Object
+                      .entries(task.tags)
+                      .map(([key, value]) => (
+                        <tr key={key}>
+                          <td>{key}</td><td>{value}</td>
+                        </tr>
+                      ))
+                  }
+                </table>
+              </td>
+            </tr>
+
+            <tr>
               <td>Requires</td>
               <td>
                 This task will be scheduled when <em>dependencies</em> are
--- a/src/lib/ui/taskinfo.less
+++ b/src/lib/ui/taskinfo.less
@@ -14,4 +14,10 @@
   }
 }
 
-
+.tag-table {
+  border: 1px solid #424242;
+  td, th {
+    border: 1px solid #424242;
+    padding: 4px;
+  }
+}


wlach:
Screenshot screen shot 2017-03-09 at 4 36 19 pm

wlach:
@eliperelman is this something you could review?

eliperelman:
@wlach yep! Are the tags always key value pairs?

wlach:
I believe so, since the "tags" property is just an object.

djmitche:
It is {string: string} -- https://docs.taskcluster.net/reference/platform/queue/docs/task-schema

djmitche:
Could we use something less JSON, like a dd?

eliperelman:
Agreed, while the JSON looks nice, I'd rather see something like:

| Tag | Value |
| :- | :- |
| **test-type** | mochitest |
| **createdForUser** | eoger@mozilla.com |

```md
| Tag | Value |
| :- | :- |
| **test-type** | mochitest |
| **createdForUser** | eoger@mozilla.com |
```

wlach:
Yeah, a table sounds good. Let me try that. :)

wlach:
(already tried the dd thing and it didn't work well, at least with bootstrap's default formatting)

wlach:
Hmm, so by default an embedded table inherits the style of the wrapping table, which has some things we don't want (like bolding the first element). It also generally doesn't look good using bootstrap's CSS styles.

I'd rather not spend 4 hours coming up with the perfect set of CSS styling for this. Do you have any suggestions?

screen shot 2017-03-10 at 7 21 00 pm

djmitche:
To be honest I can live with the table in that screenshot :)

wlach:
After chewing on it over the weekend I realized that doing some basic markup / CSS wouldn't be so bad. Put it down to me being tired and burnt out late on a Friday. Pull request updated, here's a screenshot:

screen shot 2017-03-13 at 2 05 24 pm

Thoughts?

djmitche:
I'm happy with that, too.  @eliperelman ?

src/lib/ui/taskinfo.jsx
```diff
@@ -132,6 +132,23 @@ const TaskInfo = React.createClass({
...
 
             <tr>
+              <td>Tags</td>
+              <td>
+                <table className="tag-table">
+                  <tr>
+                    <th>Tag</th><th>Value</th>
+                  </tr>
+                  {
+                    _.map(task.tags, (value, key) => (
```
eliperelman: We should be able to use `Object.entries` here:

```js
Object
  .entries(task.tags)
  .map(([key, value]) => (
    <tr key={key}>
      <td>{key}</td><td>{value}</td>
    </tr>
  ))
```

src/lib/ui/taskinfo.less
```diff
@@ -14,4 +14,10 @@
...
 }
 
-
+.tag-table {
+  border: 1px solid #424242;
+  td, th {
+    border: 1px solid #424242;
+    padding: 4px;
+  }
+}
```
eliperelman: Insert blank line at end of file.

wlach:
Made requested changes.

src/lib/ui/taskinfo.jsx
```diff
@@ -132,6 +132,25 @@ const TaskInfo = React.createClass({
...
+                  <tr>
+                    <th>Tag</th><th>Value</th>
+                  </tr>
+                  {
+                    Object
+                      .entries(task.tags)
+                      .map(([key, value]) => (
+                        <tr>
+                          <td>{key}</td><td>{value}</td>
+                        </tr>))
```
eliperelman: Nit: move these parents to the next line even with `.map`.

wlach: I'm not sure what you mean here. :)

PR #201, opened 2017-03-02
Quick-Start Enhancement

owlishDeveloper:
This adds new feature to the Quick-Start page: users can check if tc-gh integration is installed for a repo by entering the names of the org and the repo and pressing a button.

Basically, I need feedback on the implementation of the feature (usage of state and so on). Thank you!

@djmitche @eliperelman @imbstack r?

PR #204, opened 2017-03-14
Update QuickStart commands

djmitche:
This adds `git config advice.detachedHead false` (which avoids the
detached-head warning) and also fixes up the golang command so that it
doesn't always build tc-cli.

@owlishDeveloper

### DIFF
--- a/src/quickstart/YamlCreator.jsx
+++ b/src/quickstart/YamlCreator.jsx
@@ -55,27 +55,27 @@ const cmdDirectory = {
     '/bin/bash',
     '--login',
     '-c',
-    'git clone {{event.head.repo.url}} repo && cd repo && git checkout {{event.head.sha}} && npm install . && npm test',
+    'git clone {{event.head.repo.url}} repo && cd repo && git config advice.detachedHead false && git checkout {{event.head.sha}} && npm install . && npm test',
   ],
   'rail/python-test-runner': [
     '/bin/bash',
     '--login',
     '-c',
-    'git clone {{event.head.repo.url}} repo && cd repo && git checkout {{event.head.sha}} && python setup.py test',
+    'git clone {{event.head.repo.url}} repo && cd repo && git config advice.detachedHead false && git checkout {{event.head.sha}} && python setup.py test',
   ],
   'jimmycuadra/rust:latest': [
     '/bin/bash',
     '--login',
     '-c',
     'git clone {{event.head.repo.url}} repo && cd repo' +
-    '&& git checkout {{event.head.sha}} && rustc --test unit_test.rs && ./unit_test',
+    '&& git config advice.detachedHead false && git checkout {{event.head.sha}} && rustc --test unit_test.rs && ./unit_test',
   ],
   'golang:1.8': [
     '/bin/bash',
     '--login',
     '-c',
-    'go get -t github.com/taskcluster/taskcluster-cli/... &&' +
-    ' cd  /go/src/github.com/taskcluster/taskcluster-cli make && go test ./...',
+    'go get -t github.com/owner/repo/... &&' +
+    ' cd  /go/src/github.com/owner/repo && make && go test ./...',
   ],
 };
 


src/quickstart/YamlCreator.jsx
```diff
@@ -55,27 +55,27 @@ const cmdDirectory = {
...
-    '&& git checkout {{event.head.sha}} && rustc --test unit_test.rs && ./unit_test',
+    '&& git config advice.detachedHead false && git checkout {{event.head.sha}} && rustc --test unit_test.rs && ./unit_test',
   ],
   'golang:1.8': [
     '/bin/bash',
     '--login',
     '-c',
-    'go get -t github.com/taskcluster/taskcluster-cli/... &&' +
-    ' cd  /go/src/github.com/taskcluster/taskcluster-cli make && go test ./...',
+    'go get -t github.com/owner/repo/... &&' +
```
owlishDeveloper: (oh deer! I can't believe I actually put taskcluster-cli in there!!🤕)

do you think we could use {{event.head.repo.url}} here? Or other kind of placeholders? Otherwise it might be confusing to some people

djmitche:
Maybe "{{ event.base.user.login }}" and "{{ event.base.repo.name }}" ?

owlishDeveloper:
Yeah! Only, releases don't have 'base' things... We can either add, or if it's OK for 'http://' stuff to be in the link in that command, it can just be {{event.head.repo.url}} (now that I think of it, the more correct one would probably be {{event.base.repo.url}}... which releases again don't have, but that can be added)

djmitche:
Oh, actually, this is still totally wrong -- `go get` is going to check out the master branch of the upstream repo anyway.

In any case, for a pull request we definitely want `base`, since that's part of the gopath (for example, when I'm hacking on https://github.com/djmitche/taskcluster-cli, it's checked out at `$GOPATH/github.com/taskcluster/taskcluster-cli`).  But for releases, we want head.  Ugh.

Maybe we should just omit golang entirely.  @petemoore do you have ideas here?  Is there some way to tell `go get` to use a particular repository and ref that is different from that under which it checks out the source?

owlishDeveloper:
Nooo, don't omit golang! I think of all things THAT language should be there, because it's kinda confusing. I think, instead we should provide one example for the most generic use case. Say, we want to build on PRs. And clarify that use case on Quick-Start (like, you select Go, and magic div appears underneath with brief explanation and link to the docs). ?

djmitche:
If I wait until your owner/repo check patch lands, I can use that
information to fill in these values..

owlishDeveloper:
You mean the API itself or this PR https://github.com/taskcluster/taskcluster-tools/pull/201 or patch correcting the API's behavior (it still doesn't work for taskcluster org)?

djmitche:
I mean #201.

owlishDeveloper:
It's on its way! I just implemented Eli's suggestions

imbstack:
#201 is landed and deployed! Looks good too :)

owlishDeveloper:
Yay! The only thing left for me is to figure out why does it think that taskcluster/taskcluster-github repo "is not set". I'll be on it as soon as I finish gnawing on test challenges here!

src/quickstart/YamlCreator.jsx
```diff
@@ -55,27 +55,27 @@ const cmdDirectory = {
...
+    ' cd  /go/src/github.com/owner/repo && make && go test ./...',
```
petemoore: I think it isn't standard to use `make` - maybe `go install` would be a more suitable default?

src/quickstart/YamlCreator.jsx
petemoore: I think using `base` rather than `head` would be better, as in go you typically want to `go get` the upstream repo, and then fetch and checkout from your own remote, in order that the package names respect the upstream package names. Alternatively `git clone` from the `head` but using the directory path from the `base`. It is kinda finickity in go. Another new age problem (since vendoring got added to the language) is that you typically don't want to `go get` repos that don't have vendored folders checked in. In that case you probably need the vendor tool to perform the checkout, or just use `git clone`.

My suggestion would be to use a native `git clone` of the `head` repo, at the directory path of the `base` repo - unfortunately, the identifier `{{event.base.repo.url}}` is going to need some parsing to achieve that, unless the the user/org and repo name can be fetched some other way.

eliperelman:
Bump.

djmitche:
I think @imbstack is going to wrap this pu..

imbstack:
This is continued in #211

PR #207, opened 2017-04-04
Add a taskcluster-branded spinner

imbstack:
I always wanted to see if we could do something like this. Is it nice looking to you? It just replaces the default spinner with an animated tc logo svg that pulses a bit.

Logo is viewable at https://github.com/taskcluster/taskcluster-tools/blob/4c94a7ce21d3797f58a8e089cc860444c5b1b9d4/src/lib/assets/taskcluster-spinner.svg

tests are failing now because of https://github.com/nodejs/nodejs.org/issues/1191 afaict. I'll rerun them tomorrow when that is fixed.


djmitche:
👍

djmitche:
What if the inside pulsed from blue to gray?

imbstack:
@eliperelman any further thoughts on this? I don't remember what we decided to do now :p

eliperelman:
The light-gray on darker-gray isn't very high contrast. Having some color, or just spinning the original logo would help.

imbstack:
Switched to black+white rather than gray+gray!

djmitche:
I like it!

eliperelman:
Totally agree! This is :fire:!

PR #208, opened 2017-04-11
Bug 1355090: use index links for public indexed artifacts

djmitche:


djmitche:
Whoops, sorry @jonasfj I forgot I flagged you for review too.  It's only code :)  Let me know if you see anything and I can follow up.

PR #209, opened 2017-04-26
Bug 1357802 - Show task.extra in task inspector

swapneshks:
Add fix for displaying "Extra" from task.extra in "Task Details" view

### DIFF
--- a/src/lib/ui/taskinfo.jsx
+++ b/src/lib/ui/taskinfo.jsx
@@ -217,6 +217,15 @@ const TaskInfo = React.createClass({
             </tr>
 
             <tr>
+              <td>Extra</td>
+              <td>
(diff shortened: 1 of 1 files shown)


eliperelman:
Is there ever an instance where `task.extra` doesn't exist? cc: @djmitche

djmitche:
According to https://docs.taskcluster.net/reference/platform/taskcluster-queue/docs/task-schema yes, it's optional.

swapneshks:
@eliperelman : task.extra and Object.keys(task.extra).length both do the job. However, as you suggested, since task.extra would be a plain object, it's better to have the latter. Also, apologies for the missing parenthesis, fixed in latest commit.

PR #228, opened 2017-06-30
Take out obsolete top-level metadata attribute

edunham:
@imbstack told me it didn't belong and @djmitche verified that nothing bad
happens when you remove it from a task. I am moderately certain that this
commit removes the spurious metadata section while leaving the necessary one
in place, but someone better at web apps should verify that.

The old thing with an extra metadata field:

    version: 0
    metadata:
      ...
    tasks:
      - provisionerId: '{{ taskcluster.docker.provisionerId }}'
        ...
        payload:
            ...
        metadata:
            ...

The new correct version that this is supposed to emit instead:

    version: 0
    tasks:
      - provisionerId: '{{ taskcluster.docker.provisionerId }}'
        ...
        payload:
            ...
        metadata:
            ...

### DIFF
--- a/src/views/QuickStart/index.js
+++ b/src/views/QuickStart/index.js
@@ -9,12 +9,6 @@ import { info } from './styles.css';
 
 const initialYaml = {
   version: 0,
-  metadata: {
-    name: '',
-    description: '',
-    owner: '{{ event.head.user.email }}',
-    source: '{{ event.head.repo.url }}'
-  },
   tasks: [
     {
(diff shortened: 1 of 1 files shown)


edunham:
relatedly the format i tried to change it to is what you have in the docs on it, https://docs.taskcluster.net/reference/integrations/taskcluster-github/docs/usage

edunham:
Amended commit to also take out the unhelpful explanation and form fields and assorted references to those values.

PR #229, opened 2017-07-02
Automatically refresh site if remote commit version changes

eliperelman:
Summary:

* Adds a new plugin which outputs a `COMMITHASH` file to the bundle (which would be generated in Heroku)
* Entry point fetches this file every 5 minutes
* Page refreshes if hash has changed

Helps mitigate issues with users keeping tools pages open in their browser with a new version removing old bundles.


djmitche:
What if this refreshes while I have a textarea full of content?

eliperelman:
@djmitche *poof*

djmitche:
Could we shift to fetch the file on route change (with a 5-minute cache)?  I feel like losing data on a route change is expected, but having a tab silently delete the task definition I've been working on for a half-hour sounds frustrating.

djmitche:
c.f. https://bugzilla.mozilla.org/show_bug.cgi?id=1403575

eliperelman:
Closing as we need to figure out a better state persistence strategy before doing this.

PR #237, opened 2017-07-05
Fixes scroll on linux

jonasfj:


src/App/styles.css
```diff
@@ -38,7 +38,6 @@
 
 :global(html, body) {
   height: 100vh;
-  width: 100vw;
```
eliperelman: Like we discussed on IRC, can you switch this to instead add `max-width: 100%;`?

PR #243, opened 2017-07-07
Shrink space used for non-sense in task-inspector

jonasfj:
https://screenshots.firefox.com/1qopUqm2viS0SVSy/localhost

No idea if it's elegant on mobile... 

We should probably move the "notify me on changes" check box next to the progress bar... to make it even better...

Also I'm consistently confused by runs/artifacts/actions, I'm not sure how the navigation should work but a top-level tab bar isn't good... Maybe something like: https://screenshots.firefox.com/HDH8Rnott22qDIes/wireframe.cc  (https://wireframe.cc/P7EQQ7)
Where the list of task in a task-graph is a side-bar that slides in/out and overlays the task-inspector...

### DIFF
--- a/src/views/UnifiedInspector/SearchForm.js
+++ b/src/views/UnifiedInspector/SearchForm.js
@@ -87,13 +87,12 @@ export default class SearchForm extends React.PureComponent {
         <Col lg={5} md={5} sm={4}>
           <FormGroup validationState={valid.taskGroupId ? null : 'error'} style={{ marginRight: 0 }}>
             <InputGroup>
-              <InputGroup.Addon style={{ textAlign: 'left' }}>TaskGroupId</InputGroup.Addon>
+              <InputGroup.Addon style={{ textAlign: 'left' }}>Task Group ID</InputGroup.Addon>
               <FormControl
                 type="text"
                 placeholder="Enter a task group ID"
                 value={taskGroupIdInput}
-                onChange={this.handleTaskGroupIdInputChange}
-                style={{ fontFamily: 'monospace' }} />
+                onChange={this.handleTaskGroupIdInputChange} />
             </InputGroup>
           </FormGroup>
         </Col>
@@ -101,19 +100,18 @@ export default class SearchForm extends React.PureComponent {
         <Col lg={5} md={5} sm={4}>
           <FormGroup validationState={valid.taskId ? null : 'error'} style={{ marginRight: 0 }}>
             <InputGroup>
-              <InputGroup.Addon style={{ textAlign: 'left' }}>TaskId</InputGroup.Addon>
+              <InputGroup.Addon style={{ textAlign: 'left' }}>Task ID</InputGroup.Addon>
               <FormControl
                 type="text"
                 placeholder="Enter a task ID or choose one from Tasks"
                 value={taskIdInput}
-                onChange={this.handleTaskIdInputChange}
-                style={{ fontFamily: 'monospace' }} />
+                onChange={this.handleTaskIdInputChange} />
             </InputGroup>
           </FormGroup>
         </Col>
 
         <Col lg={2} md={2} sm={4} style={{ padding: 0 }}>
-          <ButtonGroup>
+          <ButtonGroup style={{ marginBottom: 15 }}>
             <Button
               type="submit"
               disabled={!valid.form}>
--- a/src/views/UnifiedInspector/SearchForm.js
+++ b/src/views/UnifiedInspector/SearchForm.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { array, func, string } from 'prop-types';
-import { Col, Form, FormGroup, InputGroup, FormControl, Button, ButtonGroup, DropdownButton, MenuItem } from 'react-bootstrap';
+import { Col, Form, FormGroup, InputGroup, FormControl, ButtonGroup, DropdownButton, MenuItem } from 'react-bootstrap';
 import { LinkContainer } from 'react-router-bootstrap';
 import Icon from 'react-fontawesome';
 import { VALID_TASK } from '../../utils';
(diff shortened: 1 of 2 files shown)


djmitche:
I suggested a side-bar slide-out too.  I think the current organization is moderately *usable* but massively confuses a bunch of things into interdependent tabs.

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -84,74 +84,72 @@ export default class SearchForm extends React.PureComponent {
...
+                onChange={this.handleTaskIdInputChange}
+                style={{ fontFamily: 'monospace' }} />
             </InputGroup>
           </FormGroup>
         </Col>
 
-        <Col lg={2} md={2} sm={3}>
-          <Col sm={12} style={{ marginBottom: 15, padding: 0 }}>
+        <Col lg={2} md={2} sm={4} style={{ padding: 0 }}>
+          <ButtonGroup>
```
eliperelman: Add the `style={{ marginBottom: 15 }}` here. In order to get full width in its container, this also needs a `justified` prop.

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -84,74 +84,72 @@ export default class SearchForm extends React.PureComponent {
...
             <Button
```
eliperelman: When switching the `ButtonGroup` to `justified`, this will no longer render correctly on small screens, since buttons don't support table-cell rendering. You'll need to switch this particular button to an anchor:

```jsx
<ButtonGroup justified style={{ marginBottom: 15 }}>
  <a
    onClick={valid.form ? this.handleSubmit : null}
    className={`btn btn-default${valid.form ? '' : ' btn-disabled'}`}>
    <Icon name="search" /> Inspect
  </a>
```

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -84,74 +84,72 @@ export default class SearchForm extends React.PureComponent {
 
     return (
       <Form onSubmit={this.handleSubmit} horizontal>
-        <Col lg={10} md={10} sm={9}>
-          <FormGroup validationState={valid.taskGroupId ? null : 'error'}>
+        <Col lg={5} md={5} sm={4}>
+          <FormGroup validationState={valid.taskGroupId ? null : 'error'} style={{ marginRight: 0 }}>
             <InputGroup>
-              <InputGroup.Addon style={{ textAlign: 'left' }}>Task Group</InputGroup.Addon>
+              <InputGroup.Addon style={{ textAlign: 'left' }}>TaskGroupId</InputGroup.Addon>
```
eliperelman: This is just language, so use `Task Group ID` instead of `TaskGroupId`.

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -84,74 +84,72 @@ export default class SearchForm extends React.PureComponent {
...
             </InputGroup>
           </FormGroup>
+        </Col>
 
-          <FormGroup validationState={valid.taskId ? null : 'error'}>
+        <Col lg={5} md={5} sm={4}>
+          <FormGroup validationState={valid.taskId ? null : 'error'} style={{ marginRight: 0 }}>
             <InputGroup>
-              <InputGroup.Addon style={{ textAlign: 'left' }}>Task</InputGroup.Addon>
+              <InputGroup.Addon style={{ textAlign: 'left' }}>TaskId</InputGroup.Addon>
```
eliperelman: This is just language, so use `Task ID` instead of `TaskId`.

eliperelman:
@dustin yup, that's known, and won't be fully addressed til the redesign. I'll try and update some text to make it clearer, but this particular layout has no plans for being changed right now.

src/views/UnifiedInspector/SearchForm.js
jonasfj: I always refer to them as `taskGroupId` and `taskId` in text as this is their meta-variable names... But I don't care enough...

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -84,74 +84,72 @@ export default class SearchForm extends React.PureComponent {
...
               <FormControl
                 type="text"
-                placeholder="Enter a task ID or choose one from Tasks, e.g. 8U3xVyssSBuinaXwRgJ_qQ"
+                placeholder="Enter a task ID or choose one from Tasks"
                 value={taskIdInput}
-                onChange={this.handleTaskIdInputChange} />
+                onChange={this.handleTaskIdInputChange}
+                style={{ fontFamily: 'monospace' }} />
```
jonasfj: Really, it's not an text field, it's an identifier input field...

src/views/UnifiedInspector/SearchForm.js
jonasfj: justified doesn't work...

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -84,74 +84,70 @@ export default class SearchForm extends React.PureComponent {
...
                 value={taskIdInput}
                 onChange={this.handleTaskIdInputChange} />
             </InputGroup>
           </FormGroup>
         </Col>
 
-        <Col lg={2} md={2} sm={3}>
-          <Col sm={12} style={{ marginBottom: 15, padding: 0 }}>
+        <Col lg={2} md={2} sm={4} style={{ padding: 0 }}>
+          <ButtonGroup style={{ marginBottom: 15 }}>
```
eliperelman: I tested this out locally and if you use `justified` here, along with converting the first `Button` to an anchor, it should work. What was wrong with it?

src/views/UnifiedInspector/SearchForm.js
eliperelman: For reference, this is what I used:

```jsx
<ButtonGroup justified style={{ marginBottom: 15 }}>
  <a
    onClick={valid.form ? this.handleSubmit : null}
    className={`btn btn-default${valid.form ? '' : ' btn-disabled'}`}>
    <Icon name="search" /> Inspect
  </a>
  <DropdownButton ...
```

jonasfj:
Thanks! I tried a million and two things and couldn't get justified to
work... what gave you the idea to use `<a>`?

--
Regards Jonas Finnemann Jensen.

2017-07-07 13:07 GMT-07:00 Eli Perelman <notifications@github.com>:

eliperelman:
@jonasfj haha yeah, I looked through some issue with rendering, and turns out buttons don't like the table-cell rendering from bootstrap, so you have to use `a`. But then `a` doesn't truly support `disabled`, so you have to fake it there. :/

Found it: https://github.com/twbs/bootstrap/issues/9249

eliperelman:
Forgot to say, yes the new layout we have proposed will use a sidebar.

PR #248, opened 2017-07-10
Improving titles of inspector tabs

eliperelman:
Attempting to improve the titles of the inspector tabs so its a *little* clearer about the hierarchy.

tabs

Thoughts @djmitche?


djmitche:
It's an improvement :)

eliperelman:
@djmitche if you have ideas for other phrasing I'm open. Just no layout changes. 😄

djmitche:
If limited to phrasing changes, this is as good as I can imagine..

eliperelman:
OK then, we'll ship this for now.

PR #249, opened 2017-07-12
Bug 1380093 - Fix latest run label in dropdown

helfi92:
For some reason, in production only, `[...status.runs].reverse()` returns the same array. `status.runs.reverse()` fixes it. There is no need to use the spread operators here since status.runs will always be an array. Also, there is an if statement right above it to make sure `status.runs` has a `length` property. When I run the [Python SimpleHTTPServer with HTML5 pushState](https://gist.github.com/jtangelder/e445e9a7f5e31c220be6) in the build directory locally, it works.

### DIFF
--- a/src/views/UnifiedInspector/RunsMenu.js
+++ b/src/views/UnifiedInspector/RunsMenu.js
@@ -42,7 +42,7 @@ export default class RunsMenu extends React.PureComponent {
 
     return (
       <NavDropdown title={this.getRunsTitle()} active={active} id="runs-dropdown">
-        {status.runs.reverse().map(({ runId }, index) => (
+        {Array.from(status.runs).reverse().map(({ runId }, index) => (
           <LinkContainer to={`/groups/${taskGroupId}/tasks/${taskId}/runs/${runId}`} key={`runs-menu-item-${index}`}>
             <MenuItem>Task Run {runId}{index === 0 ? ' (Latest)' : ''}</MenuItem>
           </LinkContainer>
--- a/src/views/UnifiedInspector/RunsMenu.js
+++ b/src/views/UnifiedInspector/RunsMenu.js
@@ -42,7 +42,7 @@ export default class RunsMenu extends React.PureComponent {
 
     return (
       <NavDropdown title={this.getRunsTitle()} active={active} id="runs-dropdown">
-        {[...status.runs].reverse().map(({ runId }, index) => (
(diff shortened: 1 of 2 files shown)


src/views/UnifiedInspector/RunsMenu.js
```diff
@@ -42,7 +42,7 @@ export default class RunsMenu extends React.PureComponent {
 
     return (
       <NavDropdown title={this.getRunsTitle()} active={active} id="runs-dropdown">
-        {[...status.runs].reverse().map(({ runId }, index) => (
+        {status.runs.reverse().map(({ runId }, index) => (
```
djmitche: isn't this mutating status?

helfi92: @djmitche You're right. We can use `Array.from(status.runs)` instead. It will return a new copy of the array.

yarn.lock
```diff
@@ -47,8 +47,8 @@ acorn-jsx@^3.0.0:
     acorn "^3.0.4"
 
 acorn@>=2.5.2, acorn@^5.0.0, acorn@^5.0.1:
-  version "5.0.3"
-  resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.0.3.tgz#c460df08491463f028ccb82eab3730bf01087b3d"
+  version "5.1.1"
+  resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.1.1.tgz#53fe161111f912ab999ee887a90a0bc52822fd75"
```
djmitche: Why the dependency changes?  (sorry if I missed this on my first comment..)

The `RunsMenu.js` diff looks fine, so just removing the `yarn.lock` diff would make this an r+

djmitche:
```
@^1.1.2@^2.0.0e@^1.0.0error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```
weird, and this didn't seem to cause an issue [on the latest push to master](https://travis-ci.org/taskcluster/taskcluster-tools/builds/252116215?utm_source=github_status&utm_medium=notification).  That's a little over my yarn-related head.

helfi92:
I'll try to fix it, one moment

djmitche:
Oh, not weird -- there is still a yarn.lock diff in this PR.

helfi92:
Should be good now :)

PR #251, opened 2017-07-12
Bug 1380098 - Make ArtifactList aware of credentials

helfi92:
Create an interactive task from any task and head over to "Run Artifacts" tab view. ArtifactList component was not aware of the user credentials so the shell.sh artifact displayed an "I" cursor and wouldn't allow you to click on it. This PR solves this issue.


helfi92:
Correct

PR #252, opened 2017-07-14
Bug 1380540 - Remove illusion of truncated log

helfi92:
When you click "Follow log", the logviewer will show illusion of a truncated log. Currently, a user needs to scroll down the main view (not the logviewer) to see the rest of the log, therefore adding a layer of confusion. This pull-request sets the height of the logviewer to sit perfectly at the bottom of the screen.

### DIFF
--- a/src/views/UnifiedInspector/LogView.js
+++ b/src/views/UnifiedInspector/LogView.js
@@ -31,16 +31,19 @@ export default class LogView extends React.PureComponent {
       streaming,
       follow: streaming,
       isFullscreen: false,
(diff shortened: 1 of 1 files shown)


helfi92:
This is simply an enhancement to the current way we display the logviewer (not something broken in tools) so no need to merge now. We can wait until @eliperelman is back :)

helfi92:
Safe to merge? :)

eliperelman:
Yep, as long as all the status checks pass, you are good to merge.

PR #254, opened 2017-07-14
Bug 1359468 - Adjust all timestamps relative to created

helfi92:


src/views/TaskCreator/TaskCreator.js
```diff
@@ -13,8 +13,8 @@ const localStorageKey = 'tasks:create';
 const defaultTask = {
   provisionerId: 'aws-provisioner-v1',
   workerType: 'tutorial',
-  created: null,
-  deadline: null,
+  created: moment().toISOString(),
+  deadline: moment().add(1, 'hours').toISOString(),
```
jonasfj: Using `taskcluster.fromNow()` that way you're sure to get it right...

jonasfj: Also, let's change the default deadline to 3 hours...

src/views/TaskCreator/TaskCreator.js
```diff
@@ -74,16 +74,30 @@ export default class TaskCreator extends React.PureComponent {
...
+        case 'object':
+          return Array.isArray(obj) ?
+            obj.map(iter) :
+            Object
+              .entries(obj)
+              .reduce((o, [key, value]) => ({ ...o, [key]: iter(value) }), {});
+
+        case 'string':
+          return jsonDate.test(obj) ?
+            moment(obj).add(offset).toISOString() :
```
jonasfj: If you want to avoid moment, you can just do: `offset = Date.now() - (new Date(task.created)).getTime();`
and then use: `new Date(offset + (new Date(obj)).getTime())`.

Just a nit...

src/views/TaskCreator/TaskCreator.js
helfi92: Changed the deadline to 3. @jonasfj Is it ok to not use fromNow if we're going to keep using moment?

jonasfj: sure... as long as it produces the right format... ship it...

PR #262, opened 2017-07-18
Bug 1381740 - Make UI responsive with creating/deleting roleIds and clientIds

helfi92:
Problems with `auth/roles`:
* Creating a new roleId is not refreshing the list of roles. You have to manually click the "refresh" button or refresh the page.
* Creating a new roleId should add the roleId to the URL for deep linkage. Right now, it's not doing that so if you create a new roleId then refresh the page, it won't select the newly created role.
* Deletion of a roleId is not working. See [Bug 1381740](https://bugzilla.mozilla.org/show_bug.cgi?id=1381740).

This pull-request tackles and fixes those problems.

### DIFF
--- a/src/components/ClientEditor/index.js
+++ b/src/components/ClientEditor/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { func } from 'prop-types';
 import {
-  Alert, OverlayTrigger, Tooltip, Modal, FormGroup, ControlLabel, FormControl, Button, Glyphicon, ButtonToolbar
+  OverlayTrigger, Tooltip, Modal, FormGroup, ControlLabel, FormControl, Button, Glyphicon, ButtonToolbar
 } from 'react-bootstrap';
 import moment from 'moment';
 import { assoc } from 'ramda';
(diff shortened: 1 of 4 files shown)


helfi92:
You're absolutely right. It needs to change as well. I'm on it!

helfi92:
Problems with `auth/clients`:
* Creating a new clientId should add the clientId to the URL for deep linkage. Right now, it's not doing that so if you create a new clientId then refresh the page, it won't select the newly created role.
* Deletion of a clientId does not update the URL and does not refresh the list of clients.

Commit [3c45515](https://github.com/taskcluster/taskcluster-tools/pull/262/commits/3c455151c6ebb0bf9b0d896fb153f3e77f6b185f) fixes these issues.

src/components/ClientEditor/index.js
```diff
@@ -70,9 +71,9 @@ export default class ClientEditor extends React.PureComponent {
       });
     }
 
-    // Load currentClientId, but avoid doing so after creating a new client
     if (this.state.client && this.state.client.clientId === props.currentClientId) {
-      this.setState({
+      return this.setState({
+        client: await props.auth.client(props.currentClientId),
```
djmitche: This should probably be wrapped in a try, like the call to `props.auth.client` below.

helfi92: Agreed. I pushed a fix. Will merge once Travis completes

PR #264, opened 2017-07-28
Use ISO timestamps in TimeInput

djmitche:
The existing use of moment was broken, treating any change to the
datestamp as invalid.


djmitche:
Thanks!  It's not the best fix (deleting decimal digits causes a red border) but at least it works..

PR #265, opened 2017-07-31
Bug 1385373: [sign-in] Handle null certificates properly. r=Eli

walac:
Manual sign-in is busted because there are some places in the code where
null certificates are passed to JSON.parse, throwing an exception.

We check if the certifcates are falsy values and handle them
accordingly.

### DIFF
--- a/src/App/index.js
+++ b/src/App/index.js
@@ -87,13 +87,14 @@ export default class App extends React.Component {
   };
 
   loadCredentials() {
-    const credentials = JSON.parse(localStorage.getItem('credentials'));
+    const storedCredentials = localStorage.getItem('credentials');
 
     // We have no credentials
-    if (!credentials) {
+    if (!storedCredentials) {
       return { credentials: null };
     }
 
+    const credentials = JSON.parse(storedCredentials);
     const { certificate } = credentials;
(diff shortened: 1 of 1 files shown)


src/App/index.js
```diff
@@ -87,13 +87,15 @@ export default class App extends React.Component {
   };
 
   loadCredentials() {
-    const credentials = JSON.parse(localStorage.getItem('credentials'));
+    let credentials = localStorage.getItem('credentials');
```
eliperelman: Avoid use of `let`:

```js
const storedCredentials = localStorage.getItem('credentials');

if (!storedCredentials) {
  return { credentials: null };
}

const credentials = JSON.parse(storedCredentials);
```

src/App/index.js
```diff
@@ -87,13 +87,15 @@ export default class App extends React.Component {
...
+    const storedCredentials = localStorage.getItem('credentials');
 
     // We have no credentials
-    if (!credentials) {
+    if (!storedCredentials) {
       return { credentials: null };
     }
 
+    const credentials = JSON.parse(storedCredentials);
+
```
eliperelman: Can you remove this extra blank line?

PR #277, opened 2017-08-11
Update log when task, run, or log name changes, not status

### LINKED ISSUE #275: Follow log doesn't work
Ticking the follow log box in "Run Logs" unticks itself when the log is updated.

eliperelman:
Fixes #275.


eliperelman:
I'm not sure how that would happen, but that is a separate issue.

PR #280, opened 2017-08-16
Bug 1355090 - In the indexed artifact browser, use links constructed to include the route, not the taskId

helfi92:


helfi92:
@djmitche Sorry, I should have elaborated. If you look in [src/views/IndexedArtifactBrowser](https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/IndexedArtifactBrowser/ArtifactView.js#L63), the render method returns the ArtifactList component with a prop `namespace`.

Now in the ArtifactList component, it was retrieving an undefined prop `indexNamespace`. To get the sent prop, it needed to get renamed to `namespace`.

djmitche:
You can take the boy out of Python, but you can't take the Python out of the boy..

PR #286, opened 2017-08-23
Prettify logviewer on mobile

helfi92:
This pull-request incorporates 2 changes:
1. Provide the logviewer a minimal height of 400px
- Currently the logviewer height is dynamically calculated so that it sits nicely on the bottom of the screen. On mobile however, the height is really tiny and so you only see one or two lines of the log. To fix this issue, I provided the component a minimal height of 400px, big enough for mobile users to interact with the viewer.

2. Use flexbox to wrap the status label in mobile
- On mobile, the status labels are overflowing the screen horizontally and this makes the eyes bleed. Flexbox is used to fix this issue. Below are the before and after visuals.

## Before
screen shot 2017-08-23 at 11 27 19 am

screen shot 2017-08-23 at 11 25 57 am

## After
screen shot 2017-08-23 at 11 28 10 am

### DIFF
--- a/src/views/UnifiedInspector/GroupProgress.js
+++ b/src/views/UnifiedInspector/GroupProgress.js
@@ -107,9 +107,12 @@ export default class GroupProgress extends React.PureComponent {
           ))}
         </ProgressBar>
 
-        <div>
+        <div style={{ display: 'flex', justifyContent: 'flex-start', flexWrap: 'wrap' }}>
           {groups.map((group, index) => (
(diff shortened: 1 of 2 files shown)


eliperelman:
We used to have a min height of 800. Any reason we are using 400 over the 800?

helfi92:
@eliperelman On a mobile device, if you set the viewer height to 800px, it will take the full screen height. That being said, if you "Follow" a log and then decide to trigger an action (Cancel Task for example), it becomes very annoying to scroll back up because you will be scrolling up the viewer component, not the parent container. That was my reasoning behind 400px.

PR #288, opened 2017-08-23
Add jump to line feature

helfi92:
1.

screen shot 2017-08-23 at 3 55 34 pm

2. 

screen shot 2017-08-23 at 3 53 22 pm

I'm thinking we can add this feature and see if people start using it. Later we can include a keyboard shortcut to open this modal (e.g., cmd + L)

### DIFF
--- a/src/views/UnifiedInspector/LogView.js
+++ b/src/views/UnifiedInspector/LogView.js
@@ -95,7 +95,8 @@ export default class LogView extends React.PureComponent {
   jump = () => {
     this.setState({
       lineNumber: this.state.lineNumber,
-      jump: true
+      jump: true,
+      follow: false
     });
   };
 
--- a/src/views/UnifiedInspector/LogView.js
+++ b/src/views/UnifiedInspector/LogView.js
@@ -1,11 +1,12 @@
 import React from 'react';
 import { findDOMNode } from 'react-dom';
-import { Button } from 'react-bootstrap';
+import { Button, FormControl } from 'react-bootstrap';
 import Icon from 'react-fontawesome';
 import { arrayOf, oneOfType, string, object, number, func } from 'prop-types';
 import { LazyLog, LazyStream, ScrollFollow } from 'react-lazylog';
 import { isNil } from 'ramda';
 import fscreen from 'fscreen';
+import ModalItem from '../../components/ModalItem';
 
 const VIEWER_HEIGHT_MIN = 400;
 const buttonStyle = {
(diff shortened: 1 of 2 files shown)


djmitche:
How will we know if people use it?

helfi92:
For now, we could rely on whether people mention it on the taskcluster irc group and then later we could perhaps add some sort of heatmap analytics, showing us where end users are clicking on a page. Nevertheless, the jump to line feature should not bother anyone :)

gregarndt:
Good question.  This was something hassan and I were talking about today (not about this particular feature though).  We should be able to evaluate how features are used within our tools and make some data driven decisions about what works.

I'm not clear on what the toolchain looks like though.

eliperelman:
We plan on implementing Google Analytics or something similar in the future for usage stats.

helfi92:
@eliperelman Makes sense 👍

src/views/UnifiedInspector/LogView.js
```diff
@@ -152,7 +185,7 @@ export default class LogView extends React.PureComponent {
               url={url}
               height={isFullscreen ? document.documentElement.clientHeight : lazyViewerHeight}
               follow={follow}
-              scrollToLine={!follow && highlight ? scrollToLine : null}
+              scrollToLine={!follow && ((this.state.jump ? lineNumber : null) || (highlight ? scrollToLine : null))}
```
eliperelman: Can you explain why we need 2 pieces of state to drive this? Can we find a way to infer this from `lineNumber` alone?

helfi92: If you only have 1 piece of state (`lineNumber`) to drive this, it will cause a small problem: While modifying the form input in the modal, the logviewer will scroll on every keystroke. I am using the `jump` state to signal that the user clicked "Go to Line". I'm not sure I have a way to infer this from `lineNumber` alone.

eliperelman:
If we get a rebase on this, we can push it forward.

helfi92:
Rebased.

eliperelman:
Hmm, still looks to be a conflict.

helfi92:
Oops, that conflict probably happened after merging the "follow log user preference" PR. Will rebase again and then squash.

helfi92:
Note: Jump to line feature will not work if "Follow Log" is checked.

eliperelman:
Maybe we should disable that button when follow log is checked.

eliperelman:
Or rather we should shut off the follow when the jump is input.

helfi92:
Oh yeah, that's definitely doable! Pushing something now.

PR #289, opened 2017-08-24
Reset task states when task ID changes

eliperelman:
@helfi92 should be a simple enough review.

@bstack do you know if I need to reset the actions and decision if the task ID changes? I thought not, but wanted to confirm.


eliperelman:
w00t

PR #290, opened 2017-08-25
Load artifacts and logs when switching runs

helfi92:
STR: Head over to the Task & Group Inspector. Set the `taskId` to `Ev_1g8xrTXS1thHi7649Ww` and click Inspect. You will notice that Task run 1 has no logs and no artifacts. Switching to run 0 is not updating that list, therefore still showing no logs and no artifacts.

### DIFF
--- a/src/views/UnifiedInspector/Inspector.js
+++ b/src/views/UnifiedInspector/Inspector.js
@@ -111,8 +111,9 @@ export default class Inspector extends React.PureComponent {
       (taskId !== this.state.selectedTaskId || (this.props.taskId && taskId !== this.props.taskId))
     ) {
       this.loadTask(nextProps);
-    } else if (runId && runId !== this.props.runId) {
+    } else if (Number.isInteger(runId) && runId !== this.props.runId) {
       this.setState({ selectedRun: runId });
+      this.loadTask(nextProps);
     } else if (this.state.error && !equal(nextProps.credentials, this.props.credentials)) {
       this.setState({ error: null });
     }
--- a/src/views/UnifiedInspector/Inspector.js
+++ b/src/views/UnifiedInspector/Inspector.js
@@ -79,7 +79,7 @@ export default class Inspector extends React.PureComponent {
     }
   }
 
-  componentWillReceiveProps(nextProps) {
+  async componentWillReceiveProps(nextProps) {
     const { taskGroupId, taskId, runId } = nextProps;
 
     if (taskGroupId !== this.props.taskGroupId) {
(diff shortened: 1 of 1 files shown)


src/views/UnifiedInspector/Inspector.js
```diff
@@ -111,8 +111,9 @@ export default class Inspector extends React.PureComponent {
       (taskId !== this.state.selectedTaskId || (this.props.taskId && taskId !== this.props.taskId))
     ) {
       this.loadTask(nextProps);
-    } else if (runId && runId !== this.props.runId) {
+    } else if (Number.isInteger(runId) && runId !== this.props.runId) {
```
eliperelman: Could you explain why we need this `isInteger` check?

src/views/UnifiedInspector/Inspector.js
```diff
@@ -111,8 +111,9 @@ export default class Inspector extends React.PureComponent {
...
       this.setState({ selectedRun: runId });
+      this.loadTask(nextProps);
```
eliperelman: I can't think of why, from a logical perspective, we would want to reload the entire task whenever we switch runs. When we load the task, we already get all the runs for the task.

eliperelman: We probably only really need the logic from inside `loadTask` which fetches the artifacts:

https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/UnifiedInspector/Inspector.js#L179

src/views/UnifiedInspector/Inspector.js
helfi92: I put an integer check so that runId 0 is considered as `true`. `(runId && runId !== this.props.runId)` will always be `false` when `runId` is 0.

src/views/UnifiedInspector/Inspector.js
helfi92: Agreed

src/views/UnifiedInspector/Inspector.js
```diff
@@ -111,8 +111,13 @@ export default class Inspector extends React.PureComponent {
       (taskId !== this.state.selectedTaskId || (this.props.taskId && taskId !== this.props.taskId))
     ) {
       this.loadTask(nextProps);
-    } else if (runId && runId !== this.props.runId) {
-      this.setState({ selectedRun: runId });
+    } else if (Number.isInteger(runId) && runId !== this.props.runId) {
```
eliperelman: Note that this could also be achieved with `runId !== null`.

PR #292, opened 2017-08-30
Add redirect for action-task loaners

imbstack:
@djmitche and I figured the best way to make these tasks work correctly is to add this in here directly this way.

### DIFF
--- a/src/views/UnifiedInspector/ActionsMenu.js
+++ b/src/views/UnifiedInspector/ActionsMenu.js
@@ -453,7 +453,7 @@ export default class ActionsMenu extends React.PureComponent {
             modalSize={actionData[action].action.schema ? 'large' : null}
             body={this.actionTaskModal(action)}
             onSubmit={this.actionTaskSubmit(action)}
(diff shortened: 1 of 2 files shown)


src/views/UnifiedInspector/ActionsMenu.js
```diff
@@ -467,7 +467,7 @@ export default class ActionsMenu extends React.PureComponent {
             modalSize={actionData[action].action.schema ? 'large' : null}
             body={this.actionTaskModal(action)}
             onSubmit={this.actionTaskSubmit(action)}
-            onComplete={onActionTask}
+            onComplete={onActionTask(action)}
```
djmitche: does this need to be `() => onActionTask(action)`?

imbstack: I think in this case we actually don't want to do that for once. The `this` in `onActionTask` is for the `UnifiedInspector` it comes from rather than the `ActionsMenu`.

PR #294, opened 2017-09-05
Bug 1396315 - Tool for examining worker-types

helfi92:


helfi92:
image 2017-09-05 at 10 27 10 am 3y0j051r0l2p

helfi92:
I pushed a change that adds a tabular view in addition to the grid layout. To switch between the views, there is a switch. In addition, I added stability colors. For the grid layout, the colors will be on the card itself whereas in the table layout, the stability text node is wrapped in a colored label.

screen shot 2017-09-07 at 3 09 44 pm

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -0,0 +1,174 @@
...
+      provisioners: null,
+      error: null
+    };
+  }
+
+  componentWillMount() {
+    this.loadProvisioners();
+  }
+
+  async loadProvisioners(token) {
```
eliperelman: This method has the ability to call `setState` many times depending on the number of continuations that occur, since every call will reset provisioners to an empty array. Try the following:

```js
async loadProvisioners(token) {
  try {
    const { provisioners, continuationToken } = await this.props.queue
      .listProvisioners(token ? { continuationToken: token, limit: 100 } : { limit: 100 };

      this.setState({
        provisioners: this.state.provisioners ?
          this.state.provisioners.concat(provisioners) :
          provisioners
      });

      if (continuationToken) {
        this.loadProvisioners(continuationToken);
      }
  } catch (err) {
    this.setState({
      provisioners: null,
      error: err
    });
  }
}
```

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  componentWillMount() {
+    if (this.props.provisionerId) {
+      this.loadWorkerTypes(this.props);
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.provisionerId !== nextProps.provisionerId) {
```
eliperelman: We also need to check that the provisioner ID wasn't removed from the URL.

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+      this.loadWorkerTypes(nextProps);
+    }
+  }
+
+  async loadWorkerTypes({ provisionerId }, token) {
+    this.setState({ workerTypes: [], loading: true, error: null }, async () => {
```
eliperelman: Unwrap this from resetting the state after every call.

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+      try {
+        const { workerTypes, continuationToken } = await this.props.queue
+          .listWorkerTypes(provisionerId, token ? { continuationToken: token, limit: 100 } : { limit: 100 });
+
+        this.setState({
```
eliperelman: There are a couple calls to `setState` every time this method is called. We should batch the results and make a single call to `setState` after the method has aggregated all the data it needs.

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+          workerTypes: this.state.workerTypes ? this.state.workerTypes.concat(workerTypes) : workerTypes
+        });
+
+        if (continuationToken) {
+          this.loadWorkerTypes(this.props, continuationToken);
+        }
+
+        const awsWorkerTypes = provisionerId === 'aws-provisioner-v1' ?
```
eliperelman: Do not set variables to `undefined`.

```js
const awsWorkerTypes = provisionerId === 'aws-provisioner-v1' && await this.props.awsProvisioner.listWorkerTypeSummaries();
```

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+
+  async getPendingTasks(provisionerId, workerType) {
+    const { pendingTasks } = await this.props.queue.pendingTasks(provisionerId, workerType);
+
+    return pendingTasks;
+  }
+
+  renderDescription = ({ description }) => (
+    <Popover id="popover-trigger-click-root-close" title="Description">
+      {
```
eliperelman:
```jsx
<Markdown>{description || '-'}</Markdown>
```

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+
+    return (
+      <Panel
+        key={`worker-type-grid-${index}`}
+        className={styles.card}
+        header={<Header key={`worker-type-header-${index}`} />} bsStyle={`${stabilityColors[workerType.stability]}`}>
+        <Table fill>
+          <tbody>
+            {
+              this.props.provisionerId === 'aws-provisioner-v1' ?
```
eliperelman:
```jsx
{this.props.provisionerId === 'aws-provisioner-v1' && ['runningCapacity', 'pendingCapacity']
  .map((property, key) => (
    <tr key={`dynamic-data-${key}`}>
      <td>{property}</td>
      <td><Badge>{workerType[property]}</Badge></td>
    </tr>
  ))
}
```

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+    <div className={styles.tabular}>
+      <Table responsive>
+        <thead>
+          <tr>
+            <th>Worker-type</th>
+            <th>Stability</th>
+            <th>Last active</th>
+            <th>Pending tasks</th>
+            {
+              this.props.provisionerId === 'aws-provisioner-v1' ?
```
eliperelman: Use `&&` over `?: null`.

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+                <Label
+                  bsStyle={stabilityColors[workerType.stability]}
+                  style={{ borderRadius: 0, marginLeft: 10, marginBottom: 5 }}>
+                  {workerType.stability}
+                </Label>
+              </td>
+              <td>{moment(workerType.lastDateActive).fromNow()}</td>
+              <td><Badge>{workerType.pendingTasks}</Badge></td>
+              {
+                this.props.provisionerId === 'aws-provisioner-v1' ?
```
eliperelman: Use `&&` over `?: null`.

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,259 @@
...
+              }
+            </tr>
+          ))}
+        </tbody>
+      </Table>
+    </div>
+  );
+
+  renderWorkerType = (workerTypes) => {
+    if (this.props.gridLayout) {
```
eliperelman:
```js
return this.props.gridLayout ?
  workerTypes.map(this.renderGridWorkerType) :
  this.renderTabularWorkerType(workerTypes);
```

package.json
```diff
@@ -56,6 +56,7 @@
     "react-loadable": "^4.0.4",
     "react-router-bootstrap": "^0.24.2",
     "react-router-dom": "^4.1.2",
+    "react-switch-button": "^2.2.2",
```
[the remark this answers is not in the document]
helfi92: Good idea, I like this. Also, react-switch-button didn't seem to be well maintained either.

src/App/index.js
```diff
@@ -188,6 +189,7 @@ export default class App extends React.Component {
               <PropsRoute path="/quickstart" component={QuickStart} credentials={credentials} />
               <PropsRoute path="/aws-provisioner/:workerType?/:currentTab?" component={AwsProvisioner} credentials={credentials} baseUrl="https://aws-provisioner.taskcluster.net/v1" provisionerId="aws-provisioner-v1" routeRoot="/aws-provisioner" />
               <PropsRoute path="/aws-provisioner-staging/:workerType?/:currentTab?" component={AwsProvisioner} credentials={credentials} baseUrl="https://provisioner-staging.herokuapp.com/v1" provisionerId="staging-aws" routeRoot="/aws-provisioner-staging" />
+              <PropsRoute path="/worker-types/:provisionerId?" component={WorkerTypes} credentials={credentials} />
```
djmitche: why isn't this a prefix of the worker dashboard (`/workers/...`)?

helfi92: If we had a `/workers` page, I would of put `/workers/worker-types/:provisionerId?`. Since we don't have a `/workers` page, I thought it would be weird to take the nested approach `/workers/worker-types/...`. @eliperelman What do you think?

eliperelman: I think we can leave it for now. Right now it seems the route scheme is lengthy in the other PR, so merging these 2 routes right now might be a larger fix. Let's fix in a followup.

eliperelman:
I want to make a change in the UI, renaming "Worker-types" to "Worker types". Basically remove the hyphen from the word. It's OK to leave it in the URL, but in the page, let's use the non-hyphenated version.

eliperelman:
Can we also change runningCapacity and friends to be actual words, e.g. "Running capacity"?

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,251 @@
...
+
+      if (continuationToken) {
+        this.loadWorkerTypes(this.props, continuationToken);
+      }
+
+      const awsWorkerTypes = provisionerId === 'aws-provisioner-v1' &&
+        await this.props.awsProvisioner.listWorkerTypeSummaries();
+
+      const workerTypesNormalized = this.state.workerTypes.map(workerType => (
+        Object.assign(
```
eliperelman: You don't have to change it, but you can use object spread here:

```js
this.state.workerTypes.map(workerType => ({
  ...workerType,
  ...(awsWorkerTypes ? find(propEq('workerType', workerType.workerType))(awsWorkerTypes) : {})
})
```

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,251 @@
...
+
+        const stable = {
+          provisionerId: workerType.provisionerId,
+          workerType: workerType.workerType,
+          pendingTasks,
+          stability: workerType.stability,
+          lastDateActive: workerType.lastDateActive
+        };
+
+        const dynamic = provisionerId === 'aws-provisioner-v1' ? {
```
eliperelman: For multiple-line ternaries, put each value in its own line section:

```js
const dynamic = provisionerId === 'aws-provisioner-v1' ?
  {
    runningCapacity: workerType.runningCapacity,
    pendingCapacity: workerType.pendingCapacity
  } ?
  {};
```

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,251 @@
...
+
+  async getPendingTasks(provisionerId, workerType) {
+    const { pendingTasks } = await this.props.queue.pendingTasks(provisionerId, workerType);
+
+    return pendingTasks;
+  }
+
+  renderDescription = ({ description }) => (
+    <Popover id="popover-trigger-click-root-close" title="Description">
+      <Markdown>{description || '`-`'}</Markdown>
```
eliperelman: There shouldn't be backticks around this hyphen: `{description || '-'}`.

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,251 @@
...
+                ))
+              }
+            </tr>
+          ))}
+        </tbody>
+      </Table>
+    </div>
+  );
+
+  renderWorkerType = workerTypes => (
```
eliperelman: No need to wrap in parentheses here:

```js
renderWorkerType = workerTypes => this.props.gridLayout ?
  workerTypes.map(this.renderGridWorkerType) :
  this.renderTabularWorkerType(workerTypes);
```

src/views/WorkerTypes/WorkerTypeTable.js
```diff
@@ -0,0 +1,251 @@
...
+  );
+
+  sort = (a, b) => {
+    if (this.props.lastActive) {
+      return moment(a.lastDateActive).diff(moment(b.lastDateActive)) < 0 ? 1 : -1;
+    }
+
+    if (this.props.orderBy) {
+      return a[this.props.orderBy] - b[this.props.orderBy] < 0 ? 1 : -1;
+    }
```
eliperelman: What is the sort if neither of these conditions is hit? As a default, consider returning 0.

src/views/WorkerTypes/WorkerTypeTable.js
helfi92: Love it your way. `Object.assign` makes my eyes bleed for some reason.

src/views/WorkerTypes/WorkerTypeTable.js
helfi92: `<Markdown>'-'</Markdown>` shows `-`
`<Markdown>{'-'}</Markdown>` shows `•`

Having it like it currently is shows `-`. That's the only reason why I have it this way.

helfi92:
@eliperelman Changes made :)

eliperelman:
Merge when ready. :shipit:

PR #296, opened 2017-09-08
Bug 1396318 - Tool for examining workers

helfi92:

PR #297, opened 2017-09-12
Bug 1396318 - Tool for examining workers

helfi92:
Continues #296


eliperelman:
This PR has conflicts, please resolve and re-request review.

helfi92:
Conflicts removed.

src/views/Worker/WorkerManager.js
```diff
@@ -0,0 +1,193 @@
...
+    const { provisionerId, workerType, workerGroup, workerId } = this.state.worker;
+
+    try {
+      const worker = await this.props.queue.declareWorker(
+        provisionerId, workerType, workerGroup, workerId, { disabled: disable }
+      );
+
+      this.setState({ worker });
+    } catch (error) {
+      this.setState({ error });
```
eliperelman: If there is an error, does the `worker` need to be set to `null`?

src/views/Worker/WorkerTable.js
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      filterStatus: 'all'
+    };
+  }
+  renderTaskDescription = description => (
```
eliperelman: Insert blank line between methods.

src/views/Worker/WorkerTable.js
```diff
@@ -0,0 +1,154 @@
...
+          <a target="_blank" rel="noopener noreferrer" href={task.metadata.source}>
+            {task.metadata.name}
+          </a>
+          &nbsp;&nbsp;
+          <OverlayTrigger trigger="click" rootClose placement="bottom" overlay={description}>
+            <Icon role="button" name="info-circle" />
+          </OverlayTrigger>
+        </td>
+        <td>
+          <a target="_blank" rel="noopener noreferrer" href={`/groups/${status.taskGroupId}/tasks/${status.taskId}`}>
```
eliperelman: Why does this open in a new window?

src/views/Worker/WorkerManager.js
```diff
@@ -0,0 +1,193 @@
...
+
+        this.setState({ worker, recentTasks, loading: false, error: null });
+      } catch (err) {
+        this.setState({ worker: null, recentTasks: [], loading: false, error: err });
+      }
+    });
+  };
+
+  loadRecentTasks = taskIds => (
+    Promise.all(taskIds.map(async (taskId) => {
```
eliperelman: This promise isn't returned so `await`ing this method isn't actually waiting for anything.

src/views/Worker/WorkerManager.js
```diff
@@ -0,0 +1,193 @@
...
+      const task = this.props.queue.task(taskId);
+      const status = this.props.queue.status(taskId);
+
+      return { task: await task, status: (await status).status };
```
eliperelman: Move the `await`s to the declarations themselves. You can also put these in a `Promise.all` so they can fire in parallel and not wait serially:

```js
const [task, { status }] = await Promise.all([
  this.props.queue.task(taskId),
  this.props.queue.status(taskId)
]);

return { task, status };
```

src/views/Worker/WorkerManager.js
```diff
@@ -0,0 +1,193 @@
...
+    }))
+  );
+
+  updateURI = (provisionerId, workerType, workerGroup, workerId, push) => {
+    const URL = [provisionerId, workerType, workerGroup, workerId]
```
eliperelman: Make this a lowercase `url`.

src/views/Worker/WorkerManager.js
helfi92: We still want to show the worker if the user attempts to disable a worker but is unsuccessful (perhaps not logged in). In that case, an error banner on top is good enough IMO.

src/views/Worker/WorkerTable.js
helfi92: When investigating a worker with possibly many tasks in it, the user will try to open all the tasks to investigate which one is acting abnormally. Taking that into consideration, I thought opening a new tab would be better for this scenario. But then again, the user can always do CMD + click to open it on a new tab. Since we're not navigating the user to a new host, perhaps the convention is not to open a new tab? @eliperelman What do you think?

src/views/Worker/WorkerManager.js
helfi92: I think JavaScript interprets everything after the open parenthesis on line 102 as a return value since we're not using open braces.

src/views/Worker/WorkerManager.js
helfi92: Good idea. We'll end up having 2 `Promise.all`:
```js
  loadRecentTasks = taskIds => (
    Promise.all(taskIds.map(async (taskId) => {
      const [task, {status}] = await Promise.all([
        this.props.queue.task(taskId),
        this.props.queue.status(taskId)
      ]);

      return {task, status};
    }))
  );
```

src/views/Worker/WorkerTable.js
helfi92: We should probably not make a change to the default behavior. I'll change it to using `<Link>`:

```js
<Link to={`/groups/${status.taskGroupId}/tasks/${status.taskId}`}>{status.taskId}</Link>
```

src/views/Worker/WorkerManager.js
eliperelman: Oh, my mistake, I thought it was a curly brace on L102 instead of the parenthesis.

eliperelman: In that case, the parentheses aren't needed at all:

```js
loadRecentTasks = taskIds => Promise.all(taskIds.map(async (taskId) => {
  // ...
});
```

src/views/Worker/WorkerManager.js
```diff
@@ -0,0 +1,195 @@
...
+  );
+
+  updateURI = (provisionerId, workerType, workerGroup, workerId, push) => {
+    const url = [provisionerId, workerType, workerGroup, workerId]
+      .reduce((uri, param) => uri.concat(param ? `/${param}` : ''), '/workers');
+
+    this.props.history[push ? 'push' : 'replace'](url);
+  };
+
+  disableWorker = async (disable) => {
```
eliperelman: It looks like `disabled` is a property of the worker, so I'm not sure we need this to be passed in since we are already access the worker on state:

```js
const { provisionerId, workerType, workerGroup, workerId, disabled } = this.state.worker;

try {
  const worker = await this.props.queue.declareWorker(
    provisionerId, workerType, workerGroup, workerId, { disabled }
  );

  // ...
}
```

That would allow you to change the usage in the render to pass a reference for this method directly:

```jsx
<Button
  onClick={this.disableWorker}
```

src/views/Worker/WorkerManager.js
helfi92: Removed and pushed the unneeded parenthesis :)

src/views/Worker/WorkerManager.js
helfi92: @eliperelman There's a reason why I did it like that. `disableWorker` can both disable and enable workers. If I didn't make it a parameter, the signature name would mislead people. That's why I added a boolean parameter to mean "disable" when the parameter is `true` and "enable" when `false`.  Thoughts?

eliperelman: Ah, then I would rename it to `toggleWorkerStatus` or something like that.

src/App/index.js
```diff
@@ -19,6 +19,7 @@ const UnifiedInspector = loadable(() => import(/* webpackChunkName: 'UnifiedInsp
 const QuickStart = loadable(() => import(/* webpackChunkName: 'QuickStart' */ '../views/QuickStart'));
 const AwsProvisioner = loadable(() => import(/* webpackChunkName: 'AwsProvisioner' */ '../views/AwsProvisioner'));
 const WorkerTypes = loadable(() => import(/* webpackChunkName: 'WorkerTypes' */ '../views/WorkerTypes'));
+const Worker = loadable(() => import(/* webpackChunkName: 'Worker' */ '../views/Worker'));
```
eliperelman: I'm going to be evil here, let's rename this to `WorkerManager`, along with the directory. I just want to avoid having clashes with platform APIs, e.g. `window.Worker`.

helfi92: You're always one step ahead in the game. Good thinking sir!

helfi92:
🚀

PR #298, opened 2017-09-12
Bug 1395358 (part 1) - use taskcluster-client-web

djmitche:
Pretty simple, but with a little extra work to use authorizedScopes for the action.

### DIFF
--- a/src/components/Clients/index.js
+++ b/src/components/Clients/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { func } from 'prop-types';
 import { camelCase } from 'change-case';
-import taskcluster from 'taskcluster-client';
+import * as taskcluster from 'taskcluster-client-web';
 
 export default class Clients extends React.PureComponent {
   static propTypes = {
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,4 @@
-import * as taskcluster from 'taskcluster-client-web';
+import { fromNowJSON } from 'taskcluster-client-web';
 import { omit } from 'ramda';
 import merge from 'deepmerge';
 import cloneDeep from 'lodash.clonedeep';
@@ -80,9 +80,9 @@ export const parameterizeTask = task => merge(omit([
   'payload'
 ], cloneDeep(task)), {
   retries: 0,
-  deadline: taskcluster.fromNowJSON('12 hours'),
-  created: taskcluster.fromNowJSON(),
-  expires: taskcluster.fromNowJSON('7 days'),
+  deadline: fromNowJSON('12 hours'),
+  created: fromNowJSON(),
+  expires: fromNowJSON('7 days'),
   scopes: task.scopes.filter(scope => !/^docker-worker:cache:/.test(scope)), // Delete cache scopes
   payload: merge(omit(['artifacts', 'cache'], task.payload || {}), {
     maxRunTime: Math.max(task.payload && task.payload.maxRunTime, 3 * 60 * 60),
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,4 @@
-import taskcluster from 'taskcluster-client';
+import * as taskcluster from 'taskcluster-client-web';
 import { omit } from 'ramda';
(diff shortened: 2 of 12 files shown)


src/components/Clients/index.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { func } from 'prop-types';
 import { camelCase } from 'change-case';
-import taskcluster from 'taskcluster-client';
+import * as taskcluster from 'taskcluster-client-web';
```
helfi92: `import taskcluster from 'taskcluster-client-web';`

eliperelman: That won't work since there isn't a default export defined by the module.

src/utils.js
```diff
@@ -1,4 +1,4 @@
-import taskcluster from 'taskcluster-client';
+import * as taskcluster from 'taskcluster-client-web';
```
eliperelman: Looks like this is a remnant from a previous version of this file, so I think we can switch to:

```js
import { fromNowJSON } from 'taskcluster-client-web';
```

src/views/UnifiedInspector/ActionsMenu.js
```diff
@@ -60,18 +59,6 @@ export default class ActionsMenu extends React.PureComponent {
       });
     }
 
-    if (nextProps.decision !== this.props.decision || nextProps.credentials !== this.props.credentials) {
```
eliperelman: Any reason why we don't want to update when the decision changes?

src/views/UnifiedInspector/ActionsMenu.js
```diff
@@ -369,7 +356,15 @@ export default class ActionsMenu extends React.PureComponent {
       input
     }, actions.variables));
 
-    await this.actionsQueue.createTask(newTaskId, newTask);
+    if (!this.props.decision) {
+      throw new Error('no action task found');  // .. how did we find an action, then?
```
eliperelman: Was something wrong here before that no longer works? cc: @bstack

src/views/UnifiedInspector/ActionsMenu.js
```diff
@@ -369,7 +356,15 @@ export default class ActionsMenu extends React.PureComponent {
...
+    }
+    // call the queue with the decision task's scopes, as directed by the action spec
```
eliperelman: Insert blank line between declarations and other statements (above the comment).

src/views/UnifiedInspector/ActionsMenu.js
```diff
@@ -369,7 +356,15 @@ export default class ActionsMenu extends React.PureComponent {
...
+    const actionsQueue = this.props.queue.use({
+      authorizedScopes: this.props.decision.scopes || []
+    });
+    await actionsQueue.createTask(newTaskId, newTask);
```
eliperelman: Insert blank line between statements and declarations.

src/components/Clients/index.js
eliperelman: It would be nice to not need to import everything here, but switching to dynamic imports would make this process async, and I'm not sure we want to go down that route yet.

eliperelman: It would be nice to not need to import everything here, but switching to dynamic imports would make this process async, and I'm not sure we want to go down that route yet.

src/views/UnifiedInspector/ActionsMenu.js
djmitche: There's nothing to update -- the removed code was updating a property on `this`, but that property is no longer used.

src/views/UnifiedInspector/ActionsMenu.js
djmitche: Sort of -- `this.actionQueue` was only updated when `props.decision` was truthy.  So if `props.decision` was updated from truthy to falsey, `this.actionQueue` would still hold its own value (corresponding to the old decision task).

djmitche:
Updated!

eliperelman:
w00t

PR #299, opened 2017-09-13
Revise error handling from changes in tc-client-web

eliperelman:


djmitche:
What was the change?

eliperelman:
@djmitche taskcluster-client-web outputs the results from requests differently than taskcluster-client does. Since we rely on fetch and not request, I changed the "schema" of the error response.

taskcluster-client:

https://github.com/taskcluster/taskcluster-client/blob/fffebc153b5f17dc161127917eaf4c42b1119be3/lib/client.js#L332-L335

taskcluster-client-web:

https://github.com/taskcluster/taskcluster-client-web/blob/5005c79bb7283e1c4a1a9da953d01f5eb3a13435/src/fetch.js#L23-L28

djmitche:
Hm, I dislike putting a match on `----` in the library.  I didn't see that in review :)

https://docs.taskcluster.net/manual/design/apis/errors is the description of how our errors work.  Ideally tools would show body.code in the error message title, (the full) body.message in the body, and body.requestInfo in a "details" link if necessary.  In particular, this PR seems to drop use of body.code, and especially for auth errors that's a really helpful piece of information (it distinguishes bad credentials from bad scopes).

eliperelman:
@djmitche regarding `---`, that seems to be a legacy thing, not really a fan myself.

I'll revise the PR, but note that the "Additional details" still stringifies the entire error.body, which would include all the information that you mention is missing.

djmitche:
Yeah, I think that "Additional Details" could omit the message and just show `JSON.stringify(body.requestInfo)`

djmitche:
BTW, this is related to https://bugzilla.mozilla.org/show_bug.cgi?id=1282508

eliperelman:
Grr, I thought I pushed this yesterday...

src/components/Error/index.js
```diff
@@ -42,28 +37,20 @@ export default class Error extends React.PureComponent {
...
-
-            const index = error.message.indexOf('----');
-
-            return index === -1 ? error.message : error.message.slice(0, index);
-          })()}
+          {error.message}
         </Markdown>
         {process.env.NODE_ENV === 'development' && <pre>{error.stack}</pre>}
-        {error.body && (
+        {error.body && error.body.requestInfo (
```
djmitche: missing `&&`

eliperelman: wow fail

PR #302, opened 2017-09-14
Bug 1399566 - Worker Types Keep settings when refreshed

helfi92:
Persist settings when you refresh or switch provisioners. Note that I don't persist the `orderBy` value when one switches provisioners since that list is dynamically generated.

### DIFF
--- a/src/views/WorkerTypes/WorkerManager.js
+++ b/src/views/WorkerTypes/WorkerManager.js
@@ -20,15 +20,13 @@ export default class WorkerManager extends React.PureComponent {
       lastActive: true,
       layout: 'grid',
       orderBy: null,
-      error: null
+      error: null,
+      ...this.getSettingsFromProps(props)
     };
   }
 
   componentWillMount() {
-    const settings = this.getSettingsFromProps(this.props);
-
     this.loadProvisioners();
-    this.setState(settings);
   }
 
   async loadProvisioners(token) {
@@ -69,7 +67,7 @@ export default class WorkerManager extends React.PureComponent {
   getSettingsFromProps = (props) => {
     const settings = parse(props.location.search.slice(1));
 
-    if (Object.prototype.hasOwnProperty.call(settings, 'lastActive')) {
+    if ('lastActive' in settings) {
       settings.lastActive = settings.lastActive !== 'false' && settings.lastActive !== 'true' ?
         true :
         JSON.parse(settings.lastActive);
@@ -82,12 +80,14 @@ export default class WorkerManager extends React.PureComponent {
     const query = { ...q };
     const oldQuery = parse(this.props.location.search.slice(1));
 
-    Object.entries(query).forEach(([key, value]) => {
-      if (typeof value !== 'boolean' && !value) {
-        delete oldQuery[key];
-        delete query[key];
-      }
-    });
+    Object
+      .entries(query)
+      .forEach(([key, value]) => {
+        if (typeof value !== 'boolean' && !value) {
+          delete oldQuery[key];
+          delete query[key];
+        }
+      });
 
     return stringify({ ...oldQuery, ...query });
   };
--- a/src/views/WorkerTypes/WorkerManager.js
+++ b/src/views/WorkerTypes/WorkerManager.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Button, ButtonToolbar, DropdownButton, MenuItem, ToggleButtonGroup, ToggleButton, Tooltip, OverlayTrigger } from 'react-bootstrap';
 import Icon from 'react-fontawesome';
+import { parse, stringify } from 'qs';
 import Error from '../../components/Error';
 import HelmetTitle from '../../components/HelmetTitle';
 import SearchForm from './SearchForm';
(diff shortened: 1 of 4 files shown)


src/views/WorkerTypes/WorkerManager.js
```diff
@@ -50,19 +54,48 @@ export default class WorkerManager extends React.PureComponent {
...
+  handleLayoutChange = () => this.setQuery({ layout: this.state.layout === 'grid' ? 'table' : 'grid' });
+
+  handleLastActiveClick = () => this.setQuery({ lastActive: !this.state.lastActive, orderBy: null });
+
+  handleOrderBySelect = orderBy => this.setQuery({ orderBy, lastActive: false });
+
+  getSettingsFromProps = (props) => {
+    const settings = parse(props.location.search.slice(1));
+
+    if (Object.prototype.hasOwnProperty.call(settings, 'lastActive') && settings.lastActive !== 'false' && settings.lastActive !== 'true') {
```
eliperelman: This can probably be shortened to just:

```js
if (settings.lastActive !== 'false' && settings.lastActive !== 'true') {
```

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -50,19 +54,48 @@ export default class WorkerManager extends React.PureComponent {
...
+
+    return settings;
+  };
+
+  constructQuery = (q) => {
+    const query = { ...q };
+    const oldQuery = parse(this.props.location.search.slice(1));
 
-  handleLastActiveClick = () => this.setState({ lastActive: !this.state.lastActive, orderBy: null });
+    Object.entries({ ...query }).forEach(([key, value]) => {
```
eliperelman: You already made a copy of `q` into `query`, do you need to copy it again here?

helfi92: Forgot to remove it 😅 Thanks for catching that duplication

src/views/WorkerTypes/WorkerManager.js
helfi92: Hm, I'm not sure. If `settings` happens to be for example `{ orderBy: 'pendingCapacity' }`, you don't want the `if` statement to return `true` since you can't have `lastActive` and `pendingCapacity` both set at the same time. To give it some more elegance, I rewrote the logic to be:

```js
if (Object.prototype.hasOwnProperty.call(settings, 'lastActive')) {
  settings.lastActive = settings.lastActive !== 'false' && settings.lastActive !== 'true' ?
    true :
    JSON.parse(settings.lastActive);
}
```

helfi92:
I added extra code to have the input field include the default search term when you refresh on a filtered list.

src/views/WorkerTypes/WorkerManager.js
eliperelman: Ah ok, you can shorten this then with the `in` operator:

```js
if ('lastActive' in settings) {
```

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -13,18 +14,21 @@ export default class WorkerManager extends React.PureComponent {
...
       lastActive: true,
-      gridLayout: true,
+      layout: 'grid',
       orderBy: null,
       error: null
     };
   }
 
   componentWillMount() {
+    const settings = this.getSettingsFromProps(this.props);
```
eliperelman: Instead of a `setState` in `componentWillMount`, this state in the constructor should be derived from these settings:

```js
constructor(props) {
  this.state = {
    // other state
    ...this.getSettingsFromProps(props)
  };
}

componentWillMount() {
  this.loadProvisioners();
}
```

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -50,19 +54,50 @@ export default class WorkerManager extends React.PureComponent {
...
+  handleLayoutChange = () => this.setQuery({ layout: this.state.layout === 'grid' ? 'table' : 'grid' });
+
+  handleLastActiveClick = () => this.setQuery({ lastActive: !this.state.lastActive, orderBy: null });
+
+  handleOrderBySelect = orderBy => this.setQuery({ orderBy, lastActive: false });
+
+  getSettingsFromProps = (props) => {
+    const settings = parse(props.location.search.slice(1));
+
+    if (Object.prototype.hasOwnProperty.call(settings, 'lastActive')) {
```
eliperelman:
```js
if ('lastActive' in settings) {
  settings.lastActive = (settings.lastActive !== 'false' && settings.lastActive !== 'true') ||
    JSON.parse(settings.lastActive);
}
```

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -50,19 +54,50 @@ export default class WorkerManager extends React.PureComponent {
...
+
+    return settings;
+  };
+
+  constructQuery = (q) => {
+    const query = { ...q };
+    const oldQuery = parse(this.props.location.search.slice(1));
 
-  handleLastActiveClick = () => this.setState({ lastActive: !this.state.lastActive, orderBy: null });
+    Object.entries(query).forEach(([key, value]) => {
```
eliperelman: Let's split the chains onto multiple lines:

```js
Object
  .entries(query)
  .forEach(...
```

eliperelman:
:tada:

PR #309, opened 2017-09-19
Use sourcemap for development builds

djmitche:
This is an elaborate way of asking, "why `eval`"?

### DIFF
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -129,7 +129,7 @@ module.exports = {
   ],
   env: {
     NODE_ENV: {
-      development: ({ config }) => config.devtool('eval')
+      development: ({ config }) => config.devtool('sourcemap')
     }
   }
 };


helfi92:
`eval` => fast builds
`sourcemap` => slow builds

If we want source-maps for development, we should go with `eval-source-map`. Build speed is faster than  `sourcemap` but slower than `eval`. It's a trade off.

eliperelman:
What @helfi92 says is correct, its a tradeoff between build speed and accuracy of line number source mapping during development.

https://webpack.js.org/configuration/devtool/#devtool

This just gave me another thought, we shouldn't stop generating sourcemaps in Travis so we get faster builds there.

.neutrinorc.js
```diff
@@ -129,7 +129,7 @@ module.exports = {
   ],
   env: {
     NODE_ENV: {
-      development: ({ config }) => config.devtool('eval')
+      development: ({ config }) => config.devtool('sourcemap')
```
eliperelman: I'm not sure `sourcemap` is a valid value here, I think it's `source-map`. Regardless, we probably don't actually want this in development, but rather a different option from:

https://webpack.js.org/configuration/devtool/#development

`eval-source-map` is probably a good option.

djmitche:
Hm, 'sourcemap' is working fine for me.  And it's not any slower than eval
was.

eliperelman:
Resolved in #342.

PR #312, opened 2017-09-21
Delete pending label

helfi92:
A task will never be in this list for a particular worker in a state of pending. Tasks are only associated with workers once a run has been claimed.


djmitche:
It would certainly be a bug, but we'd like to not hide such tasks if it occurs.  From my read we'd still see pending tasks when no filter is applied, so this is fine.

helfi92:
@djmitche If it ever happens, pending tasks will still appear in the list. I am just removing `pending` from the "filter by" dropdown

PR #313, opened 2017-09-23
Bug 1401040 - workers page sitting in between

helfi92:


helfi92:
screen shot 2017-09-23 at 4 03 06 pm

helfi92:
Made navigation buttons less confusing by joining them together and making use of the `disabled` property:

screen shot 2017-09-24 at 9 41 47 pm

src/views/Workers/WorkerManager.js
```diff
@@ -0,0 +1,141 @@
...
+      error: null,
+      filter: 'None'
+    };
+  }
+
+  componentWillMount() {
+    this.loadWorkers(this.props);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
```
eliperelman: Why `componentDidUpdate` over `componentWillUpdate`?

src/views/Workers/WorkerManager.js
```diff
@@ -0,0 +1,141 @@
...
+
+    if (prevState.filter !== this.state.filter) {
+      this.setState({ workers: null, loading: true }, () => this.loadWorkers(this.props));
+    }
+  }
+
+  async loadWorkers({ provisionerId, workerType }) {
+    try {
+      const workers = await this.props.queue.listWorkers(provisionerId, workerType, {
+        ...this.state.workerToken ? { continuationToken: this.state.workerToken } : {},
```
eliperelman: To avoid ambiguity, put parentheses around a ternary you are going to spread:

```js
...(this.state.workerToken ? { continuationToken: this.state.workerToken } : {})
```

src/views/Workers/WorkerManager.js
helfi92: The react docs [#componentWillUpdate](https://facebook.github.io/react/docs/react-component.html#componentwillupdate) makes mention not to use `this.setState` in `componentWillUpdate`.

src/views/Workers/WorkerManager.js
helfi92: Agreed 💯.

src/views/Workers/WorkerManager.js
eliperelman: Ah you're right, I read this as `componentDidMount`. Proceed 😆

src/views/Workers/WorkerManager.js
```diff
@@ -0,0 +1,141 @@
...
+  componentWillMount() {
+    this.loadWorkers(this.props);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.workerToken !== this.state.workerToken) {
+      this.setState({ loading: true }, () => this.loadWorkers(this.props));
+    }
+
+    if (prevState.filter !== this.state.filter) {
```
eliperelman: Is it possible for both `workerToken` and `filter` to change? If so, then this would potentially cause 2-4 `setState`s.

src/views/Workers/WorkerManager.js
```diff
@@ -0,0 +1,141 @@
...
+            {!loading && workers && (
+              workers.workers.map(({ workerGroup, workerId, latestTask, firstClaim, disabled }, index) => (
+                <tr key={`worker-${index}`}>
+                  <td>
+                    <Link
+                      to={`/workers/provisioners/${this.props.provisionerId}/worker-types/${this.props.workerType}/workers/${workerGroup}/${workerId}`}>
+                      {workerId}
+                    </Link>
+                  </td>
+                  <td>{latestTask ? <Link to={`/tasks/${latestTask}`}>{latestTask}</Link> : '-'}
```
eliperelman: Either move the closing `</td>` to the same line, or move the tag content to its own line.

src/views/Workers/WorkerManager.js
helfi92: You're absolutely right @eliperelman. Good catch!

helfi92: @eliperelman Not so pretty but it's now:

```js
componentDidUpdate(prevProps, prevState) {
  const newState = {};

  if (prevState.workerToken !== this.state.workerToken) {
    newState.loading = true;
  }

  if (prevState.filter !== this.state.filter) {
    newState.loading = true;
    newState.workers = null;
  }

  if (Object.keys(newState).length) {
    this.setState(newState, () => this.loadWorkers(this.props));
  }
}
```

eliperelman: Try this:

```js
componentDidUpdate(prevProps, { workerToken, filter }) {
  const filterChanged = filter !== this.state.filter

  if (workerToken !== this.state.workerToken || filterChanged) {
    this.setState({
      loading: true,
      ...(filterChanged ? { workers: null } : {})
    }, () => this.loadWorkers(this.props));
  }
}
```

helfi92: Much better!

helfi92:
@eliperelman I think this is ready now. OK to 🚢 ?

eliperelman:
:shipit:

PR #321, opened 2017-09-27
Reload group and tasks when listeners reconnect

eliperelman:


eliperelman:
@jonasfj does that make more sense?

jonasfj:
r+

eliperelman:
🎉

PR #323, opened 2017-09-27
Reduce re-renders in inspector, improve travis build times by removing sourcemap

eliperelman:
I'm trying to improve the performance of the inspector by:

* Loading minimal records up-front, then downloading in batches of 200
* Reduce requests made in series in favor of parallel
* Improve rendering by storing up a couple `setState`s into a single call
* Don't `await` tasks continuation, just continue during animation frame

Also this should hopefully improve the time of travis builds by shutting off source maps in CI.

### DIFF
--- a/src/views/UnifiedInspector/Inspector.js
+++ b/src/views/UnifiedInspector/Inspector.js
@@ -123,7 +123,7 @@ export default class Inspector extends React.PureComponent {
     }
   }
 
-  async loadActions(props) {
+  async getActions(props) {
     const { queue, taskGroupId } = props;
 
     if (!taskGroupId) {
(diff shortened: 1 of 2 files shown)


src/views/UnifiedInspector/Inspector.js
```diff
@@ -179,7 +185,7 @@ export default class Inspector extends React.PureComponent {
       });
 
       if (continuationToken) {
-        await this.loadTasks(props, continuationToken);
+        requestAnimationFrame(() => this.loadTasks(props, continuationToken));
```
djmitche: For my edification, what's the difference here?  In the old mode, the `loadTasks` promise won't resolve until all tasks are read.  In the new, it will resolve after the first batch is read, with a new Promise created for each batch.  Nothing's waiting for any of those Promises anyway..

helfi92: Why do we need to use `requestAnimationFrame` in this context? From my understanding, when `requestAnimationFrame` triggers its callback when it's the right time to paint/render the screen. However, when that moment happen, we still need to wait over an asynchronous call. The right time to paint won't be the right time after all. Perhaps what we want here would be something like `requestIdleCallback`, where this method queues a function to be called during a browser's idle periods.

eliperelman: You're right, nothing is listening to these, and I'm just trying to make that clearer by not making the method await a new stack into oblivion and trying to batch setStates during animation frames.

eliperelman: @helfi92 I think you're right here as well, but `requestIdleCallback` support isn't very good right now, hence the `rAF`.

If anything, we should probably just remove the `rAF`.

PR #326, opened 2017-09-28
Add Task state

helfi92:
Now that we already make a call to the /status endpoint of a task, it would be useful to show the state of the most recent task.


helfi92:
screen shot 2017-09-27 at 9 14 41 pm

PR #327, opened 2017-09-28
Bug 1403901 Persist filter when switching tabs

helfi92:


src/views/UnifiedInspector/Inspector.js
```diff
@@ -487,6 +488,8 @@ export default class Inspector extends React.PureComponent {
     });
   }
 
+  onFilterChange = filterStatus => this.setState({ filterStatus });
```
eliperelman: Rename this to `handleFilterChange`.

PR #332, opened 2017-09-28
Bug 1403941 - [dashboard] 404 error

helfi92:
Previously, we use to change the URL whenever the provisioner and workerType were modified. This was working just fine since we didn't share the URL with different pages. Now that we share it with the worker-types and workers page, modifying the URL was causing issues. This PR fixes the issue.

### DIFF
--- a/src/views/WorkerManager/SearchForm.js
+++ b/src/views/WorkerManager/SearchForm.js
@@ -26,38 +26,72 @@ export default class SearchForm extends PureComponent {
   constructor(props) {
     super(props);
     this.state = {
+      provisionerId: this.props.provisionerId,
+      workerType: this.props.workerType,
       workerGroupInput: this.props.workerGroup,
       workerIdInput: this.props.workerId
     };
   }
 
-  onSubmit = e => {
-    const { provisionerId, workerType } = this.props;
+  componentWillUpdate(nextProps, { provisionerId }) {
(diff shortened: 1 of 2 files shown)


src/views/WorkerManager/SearchForm.js
```diff
@@ -26,38 +26,72 @@ export default class SearchForm extends PureComponent {
   constructor(props) {
     super(props);
     this.state = {
+      provisionerId: this.props.provisionerId,
```
eliperelman: I'm always find setting state equal to props to be a code smell. Since these are initial values for an input, let's change their name to have `Input` to reduce confusion.

helfi92: @eliperelman Agreed 💯. Btw, do you mean something like that:

```js
this.state = {
  provisionerId: this.props.provisionerIdInput,
  workerType: this.props.workerTypeInput,
  workerGroupInput: this.props.workerGroupInput,
  workerIdInput: this.props.workerIdInput
};

```

eliperelman: No, the state entries should have input names, not the props.

eliperelman: The parent container ideally should not have knowledge of the state of the child.

helfi92: You're right. OK, will make that change.

PR #334, opened 2017-09-29
Bug 1400550 - handle secrets better

djmitche:


djmitche:
I think that never happens, but if it does, it just gets shown using the fallback error display.

PR #338, opened 2017-10-01
Bug 1313743 - Task inspector does not show task expiration

helfi92:


djmitche:
Haha, thanks!  I was sort of saving this for potential contributors :)

PR #339, opened 2017-10-02
Bug 1403987 - Link to worker page from the task inspector

helfi92:
This will need to land **after** the dashboard is made public

### DIFF
--- a/src/views/UnifiedInspector/RunDetails.js
+++ b/src/views/UnifiedInspector/RunDetails.js
@@ -79,10 +79,14 @@ export default class RunDetails extends React.PureComponent {
           <tr>
             <td>WorkerId</td>
             <td>
-              <Link
-                to={`/workers/provisioners/${task.provisionerId}/worker-types/${task.workerType}/workers/${run.workerGroup}/${run.workerId}`}>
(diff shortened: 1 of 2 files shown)


src/views/UnifiedInspector/RunDetails.js
```diff
@@ -76,7 +78,12 @@ export default class RunDetails extends React.PureComponent {
 
           <tr>
             <td>WorkerId</td>
-            <td>{run.workerId ? <code>{run.workerId}</code> : '-'}</td>
+            <td>
+              <Link
+                to={`/workers/provisioners/${task.provisionerId}/worker-types/${task.workerType}/workers/${run.workerGroup}/${run.workerId}`}>
+                {run.workerId ? run.workerId : '-'}
```
djmitche: This should probably be around the whole Link, right?  Otherwise we'll have "-" with a link that will show an error.

eliperelman: Correct, the ternary should either render the `Link` or the `-`, not a dash in the link.

helfi92: My mistake. Fixing that now.

PR #340, opened 2017-10-02
Bug 1405082 - Page listing provisioners with their descriptions

helfi92:


helfi92:
screen shot 2017-10-02 at 2 56 09 pm

helfi92:
@djmitche Is there anything else you want to display on that page or is that good enough? :)

djmitche:
That's great, but it should probably have some way of linking over to the workerType explorer with that provisioner pre-selected.  Maybe just a button:
```
[ -> WorkerTypes for this Provisioner ]
```

src/App/index.js
```diff
@@ -39,6 +39,9 @@ const WorkerManager = loadable(() =>
 const Workers = loadable(() =>
   import(/* webpackChunkName: 'Workers' */ '../views/Workers')
 );
+const ProvisionersDocs = loadable(() =>
```
djmitche: This should just be "Provisioners", I think.  It's no more "docs" than the hooks page is "docs".

src/App/index.js
```diff
@@ -215,6 +218,10 @@ export default class App extends React.Component {
                   routeRoot="/aws-provisioner-staging"
                 />
                 <PropsRoute
+                  path="/workers/provisioners/docs/:provisionerId?"
```
djmitche: `/workers/provisioners/:provisionerId?`

helfi92: @djmitche It would collide with:
https://github.com/taskcluster/taskcluster-tools/blob/master/src/App/index.js#L227-L230

djmitche: Hm, maybe that's telling us something?  I see us having:

 * /workers/provisioners -- list of provisioners
 * /workers/provisioners/:provisionerId -- information about a provisioner
 * /workers/provisioners/:provisionerId/worker-types -- list of workerTypes
 * /workers/provisioners/:provisionerId/worker-types/:workerType -- information about a workerType
 * /workers/provisioners/:provisionerId/worker-types/:workerType/workers -- list of workers
 * /workers/provisioners/:provisionerId/worker-types/:workerType/workers/:workerGroup?/:workerId? -- information about a worker

So it all sort of flows together, with links forward and breadcrumbs backward.

src/views/ProvisionersDocs/index.js
```diff
@@ -0,0 +1,18 @@
...
+import Clients from '../../components/Clients';
+import ProvisionersDocs from './ProvisionersDocs';
+
+const View = ({ match }) => (
+  <Clients Queue>
+    {clients => (
+      <ProvisionersDocs
+        {...clients}
+        provisionerId={
+          match.params.provisionerId ? match.params.provisionerId : ''
```
eliperelman:
```js
match.params.provisionerId || ''
```

src/views/ProvisionersDocs/ProvisionersDocs.js
```diff
@@ -0,0 +1,154 @@
...
+            <ListGroup>
+              {this.state.provisioners.map((provisioner, key) => (
+                <LinkContainer
+                  activeStyle={{
+                    color: '#555',
+                    backgroundColor: '#f5f5f5',
+                    border: '1px solid #ddd'
+                  }}
+                  isActive={() =>
+                    this.props.provisionerId === provisioner.provisionerId}
```
eliperelman: Is this supposed to be a function?

src/App/index.js
eliperelman: While I agree with your routing logic, I can't help but think this top level `workers` route is also a smell. There is nothing from a collections standpoint that links workers with provisions using this. For example speaking logically, would there be a difference in how these 2 routes render their collections?

* `/workers/provisioners`
* `/provisioners`

src/App/index.js
eliperelman: Agreed.

djmitche:
I see what you mean about the /workers root path.  Starting with /provisioners makes sense -- then the worker explorer is just a drilled-down view "inside" the provisioners view.  I think that's an accurate reflection of the data model.

src/views/ProvisionersDocs/ProvisionersDocs.js
helfi92: Not needed anymore. Forgot to remove it :)

helfi92:
UI changes:

* A link to the worker-types page was added (link with arrow):
screen shot 2017-10-03 at 10 04 41 am

* Modified the breadcrumb. On the worker page, it looks like the following:
screen shot 2017-10-03 at 10 06 35 am

src/views/Provisioners/Provisioners.js
```diff
@@ -0,0 +1,160 @@
...
+      }
+    } catch (error) {
+      this.setState({
+        provisioners: null,
+        error
+      });
+    }
+  }
+
+  onProvisionerClick = ({ target }) => this.loadProvisioner(target.innerText);
```
eliperelman: Rename this to `handleProvisionerClick`.

src/views/Provisioners/Provisioners.js
```diff
@@ -0,0 +1,160 @@
...
+      <div>
+        <div>
+          <HelmetTitle title="Worker Types Explorer" />
+          <h4>Provisioners</h4>
+        </div>
+        {this.state.error && <Error error={this.state.error} />}
+        <Row>
+          <Col md={6}>
+            <ListGroup>
+              {this.state.provisioners.map((provisioner, key) => (
```
eliperelman: Looks like the only thing we use from `provisioner` is the ID, so you can destructure this to `{ provisionerId }`.

src/views/Provisioners/Provisioners.js
```diff
@@ -0,0 +1,160 @@
...
+        provisioners: null,
+        error
+      });
+    }
+  }
+
+  onProvisionerClick = ({ target }) => this.loadProvisioner(target.innerText);
+
+  render() {
+    return (
```
eliperelman: Since we are access `state` a lot in this method, let's declare a few things for simplicity:

```js
const { provisioners, provisioner, error } = this.state;
```

src/views/Provisioners/index.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import Clients from '../../components/Clients';
+import Provisioners from './Provisioners';
+
+const View = ({ match }) => (
+  <Clients Queue>
+    {clients => (
```
eliperelman: We only have a single client, so let's destructure this to just pass `queue`:

```jsx
{({ queue }) => (
  <Provisioners
    queue={queue}
```

src/views/Provisioners/styles.css
```diff
@@ -0,0 +1,9 @@
+.dataContainer {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.dataContainer > div {
+  margin: 0 10px 10px 0;
+}
+
```
eliperelman: Remove extra line.

src/views/WorkerManager/WorkerManager.js
```diff
@@ -184,13 +184,16 @@ export default class WorkerManager extends React.PureComponent {
           <h4>Worker Explorer</h4>
         </div>
         <Breadcrumb>
-          <Breadcrumb.Item
-            href={`/workers/provisioners/${this.props.provisionerId}`}>
+          <Breadcrumb.Item href={`/provisioners/${this.props.provisionerId}`}>
```
eliperelman: Declare `const { provisionerId } = this.props;` before this so we can shorten this and not have to access `props` constantly.

src/views/WorkerManager/WorkerManager.js
```diff
@@ -184,13 +184,16 @@ export default class WorkerManager extends React.PureComponent {
...
             {this.props.provisionerId}
           </Breadcrumb.Item>
           <Breadcrumb.Item
-            href={`/workers/provisioners/${this.props
-              .provisionerId}/worker-types/${this.props.workerType}`}>
+            href={`/provisioners/${this.props.provisionerId}/worker-types`}>
+            worker-types
+          </Breadcrumb.Item>
+          <Breadcrumb.Item
+            href={`/provisioners/${this.props.provisionerId}/worker-types/${this
```
eliperelman: Also pull `workerType` into a declaration before this.

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -140,6 +140,16 @@ export default class WorkerManager extends React.PureComponent {
           <HelmetTitle title="Worker Types Explorer" />
           <h4>Worker Types Explorer</h4>
         </div>
+        <Breadcrumb>
+          <Breadcrumb.Item href={`/provisioners/${this.props.provisionerId}`}>
```
eliperelman: Just FYI, my general rule of thumb is if you access it more than 1-2 times, pull it out into a declaration.

helfi92: You mean something like:

```js
const Breadcrumb = (
  <Breadcrumb>
      // ...
  </Breadcrumb>
)

// then replace L143 with {Breadcrumb}
```

I was thinking we could create a BreadCrumb view component i.e. <Breadcrumb list=(array) active=(array or string) />, then we could use that component in all the other pages that use breadcrumbs. @eliperelman Thoughts?

eliperelman: That's possible, but my comment was in reference to `this.props.provisionerId`.

helfi92: haha, ok :)

djmitche:
I'd like this:

to eventually be
```
provisioners / aws-provisioner-v1 / worker-types / gecko-3-b-android / workers / ..
```
that is, to include the provisioners, worker-types, and workers lists in the breadcrumbs.  But that's good follow-up fodder, along with making sure there's suitable forward links between all of those views, too.

helfi92:
@djmitche , `provisioners / aws-provisioner-v1` points to the same page as `provisioners`, except in the former, a provisioner is selected (the second half of the page will show the selected provisioner). I don't think we want to have 2 links pointing to the same page, but instead links pointing to different pages.

djmitche:
OK, we can talk about the breadcrumbs separately :)

helfi92:
@eliperelman Went over all of your requests :) Anything else before we merge?

PR #343, opened 2017-10-03
Add breadcrumb component

helfi92:
We're using `Breadcrumb` in many of our views now so I think putting it under the components folder would be nice :) With this PR, you only need to do:

```jsx
// inside link2 view
const navList = [
  {
    title: 'titleA',
    href: '/link1'
  },
  {
    title: 'titleB',
    href: '/link2
  }
];

<Breadcrumb navList={navList} active="titleB" />

### DIFF
--- a/src/components/Breadcrumb/index.js
+++ b/src/components/Breadcrumb/index.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import { Breadcrumb as Nav } from 'react-bootstrap';
+import { arrayOf, string, object, oneOfType } from 'prop-types';
+
+export default class Breadcrumb extends React.PureComponent {
+  isActive = title =>
+    typeof this.props.active === 'string'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by eliperelman** [comment]
You'll need to rebase because of my v7 fixes patch.

File: src/components/Breadcrumb/index.js [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+            {entry.title}
+          </Nav.Item>
+        ))}
+      </Nav>
+    );
+  }
+}
+
+Breadcrumb.propTypes = {
+  active: oneOfType([string, arrayOf(string)]).isRequired,
```
eliperelman: Is it possible to default the last item in the `navList` array to be the active one if `active` isn't specified?

helfi92: 💯 Yep!

PR #354, opened 2017-10-06
Save GroupDetails re-renders when tasks update but count does not change

eliperelman:


eliperelman:
Going to use a different implementation based on web workers.

helfi92:
Would be interesting to see how web workers would affect performance in this case.

eliperelman:
The workers have their own tradeoff of transport performance which eliminates any of the gains of using workers. Let's move forward with this for now.

PR #356, opened 2017-10-06
Hotfix - tasks is null on group details

eliperelman:


helfi92:
@eliperelman Is that necessary given [GroupDetails.js#L34](https://github.com/eliperelman/taskcluster-tools/blob/ff2a581d941df15892c13bedc0408df954a6c43c/src/views/UnifiedInspector/GroupDetails.js#L34) where array is the only data type allowed as tasks props value?

eliperelman:
Yes, since it's not required so may be undefined.

PR #359, opened 2017-10-09
Bug 1344409 - Drop the timestamp warning in Task Creator after "Update Timestamps" is clicked

ydidwania:
All the error warnings are dropped when "Update Timestamps" is clicked. User needs to click "Create Task" again to check.


ydidwania:
Yeah, even I was thinking about the same.  
Made the changes.  :)

djmitche:
Thank you!

PR #361, opened 2017-10-10
Bug 1407038 - Task name in recent tasks display should link to the task, not the source

helfi92:


helfi92:
Removed link to source, added arrow icon to make it consistent with the group inspector and last, put the info icon in the beginning to make it look nicer.

screen shot 2017-10-10 at 6 00 55 pm

djmitche:
I like it.  Maybe the info could include the source link, so the repo is still visible on hover?

helfi92:
screen shot 2017-10-11 at 9 10 57 am

src/views/WorkerManager/WorkerTable.js
```diff
@@ -31,11 +31,18 @@ export default class WorkerTable extends React.PureComponent {
...
 
-  renderTaskDescription = description => (
+  renderTaskDescription = ({ description, source }) => (
     <Popover
       className={styles.taskPopover}
       id="popover-trigger-click-root-close"
-      title="Description">
+      title={
+        <div>
+          Description&nbsp;&nbsp;
```
djmitche: "Source" :)

djmitche: Oh, I see, sorry -- I thought those were two different divs with different pieces of information in them:
```
Source: ...
Description: ...
```
To support the use-case of seeing the project, I think we should show the actual link in the popover rather than making users click and then hover.

helfi92: Sounds good.

helfi92:
@djmitche 
screen shot 2017-10-11 at 9 33 20 am

PR #364, opened 2017-10-16
Bug 1395349 - Invent a way to support CLI logins

helfi92:


helfi92:
#### Workflow

Page first suggests that the user sign user in if not already signed in:

screen shot 2017-10-16 at 11 24 36 am

When signed in then calculate clientId base like client editor does (`${userSession.credentials.clientId}/${querystring.name}`).

If the resulting clientId exists, offer to reset accessToken on it:

screen shot 2017-10-16 at 1 57 47 pm

If they do not want to reset it, show the new-client form:

screen shot 2017-10-16 at 2 01 07 pm

if the resulting clientId doesn't exists, show a new-client form with the values read from the query string:

screen shot 2017-10-16 at 2 04 10 pm

Creating the client will perform a redirect to <callback_url>?clientId=..&accessToken=..

src/App/index.js
```diff
@@ -93,6 +93,9 @@ const DevelopmentLogin = loadable(() =>
 const ManualLogin = loadable(() =>
   import(/* webpackChunkName: 'ManualLogin' */ '../views/ManualLogin')
 );
+const AuthNew = loadable(() =>
```
eliperelman: Let's rename this to `NewAuthClient`.

src/components/CredentialsMenu/index.js
```diff
@@ -72,7 +78,31 @@ class CredentialsMenu extends React.PureComponent {
     );
   }
 
+  renderAsList = () => {
```
eliperelman: This doesn't need to be an arrow function.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+    const { name } = parse(location.search.slice(1));
+    const id = `${userSession.credentials.clientId}/${name}`;
+
+    try {
+      const previousClient = (await auth.listClients({
+        prefix: id
+      })).filter(({ clientId }) => clientId === id)[0];
+      const client = this.constructClient(previousClient);
+
+      if (!previousClient) {
```
eliperelman: I think this check should be performed prior to calling `this.constructClient`.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+        The accessToken is expiring{' '}
+        <strong>{<DateView date={this.state.client.expires} />}</strong>. What
+        do you want to do?
+      </h4>
+      <ButtonToolbar className={styles.flexCenter}>
+        <Button bsStyle="primary" onClick={this.handleCreateNewClient}>
+          Create a new clientId
+        </Button>
+        <Button bsStyle="warning" onClick={this.handleResetAccessToken}>
+          <Glyphicon glyph="fire" /> Reset accessToken
```
eliperelman: `fire` doesn't seem like an appropriate icon here. Let's try another one.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+          authController={this.props.authController}
+          inline={true}
+        />
+      </div>
+    </div>
+  );
+
+  render() {
+    if (!this.props.userSession) {
+      return this.requestLogin();
```
eliperelman: Having the login links in the body seems a little weird here to me, but I could be convinced otherwise. What about just having an `Error` here asking the user to use the login menu to sign in? That way we aren't duplicating login menu logic.

src/views/AuthNew/ClientEditor.js
```diff
@@ -0,0 +1,172 @@
...
+
+    return (
+      <div className={styles.editor}>
+        <h4 style={{ marginTop: 0 }}>Create New Client</h4>
+        <hr style={{ marginBottom: 10 }} />
+        <Form horizontal>
+          {!isNewClient && (
+            <FormGroup>
+              <Col componentClass={ControlLabel} sm={2}>
+                <ControlLabel>Old ClientId</ControlLabel>
```
eliperelman: Instead of "old", let's say "previous".

src/views/AuthNew/ClientEditor.js
```diff
@@ -0,0 +1,172 @@
...
+          <FormGroup controlId="expires">
+            <Col componentClass={ControlLabel} sm={2}>
+              Expires
+            </Col>
+            <Col sm={10}>
+              <TimeInput
+                value={new Date(currentClient.expires)}
+                onChange={this.handleExpiresChange}
+                className="form-control"
+              />
```
eliperelman: Move closing bracket to previous line.

src/views/AuthNew/ClientEditor.js
```diff
@@ -0,0 +1,172 @@
...
+                  <FormControl
+                    type="text"
+                    disabled={true}
+                    value={client.clientId}
+                  />
+                  <InputGroup.Button>
+                    <Button
+                      onClick={this.props.resetAccessToken}
+                      bsStyle="warning">
+                      <Glyphicon glyph="fire" /> Reset accessToken
```
eliperelman: Switch "fire" icon to another more appropriate one.

eliperelman:
So far everything looks pretty good. The only major thing is the login links. I'm still thinking we should direct the user to the existing sign in methods in the toolbar rather than re-duplicating that logic into the body of the page.

djmitche:
Sorry this took me so long to get to :(

src/views/AuthNew/AuthNew.js
djmitche: Agree.  It doesn't have to be an `Error`, but something saying "hey, please sign in" is good.  An arrow pointing to the menu might be cool as a guide to the user.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+  componentWillMount() {
+    this.load(this.props);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.load(nextProps);
+  }
+
+  handleResetAccessToken = () => {
+    const { credentials } = this.props.userSession;
```
djmitche: When I'm signed in:
```
Unhandled promise rejection 
TypeError: userSession.credentials is undefined
```

Note that [per the docs](https://github.com/taskcluster/taskcluster-tools/blob/master/src/auth/UserSession.js) this field is only available when UserSession.type is `credentials` (which is to say, a manual login).  Instead, you want to use the async `getCredentials` method.  See the client editor for how that gets loaded.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+
+    this.triggerCallback(this.state.client.clientId, credentials.accessToken);
```
djmitche: You're returning an unrelated clientId and accessToken.  That won't work :)

Instead, you'll need to call `auth.resetAccessToken(this.state.client.clientId)` and then trigger the callback with the appropriate access token.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+  };
+
+  renderUserOption = () => (
+    <div className={styles.textCenter}>
+      <h4>
+        The accessToken is expiring{' '}
+        <strong>{<DateView date={this.state.client.expires} />}</strong>. What
+        do you want to do?
+      </h4>
```
djmitche: The general message here is "This clientId already exists.  You can re-use it by resetting the access token, but that will cause any other uses of that clientId (with the old access token) to stop working."  I don't think the expiration date of the existing clientId is really relevant.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+    };
+  };
+
+  load = async ({ userSession, location, auth }) => {
+    if (!userSession) {
+      return;
+    }
+
+    const { name } = parse(location.search.slice(1));
+    const id = `${userSession.credentials.clientId}/${name}`;
```
djmitche: Check what the client editor does here -- `userSession.credentials` is generally not set.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  handleResetAccessToken = () => {
+    const { credentials } = this.props.userSession;
+
+    this.triggerCallback(this.state.client.clientId, credentials.accessToken);
+  };
+
+  triggerCallback = (clientId, accessToken) => {
+    const callbackUrl = parse(this.props.location.search.slice(1)).callback_url;
```
djmitche: What do you think about parsing all of the URL parameters into state when the component mounts, and resetting the location?

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+
+  constructClient = previousClient => {
+    const nextClient = parse(this.props.location.search.slice(1));
+    const description =
+      nextClient.description ||
+      `Client created ${new Date()} for ${nextClient.callback_url}`;
+
+    return {
+      ...previousClient,
+      ...{
```
djmitche: I *think* this notation is redundant, right?

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+        clientId: `${this.props.userSession.credentials
+          .clientId}/${nextClient.name}`,
```
djmitche: Check what the client editor does here -- `userSession.credentials` is generally not set.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+      return;
+    }
+
+    const { name } = parse(location.search.slice(1));
+    const id = `${userSession.credentials.clientId}/${name}`;
+
+    try {
+      const previousClient = (await auth.listClients({
+        prefix: id
+      })).filter(({ clientId }) => clientId === id)[0];
```
djmitche: Why list a subset and then search it for a single client?  `auth.client` can give you the info you need :)

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+      const client = this.constructClient(previousClient);
+
+      if (!previousClient) {
+        return this.setState({ client, creating: true, isNewClient: true });
```
djmitche: `isNewClient` is a little confusing -- any time `creating` is true, we'll be creating a new client -- creating an existing client doesn't make sense.  I think this is intended to mean that the user needs to invent a new client name.  Maybe `clientAlreadyExists`?

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+      <div>
+        {this.state.error && <Error error={this.state.error} />}
+        {!this.state.creating && this.state.client && this.renderUserOption()}
+        {this.state.creating &&
+          this.state.client && (
+            <ClientEditor
+              isNewClient={this.state.isNewClient}
+              client={this.state.client}
+              createClient={this.createClient}
+              userSession={this.props.userSession}
```
djmitche: the component appears not to use this prop.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+              resetAccessToken={this.resetAccessToken}
+            />
```
djmitche: Eli let you off the hook on this dangling closing bracket!

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+
+    return (
+      <div>
+        {this.state.error && <Error error={this.state.error} />}
+        {!this.state.creating && this.state.client && this.renderUserOption()}
+        {this.state.creating &&
+          this.state.client && (
+            <ClientEditor
+              isNewClient={this.state.isNewClient}
+              client={this.state.client}
```
djmitche: How about `existingClientId={this.state.client && this.state.client.clientId}` -- avoiding entirely the need for `isNewClient` or `clientAlreadyExists`.

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+              createClient={this.createClient}
```
djmitche: should this be `onCreateClient={this.handleCreateClient}`?

src/views/AuthNew/AuthNew.js
```diff
@@ -0,0 +1,158 @@
...
+              userSession={this.props.userSession}
+              resetAccessToken={this.resetAccessToken}
```
djmitche: In your screenshots this looked like it might be confusing -- this is re-presenting the option (to reset the accessToken of the existing client) that the user has already rejected.  Maybe remove that altogether, along with the display of the existing clientId?

src/views/AuthNew/ClientEditor.js
djmitche: It's not really "old" either..

djmitche:
It's worth noting that most of these clients will expire within a few days, so someone with `tctalker-395` would have to run `tctalker login` 396 times in three days.  For the most part people would see single digits.

src/views/AuthNew/AuthNew.js
helfi92: I can see why we would want to parse the URL parameters into `state.query`, but why do you want to reset the location (i.e., remove the queries from the URL)? I feel like we should just wait until the redirect to the callback URL happens.

src/views/AuthNew/AuthNew.js
helfi92: This simplifies things a lot :)

src/views/AuthNew/AuthNew.js
helfi92: For now prettify is forcing me to do that. After this gets merged I can modify the prettify configuration to inline the closing bracket :)

src/views/AuthNew/AuthNew.js
djmitche: I don't really mind either way, just a mild preference for not duplicating information.  Once the page is loaded and all of the components are mounted, that information has been communicated and is no longer necessarily up-to-date.

src/views/AuthNew/AuthNew.js
djmitche: I don't actually mind I was just trying to get in on the code-style action.

helfi92:
@djmitche I love the idea of appending `-1`,`-2`, etc. Thanks for the suggestion. I'll have it implemented.

src/views/ClientCreator/ClientCreator.js
```diff
@@ -0,0 +1,194 @@
...
+      this.triggerCallback(client.clientId, client.accessToken);
+    } catch (error) {
+      this.setState({ error });
+    }
+  };
+
+  triggerCallback = (clientId, accessToken) => {
+    window.location.replace(
+      `${this.state.query
+        .callback_url}?clientId=${clientId}&accessToken=${accessToken}`
```
djmitche: I wouldn't call this "pretty", prettify :(

src/views/ClientCreator/ClientCreator.js
```diff
@@ -0,0 +1,194 @@
...
+
+  constructClient = async clientName => {
+    const description =
+      this.state.query.description ||
+      `Client created ${new Date()} for ${this.state.query.callback_url}`;
+    const clientId = await this.nextAvailableClientId(clientName, 0);
+
+    return {
+      clientId,
+      scopes: this.state.query.scope || [],
```
djmitche: With just one scope: `Warning: Failed prop type: Invalid prop `scopes` of type `string` supplied to `ScopeEditor`, expected an array.`

helfi92: Good catch. I was under the impression that scope will always be an array if provided. I'll fix.

src/views/ClientCreator/ClientCreator.js
```diff
@@ -0,0 +1,198 @@
...
+
+        this.triggerCallback(newClient.clientId, newClient.accessToken);
+      } catch (error) {
+        this.setState({ error });
+      }
+    });
+  };
+
+  async nextAvailableClientId(clientName, suffix) {
+    const clientWithSuffix = (clientName, suffix) =>
```
eliperelman: This function call is unnecessary, as you can inline the return value directly:

```js
const clientWithSuffix = suffix ? `${clientName}-${suffix}` : clientName;

try {
   await this.props.auth.client(clientWithSuffix);

  return this.nextAvailableClientId(clientName, suffix + 1);
} catch (err) {
  return clientWithSuffix;
}
```

src/views/ClientCreator/ClientCreator.js
```diff
@@ -0,0 +1,198 @@
...
+    if (!userSession) {
+      return;
+    }
+
+    try {
+      const { clientId } = await userSession.getCredentials();
+
+      this.setState({ clientPrefix: clientId });
+
+      return clientId;
```
eliperelman: The return value of this method is never used, and probably shouldn't be since they should rely on updated state rather than this return value. Delete lines 121, 122, 125, and 126.

PR #365, opened 2017-10-19
Added ScopeGrants tool for easy granting of scopes

jonasfj:
This is the scope-grant tool I was talking about...

 * Patterns are easy to read, see `patterns.js`; adding new ones are easy too,
 * This far more robust than manually copy/pasting role-scope patterns we've agreed on,
 * This tool can list existing instantiations of selected pattern
   (this is done by substituting the regular expressions for the parameters into the role/scope-patterns and matching them against the existing roles; yes, it's complicated xcrossproduct magic..)

Some screenshots:  
 * https://screenshots.firefox.com/YVkSJBVTq17tyzV1/localhost
 * https://screenshots.firefox.com/f9ws1ZgBAqacO0zQ/localhost
 * https://screenshots.firefox.com/6QxiUhyNZuaY5Mzj/localhost

I hope this can become a good way to document the role-scope patterns we use.
I deliberately keep any scripting away from `patterns.js`, so that the patterns are strictly declarative.

Immediate use-cases:
 * Project creation,
 * Authorization of github repositories,
 * Setting up the correct role for a new workerType,
 * Granting project ownership to an LDAP group,

Over time I think we could end up with many more...
@djmitche, I think this could be used to define projects... I agree with your concerns regarding updating the patterns, I've outlined a possible solution in `patterns.js` along with a warning against modifing patterns...

@eliperelman, I recognize that the `ScopeGrants` component is a but clumpsy, and maybe in some future we refactor it to a few sub-components.. But I don't want to over-engineer this before we have thought a bit more about how we want to solve the problem with modified patterns...

### DIFF
--- a/src/views/ScopeGrants/ScopeGrants.js
+++ b/src/views/ScopeGrants/ScopeGrants.js
@@ -66,6 +66,9 @@ export default class ScopeGrants extends React.PureComponent {
     }
   }
 
+  handleGrantsRedirect = () => this.props.history.push(`/auth/grants`);
+  handleNewGrant = () => this.setState({ selected: null, args: {} });
+
   renderGrants() {
     if (this.state.error) {
       return <Error error={this.state.error} />;
@@ -85,7 +88,7 @@ export default class ScopeGrants extends React.PureComponent {
           <HelmetTitle title={pattern.title} />
           <Row>
             <Col md={1}>
-              <Button onClick={() => this.props.history.push(`/auth/grants/`)}>
+              <Button onClick={this.handleGrantsRedirect}>
                 <Glyphicon glyph="chevron-left" /> Back
               </Button>
             </Col>
@@ -102,20 +105,22 @@ export default class ScopeGrants extends React.PureComponent {
               <Table hover>
                 <thead>
                   <tr>
-                    {params.map((param, index) => <th key={index}>{param}</th>)}
+                    {params.map((param, index) => (
+                      <th key={`grant-${index}`}>{param}</th>
+                    ))}
                   </tr>
                 </thead>
                 <tbody>
                   {instantiatedArgs.map((args, index) => (
                     <tr
-                      key={index}
+                      key={`args-${index}`}
                       style={{ cursor: 'pointer' }}
                       className={
                         R.equals(args, this.state.selected) ? 'info' : null
                       }
                       onClick={() => this.setState({ selected: args })}>
                       {params.map((param, index) => (
-                        <td key={index}>
+                        <td key={`params-${index}`}>
                           <code>{args[param]}</code>
                         </td>
                       ))}
@@ -124,17 +129,16 @@ export default class ScopeGrants extends React.PureComponent {
                 </tbody>
               </Table>
               <hr />
-              <Button
-                onClick={() => this.setState({ selected: null, args: {} })}>
+              <Button onClick={this.handleNewGrant}>
                 <Glyphicon glyph="plus" /> New Grant
               </Button>
               <br />
               <br />
             </Col>
             <Col md={6}>
-              {!this.state.selected
-                ? this.renderCreateForm(pattern)
-                : this.renderPatternInstance(pattern, this.state.selected)}
+              {this.state.selected
+                ? this.renderPatternInstance(pattern, this.state.selected)
+                : this.renderCreateForm(pattern)}
             </Col>
           </Row>
         </Col>
@@ -142,56 +146,61 @@ export default class ScopeGrants extends React.PureComponent {
     );
   }
 
+  handleSave = async pattern => {
+    const instance = instantiate(pattern, this.state.args);
+    const roles = await this.props.auth.listRoles();
+    const toCreate = [];
+    const toUpdate = [];
+
+    Object.keys(instance).forEach(r => {
+      if (roles.some(({ roleId }) => r === roleId)) {
+        toUpdate.push(r);
+      } else {
+        toCreate.push(r);
+      }
+    });
+
+    await Promise.all([
+      ...toCreate.map(role =>
+        this.props.auth.createRole(role, {
+          description: '',
+          scopes: instance[role]
+        })
+      ),
+      ...toUpdate.map(role => {
+        const { scopes, description } = roles.find(
+          ({ roleId }) => roleId === role
+        );
+
+        return this.props.auth.updateRole(role, {
+          description,
+          scopes: R.uniq([...scopes, ...instance[role]])
+        });
+      })
+    ]);
+  };
+
+  handleCreatedCompleted = async () => {
+    const { args } = this.state;
+
+    this.setState({ selected: null, roles: null, args: {} });
+    await this.load();
+    this.setState({ selected: args });
+  };
+
+  handleParamChanged = (param, e) => {
+    this.setState({
+      args: R.merge(this.state.args, { [param]: e.target.value })
+    });
+  };
+
+  validateArg = (pattern, param) =>
+    this.state.args[param] &&
+    pattern.params[param].exec(this.state.args[param]);
+
   renderCreateForm(pattern) {
     const params = Object.keys(pattern.params);
     const { args } = this.state;
-    const save = async () => {
-      const instance = instantiate(pattern, this.state.args);
-      const roles = await this.props.auth.listRoles();
-      const toCreate = [];
-      const toUpdate = [];
-
-      Object.keys(instance).forEach(r => {
-        if (roles.some(({ roleId }) => r === roleId)) {
-          toUpdate.push(r);
-        } else {
-          toCreate.push(r);
-        }
-      });
-
-      await Promise.all([
-        ...toCreate.map(role =>
-          this.props.auth.createRole(role, {
-            description: '',
-            scopes: instance[role]
-          })
-        ),
-        ...toUpdate.map(role => {
-          const { scopes, description } = roles.find(
-            ({ roleId }) => roleId === role
-          );
-
-          return this.props.auth.updateRole(role, {
-            description,
-            scopes: R.uniq([...scopes, ...instance[role]])
-          });
-        })
-      ]);
-    };
-
-    const reload = async () => {
-      this.setState({ selected: null, roles: null, args: {} });
-      await this.load();
-      this.setState({ selected: args });
-    };
-
-    const setArg = (param, text) => {
-      this.setState({ args: R.merge(this.state.args, { [param]: text }) });
-    };
-
-    const validateArg = param =>
-      this.state.args[param] &&
-      pattern.params[param].exec(this.state.args[param]);
 
     return (
       <span>
@@ -200,7 +209,9 @@ export default class ScopeGrants extends React.PureComponent {
           {params.map((param, index) => (
             <FormGroup
               key={index}
-              validationState={validateArg(param) ? 'success' : 'error'}>
+              validationState={
+                this.validateArg(pattern, param) ? 'success' : 'error'
+              }>
               <Col componentClass={ControlLabel} sm={2}>
                 {param}
               </Col>
@@ -208,7 +219,7 @@ export default class ScopeGrants extends React.PureComponent {
                 <FormControl
                   type="text"
                   placeholder={pattern.params[param].source.slice(1, -1)}
-                  onChange={e => setArg(param, e.target.value)}
+                  onChange={e => this.handleParamChanged(param, e)}
                 />
                 <FormControl.Feedback />
               </Col>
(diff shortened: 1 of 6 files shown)


djmitche:
Looking at the screenshots -- nice work on this :)

I have a few big-picture issues with this, though.

1. It embeds logic for conventions into the source code of tools.  What sort of patterns would and wouldn't be appropriate there?  Gecko workerTypes?  Mozilla-LDAP-related roles?  Shipit-related roles?  Are we open to anyone who would like to define a pattern defining one?
1. This fills the same niche as taskcluster-admin.  Some of the scripts in that tool will be [moved into the tree](https://github.com/taskcluster/taskcluster-rfcs/issues/91), but maybe others will find a home here, if we can answer the previous point?
1. The role-recognition stuff worries me, and the warnings in `pattern.js` don't make me feel better :).  The approach in tc-admin is to use authoritative upstream data sources -- the list of workerTypes, list of branches, etc. -- to define what the roles should look like, then present a diff and/or modify the roles to look like that.  This is much more in line with other configure-my-cloud-service utilities like terraform or ansible do.  But like those tools, I don't think that requires a UI.

jonasfj:
> It embeds logic for conventions into the source code of tools. What sort of patterns would and wouldn't be appropriate there? Gecko workerTypes? Mozilla-LDAP-related roles? Shipit-related roles? Are we open to anyone who would like to define a pattern defining one?

I didn't consider this before, but in theory we could store the patterns as JSON objects in secrets.
So secrets named `grant-patterns/<name>` constitute a grant pattern loaded by taskcluster-tools and displayed here. That way people can define their own patterns, we could even give `secrets:put:grant-patterns/project:*` to project admins.

In such case, re-instantiating the pattern for all existing grants would be easy to do as part of the pattern modification process.

---

The motivation for a UI is that:
 * it's easy to click the buttons,
 * doesn't require people to install anything,
 * give a great overview to people who aren't necessarily us.

It's basically, roles-scopes assignments made more high-level.

---
Note: the _role-recognition stuff_ is mathematically sound, as in it will detect all pattern instantiations that exists in the roles from tc-auth. The qualifier is that partial pattern instantiations won't be detected.
When we have parameterized roles most patterns will take the form: `rolePrefix<param>roleSuffix: ['assume:paramRole:<param>']` where `assume:paramRole:<param>` is a parameterized role. This would prevent the pit-falls where partial pattern instantiations goes undetected.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+
+          return this.props.auth.updateRole(role, {
+            description,
+            scopes: R.uniq([...scopes, ...instance[role]])
+          });
+        })
+      ]);
+    };
+
+    const reload = async () => {
```
eliperelman: This can be moved to a class property:

```js
reload = async () => {
  this.setState({ selected: null, roles: null, args: {} });
  await this.load();
  this.setState({ selected: args });
};
```

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+      this.setState({ selected: null, roles: null, args: {} });
+      await this.load();
+      this.setState({ selected: args });
+    };
+
+    const setArg = (param, text) => {
```
eliperelman: Move this to a class property with a minimal arrow function or bind inline in the JSX:

```jsx
// Choose appropriate name
handleParamChange = (e, param) => {
  this.setState({ args: R.merge(this.state.args, { [param]: e.target.value }) });
};

renderCreateForm(pattern) {
  // ...
    onChange={e => this.handleParamChange(e, param)}
```

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+      this.setState({ args: R.merge(this.state.args, { [param]: text }) });
+    };
+
+    const validateArg = param =>
```
eliperelman: Move to class property:

```jsx
validateArg = (param) => this.state.args[param] &&
  pattern.params[param].exec(this.state.args[param]);

validationState={this.validateArg(param) ? 'success' : 'error'}>
```

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+            </Col>
+          </FormGroup>
+        </Form>
+      </span>
+    );
+  }
+
+  renderPatternInstance(pattern, args) {
+    const params = Object.keys(pattern.params);
+    const reload = () => {
```
eliperelman: Move to class property.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+      this.setState({ selected: null, roles: null });
+      this.load();
+    };
+
+    const remove = async () => {
```
eliperelman: Move to class property.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+    };
+
+    return (
+      <span>
+        <h3>Selected Parameters</h3>
+        <dl className="dl-horizontal">
+          {R.unnest(
+            params.map((param, index) => [
+              <dt key={param}>{param}</dt>,
+              <dd key={index}>
```
eliperelman: Avoid numeric-only indices to avoid potential key conflicts by giving stronger key names.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+        {this.renderInstanceGrants(pattern, args)}
+        <hr />
+        <ModalItem
+          button={true}
+          bsStyle="danger"
+          onSubmit={remove}
+          onComplete={reload}
+          body={
+            <span>
+              Are you sure you want to <b>remove scope grants</b> from role?
```
eliperelman: Prefer semantic `<strong>` tags instead of `<b>` tags.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+            </span>
+          }>
+          <Icon name="trash" /> Delete Grant
+        </ModalItem>
+      </span>
+    );
+  }
+
+  renderInstanceGrants(pattern, args) {
+    let inst;
```
eliperelman: Prefer `const`:

```jsx
renderInstanceGrants(pattern, args) {
  try {
    const inst = instantiate(pattern, args);

    return (
      <ul>
        {/* ... */}
      </ul>
    );
  } catch (err) {
    return <Error error={err} />
  }
}
```

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+      </ul>
+    );
+  }
+
+  renderPatternSelector() {
+    return (
+      <Col smOffset={1} sm={10}>
+        <h2>Scope Grant Patterns</h2>
+        <p>
+          This <i>scope grant</i> tool is an administrative utility for granting
```
eliperelman: Prefer semantic `<em>` tag over `<i>` tag.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+          scopes by creating/updating roles following predefined patterns. These
```
eliperelman: replace `/` with ` or `.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+          patterns are declared in the taskcluster-tools repository. Each
+          pattern takes a set of parameters that is used to instantiate scope
+          patterns which are then granted to instantiated role patterns. Below
+          is a list of patterns supported, after selecting a pattern, this tool
```
eliperelman: Change to: "Below is a list of supported patterns. After selected a pattern, this tool..."

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+          will display existing instantiations of the pattern by inspecting
+          existing roles and offer a UI for creating new instantiations.
```
eliperelman: Let's change "offer a UI" to something else. Maybe "offer options" or something.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+        </p>
+        <br />
+        <ListGroup>
+          {PATTERNS.map(({ name, title, icon, description }, index) => (
+            <ListGroupItem
+              key={index}
```
eliperelman: Avoid numeric-only indices to avoid potential key conflicts by giving stronger key names.

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+              header={
+                <h3>
```
eliperelman: Wrap multi-line JSX in parens:

```jsx
header={(
  <h3>
    <Icon name={icon} /> {title}
  </h3>
)}
```

src/views/ScopeGrants/granting.js
```diff
@@ -0,0 +1,171 @@
+import assert from 'assert';
```
eliperelman: Not optimal, but let's avoid the use of internal Node.js modules. Prefer `throw`ing over `assert`ing.

jonasfj:
@eliperelman, I was talking to dustin last week... and one borderline crazy idea would be to save `patterns.js` in secrets... Ie. load patterns from secrets named `public/scope-patterns/<project>/<name>`.

That way we can have taskcluster patterns and project specific patterns, and each project can define it's own scope patterns.

Anyways, this would definitely be a future iteration :)

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+      selected: null,
+      args: {}
+    };
+  }
+
+  componentWillMount() {
+    this.load();
+  }
+
+  componentWillReceiveProps(nextProps) {
```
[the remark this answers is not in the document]
jonasfj: If the `window.location.hash` changes... regardless it's just good practice...

src/views/ScopeGrants/ScopeGrants.js
```diff
@@ -0,0 +1,377 @@
...
+    const instantiatedArgs = instantiatedArguments(pattern, this.state.roles);
+    const params = Object.keys(pattern.params);
+
+    return (
+      <Row>
+        <Col md={12}>
+          <HelmetTitle title={pattern.title} />
+          <Row>
+            <Col md={1}>
+              <Button onClick={() => this.props.history.push(`/auth/grants/`)}>
```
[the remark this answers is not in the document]
jonasfj: Okay, but I have to say this is ugly...

couldn't we use something https://www.npmjs.com/package/react-autobind
In the constructor... or even better in the constructor for a base class that we inherit everything from? (probably a topic for a different day, just an idea)

src/views/ScopeGrants/ScopeGrants.js
jonasfj: I do not want to catch errors  from the rendering of `<ul>`

src/views/ScopeGrants/ScopeGrants.js
jonasfj: linter says otherwise...

src/views/ScopeGrants/ScopeGrants.js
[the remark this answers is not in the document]
eliperelman: > Okay, but I have to say this is ugly...

lol. Gonna have to get used to this one, seeing as how class properties are stage 3, this is going to be part of the language soon.

src/views/ScopeGrants/ScopeGrants.js
eliperelman: There aren't going to be `ul` rendering errors. If there are, those need to be caught as well.

Regardless, I'm fine with keeping the `let` if you think it keeps things readable here. :+1:

src/views/ScopeGrants/ScopeGrants.js
eliperelman: Hmmmm, not sure how that happened, it's supposed to catch those. I can take a look later. Thanks for pointing that out!

eliperelman:
@jonasfj I'm curious if that means we are going to be executing remote code, or if these are reviewed somehow.

PR #366, opened 2017-10-19
change tooltip

djmitche:


src/auth/Auth0LoginMenuItem.js
```diff
@@ -7,8 +7,8 @@ export default class Auth0LoginMenuItem extends React.PureComponent {
   render() {
     const tooltip = (
       <Tooltip id="auth0-signin">
-        We are testing this login method, which will be live in October. Please
-        try it and report any issues!
+        Sign in with the LDAP account you use to push to version control, or
+        with email if you do not have version-control access.
```
eliperelman: Should probably make "version control" consistent without the hyphen.

djmitche: done

PR #367, opened 2017-10-24
Bug 1404432 - put authController in context

djmitche:
Looking for advice.  Things I don't like:

* Clients renders twice, since it doesn't seem possible to create initial state based on context -- accessing `this.context` before `componentDidMount` throws an exception.
* There's still a lot of boilerplate for components that want to monitor for userSession changes.  Is there a way to build a HOC that calls a defined method on its child, or make a mixin, or something of the sort so I could just define `handleUserSessionChanged` and have the rest happen by magic?


src/auth/AuthController.js
```diff
@@ -9,9 +10,10 @@ import { renew as auth0Renew } from './auth0';
  * items, ongoing expiration monitoring, and any additional required UI.  It also handles synchronizing
  * sign-in status across tabs.
  */
-export default class AuthController {
+export default class AuthController extends EventEmitter {
```
eliperelman: Maybe we can use the same mitt library that taskcluster-client-web uses?

src/components/Clients/index.js
```diff
@@ -1,28 +1,47 @@
...
+
   static propTypes = {
     children: func.isRequired
   };
 
   constructor(props) {
     super(props);
-    this.state = this.getClients(props);
+
+    this.handleUserSessionChanged = this.handleUserSessionChanged.bind(this);
```
eliperelman: These binds are unnecessary if you define `handleUserSessionChanged` as an arrow function class property.

src/components/Clients/index.js
```diff
@@ -33,11 +52,15 @@ export default class Clients extends React.PureComponent {
...
-      {}
+      { userSession }
     );
   }
 
   render() {
-    return this.props.children(this.state);
+    const { ready, ...childProps } = this.state;
+
+    // TODO: this gets called twice, since context is set at mount time and
```
eliperelman: Maybe this is rendering twice because you are no longer guarding the re-render when props change in `componentWillReceiveProps`?

src/components/CredentialsMenu/index.js
```diff
@@ -15,8 +16,37 @@ const ManualLoginMenuItem = loadable(() =>
 );
 
 class CredentialsMenu extends React.PureComponent {
+  static contextTypes = {
+    authController: object.isRequired
+  };
+
+  constructor(props) {
+    super(props);
+    this.handleUserSessionChanged = this.handleUserSessionChanged.bind(this);
```
eliperelman: Remove bind in favor of arrow function class property.

src/views/AwsProvisioner/index.js
```diff
@@ -21,12 +20,11 @@ const View = ({
   }
 
   return (
-    <Clients userSession={userSession} Queue AwsProvisioner={{ baseUrl }}>
-      {clients => (
+    <Clients Queue AwsProvisioner={{ baseUrl }}>
+      {props => (
```
eliperelman: Why rename clients to props?

src/components/Clients/index.js
```diff
@@ -1,28 +1,47 @@
...
+  componentWillUnmount() {
+    this.context.authController.removeListener(
+      'user-session-changed',
+      this.handleUserSessionChanged
+    );
+  }
+
+  getClients() {
+    const { children, ...clients } = this.props;
+    const userSession = this.context.authController.getUserSession();
```
eliperelman: I feel like the scope of this class is expanding beyond its original intent. Should this work be done in App?

src/views/CredentialsManager/index.js
```diff
@@ -1,174 +1,11 @@
...
-import DateView from '../../components/DateView';
+import Clients from '../../components/Clients';
+import CredentialsManager from './CredentialsManager';
 
-export default class CredentialsManager extends React.PureComponent {
-  constructor(props) {
-    super(props);
+const View = () => (
+  <Clients>
+    {({ userSession }) => <CredentialsManager userSession={userSession} />}
```
eliperelman: Doesn't `userSession` come from a different place than context, or is this being proxied? It wasn't immediately apparent which usage should be correct. If `userSession` comes from context, albeit indirectly, why should we pass it as props like this?

src/components/Clients/index.js
djmitche: No, it renders once with `state.userSession` set to null and then again when it is set to an actual UserSession via `componentDidMount`.  I tried doing the same in `componentWillMount`, but `this.context` isn't available yet at that point.

src/views/CredentialsManager/index.js
djmitche: I'm following the best-practice for context, which is to only put static values there (like AuthController).  That means anyone who needs `userSession` has to jump through some hoops to wait for the component to mount, call context.authController.getUserSession(), and then listen for user-session-changed events.  I added this functionality to Clients since it seemed to be well-positioned for the purpose, but it is an odd fit as you mentioned in another comment.  At any rate, passing the userSession down from Clients as props where necessary makes sense, I think.

src/components/Clients/index.js
djmitche: Maybe I need another HOC like Clients that will also provide a userSession as a prop.  You're right that this is kind of overloading the meaning of Clients.

src/auth/AuthController.js
djmitche: Yeah, I guess I didn't test this enough to find out that EventEmitter is not defined in a browser :)

djmitche:
Still an open TODO on the double-render.  The issue is that React does not seem to allow access to `this.context` before the component is mounted, yet we need to use `this.context.authController` to get the current `userSession` in order to render children.

djmitche:
Hm, I have some funnier business I can do.  Apparently `this.context` *is* available from render methods, so I can use it there, but then also modify state to trigger a re-render when that value changes.

djmitche:
Well, that's a little better.  Because userSession can't be determined before the first render, we just always fetch it from the authController when necessary, then forceUpdate on user-session-changed.  This only occurs in three places (WithClients, WithUserSession, and CredentialsMenu).

djmitche:
@eliperelman ping?

eliperelman:
I've been reviewing this little-by-little, and for the most part I think it's ok, just had to get my head wrapped around the context paradigm. I do think there is a bit of risk with this, so that's why I haven't pushed on getting it merged yet. Unless there is something blocked by this, which I don't think there is, we will probably hold off on merging this until after the change freeze.

djmitche:
That's fair.  I'm just worried about bitrot.

eliperelman:
I'm not too concerned about bitrot at the moment since we won't be doing a bunch of landing into master while in the change freeze.

djmitche:
@eliperelman let's land this!

src/components/CredentialsMenu/index.js
```diff
@@ -15,8 +16,29 @@ const ManualLoginMenuItem = loadable(() =>
...
+  static contextTypes = {
+    authController: object.isRequired
+  };
+
+  handleUserSessionChanged = () => {
+    this.forceUpdate();
+  };
+
+  componentDidMount() {
+    const { authController } = this.context;
```
eliperelman: Since we are only using `authController` in a single usage, we can make it like `componentWillUnmount`:

```js
this.context.authController.on(
  'user-session-changed',
  this.handleUserSessionChanged
);
```

djmitche:
I'll flag you for review again after un-bitrotting..

PR #368, opened 2017-10-25
Show Group Status through the favicon

helfi92:


eliperelman:
No white, but I'd say a gap, that way it is transparent. I can help edit the image if you need.

helfi92:
👀 @djmitche @eliperelman 

screen shot 2017-10-25 at 4 39 45 pm

helfi92:
I'll send a different PR to remove the pop notification system we have after this gets merged.

djmitche:
I like :)

src/views/UnifiedInspector/GroupProgress.js
```diff
@@ -96,6 +97,20 @@ export default class GroupProgress extends React.PureComponent {
     );
   }
 
+  handleGroupStateChange(percents) {
```
eliperelman:
```js
handleGroupStateChange(percents) {
  if (percents.completed === 100) {
    this.props.onGroupStateChange('success');
  } else if (percents.failed) {
    this.props.onGroupStateChange('failed');
  } else {
    this.props.onGroupStateChange('pending');
  }
}
```

src/views/UnifiedInspector/Inspector.js
```diff
@@ -483,6 +491,24 @@ export default class Inspector extends React.PureComponent {
     return artifacts.filter(({ name }) => /^public\/logs\//.test(name));
   }
 
+  getFavicon = () => {
```
eliperelman: I don't think this needs to be an arrow function.

src/views/UnifiedInspector/Inspector.js
```diff
@@ -483,6 +491,24 @@ export default class Inspector extends React.PureComponent {
...
+    let FavIcon;
```
eliperelman:
```js
getFavicon() {
  switch (this.state.groupState) {
    case 'pending':
      return pendingFavIcon;
    case 'failed':
      return failedFavIcon;
    default:
      return successFavIcon;
  }
}
```

PR #370, opened 2017-10-27
Remove the notification system in favor of favicons

helfi92:
To be merged after https://github.com/taskcluster/taskcluster-tools/pull/368.


gregarndt:
We should probably figure out how often this feature is removed to understand if it's disruptive removing it.

eliperelman:
I agree. We won't be removing this right now as it's a different use case than the favicons.

eliperelman:
Going to close this for now, and we can revisit in the future if we find the need.

Issue #369, opened 2017-10-27
Inspector - filtered list of tasks isn't updated


nthomas-mozilla:
RelEng releaseduty spend time in the inspector to keep an eye on the state of release automation. The status bar seems to keep task counts for each state up to date, but the task list will get out of date. This regressed in the last week or so.

<img width="1153" alt="screen shot" src="https://user-images.githubusercontent.com/942272/32086215-1ee4669e-bb30-11e7-868b-bc6b5a6d90df.png">

Steps to reproduce:
* load a large & active task group, probably anything from a gecko push would be fine
* use the Status filter to select a state which isn't All, eg Running
* do something else while some tasks change state
* observe the task count for the filter has changed, but not the list

Workaround - change the filter to some other value and back again.

helfi92:
Resolved in #371 :)

nthomas-mozilla:
Thanks Hassan.

PR #371, opened 2017-10-30
Update list when task status changes

helfi92:
When a task status changes, the list is not updated. This is problematic when you're looking at filtered group tasks.


helfi92:
Should fix https://github.com/taskcluster/taskcluster-tools/issues/369 when merged.

PR #374, opened 2017-11-01
Create CODEOWNERS

eliperelman:
This is for branch protection during change freeze.

### DIFF
--- a/. github/CODEOWNERS
+++ b/. github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @eliperelman


djmitche:
You had me until "change freeze".  What will this accomplish?

eliperelman:
It will ensure that code isn't merged unless I review it, since I am actively weighing the risk of merging things into master right now.

djmitche:
Seems like overkill to me, but it doesn't hurt.

djmitche:
I'll defer to greg here.

gregarndt:
Eh, I do agree it might be overkill.  We haven't needed to implement this on the backend side, which is much more risky to make changes that could impact production.

There's nothing wrong with this, so I won't block from landing, I just think it's unnecessary.  Those with permissions to merge to master should understand consequences.

PR #376, opened 2017-11-02
Attach actions to the the worker-types, workers, and worker page

helfi92:


helfi92:
This PR adds action buttons to the the worker-types, workers, and worker page. When an action is triggered, a POST request is constructed ([see how actions are constructed](https://docs.taskcluster.net/reference/platform/taskcluster-queue/docs/actions)) and sent. If the request doesn't throw, a notification banner pops out at the bottom of the screen notifying the user that the action was successful. Otherwise the Error component will take care of the error.

src/views/WorkerManager/WorkerManager.js
```diff
@@ -159,6 +161,35 @@ export default class WorkerManager extends React.PureComponent {
...
+      workerId
+    } = this.state.worker;
+    const query = [
+      `provisionerId=${provisionerId}`,
+      `workerType=${workerType}`,
+      `workerGroup=${workerGroup}`,
+      `workerId=${workerId}`
+    ].join('&');
+
+    fetch(`${action.url}?${query}`, {
```
eliperelman: Using async functions:

```js
try {
  const response = await fetch(/* ... */);

  await response.json();
  this.notification.show(/* ... */);
} catch (err) {
  this.setState({ error: err });
}
```

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -127,6 +135,24 @@ export default class WorkerManager extends React.PureComponent {
     });
   };
 
+  handleActionClick = action => {
+    const query = `provisionerId=${this.props.provisionerId}`;
+
+    fetch(`${action.url}?${query}`, {
```
eliperelman: Use async function.

src/views/Workers/WorkerManager.js
```diff
@@ -111,6 +114,27 @@ export default class WorkerManager extends React.PureComponent {
       filter: filter.includes('disabled') ? 'disabled' : filter
     });
 
+  handleActionClick = action => {
+    const query = [
+      `provisionerId=${this.props.provisionerId}`,
+      `workerType=${this.props.workerType}`
+    ].join('&');
+
+    fetch(`${action.url}?${query}`, {
```
eliperelman: Use async functions.

src/views/Workers/WorkerManager.js
```diff
@@ -134,21 +158,42 @@ export default class WorkerManager extends React.PureComponent {
...
-          </DropdownButton>
+          <ButtonToolbar className={styles.buttonToolbar}>
+            <DropdownButton
+              id="workers-dropdown"
+              bsSize="small"
+              title={`Filter by: ${filter || 'None'}`}
+              onSelect={this.onFilterSelect}>
+              <MenuItem eventKey="None">None</MenuItem>
+              <MenuItem divider />
+              {['Status: disabled'].map((property, key) => (
```
eliperelman: There is only a single entry here, we can remove the map.

helfi92: I have a [pull-request]((https://github.com/taskcluster/taskcluster-queue/pull/222)) on  the taskcluster-queue repo which switches the disabled property to be quarantined and so I was already planning to make the appropriate changes to the UI in a different pull-request. I will take this change into consideration there.

src/components/Notification/index.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+import { Alert } from 'react-bootstrap';
+import styles from './styles.css';
+
+class Notification extends React.PureComponent {
```
eliperelman: Move export default to here.

src/views/Workers/styles.css
```diff
@@ -5,3 +5,7 @@
 .pagination {
   margin: 20px 0 20px 0;
 }
+
+.buttonToolbar > div{
```
eliperelman: Space before brace.

helfi92:
Pushed the last bit of nits. Should be okay now :)

src/components/Notification/index.js
```diff
@@ -0,0 +1,46 @@
...
+            <Alert className={styles.alert} bsStyle="success">
+              {this.state.message}
+            </Alert>
+          )}
+        </ReactCSSTransitionGroup>
+      </div>
+    );
+  }
+
+  show = message => this.setState({ message, show: true });
```
eliperelman: nit: make render the last method in a component.

helfi92:
@eliperelman I ended moving the actions in worker-types explorer to provisioner explorer as well as modifying the URL and method when sending the request. This needed to be done because there have been some changes in the backend.

src/components/Notification/index.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+import { Alert } from 'react-bootstrap';
+import styles from './styles.css';
+
+export default class Notification extends React.PureComponent {
```
eliperelman: In thinking about this component, I wonder if using refs is the right approach. I think it would be preferable to control this via props, but then I'm not sure how that would operate with the transitionary close.

I think for inspiration here, let's draw on the [`Snackbar` component from react-md](https://react-md.mlaursen.com/components/snackbars). It doesn't use refs, supports autoclose, and you can control the presence of the toast in state.

We also may want to rename this to Snackbar for future consistency with the Material work.

PR #378, opened 2017-11-06
Bug 1395349 - use the production site for devel logins

djmitche:
Use the production site for development logins

### DIFF
--- a/src/auth/DevelopmentLoginMenuItem.js
+++ b/src/auth/DevelopmentLoginMenuItem.js
@@ -22,11 +22,16 @@ export default class DevelopmentLoginMenuItem extends React.PureComponent {
   onSelect() {
     const query = new URLSearchParams();
 
-    query.set('target', new URL('/login/development', window.location));
+    query.set('callback_url', new URL('/login/development', window.location));
+    query.set('scope', 'xyz'); // defeat the default '*' and require user to enter desired scopes
+    query.set('name', 'tools-devel');
     query.set(
       'description',
-      'Taskcluster Tools offers various way to create and inspect both tasks and task groups.'
+      'Temporary client for developing and testing taskcluster-tools'
     );
-    window.open(`https://login.taskcluster.net/?${query.toString()}`, '_blank');
+    window.open(
+      `https://tools.taskcluster.net/auth/clients/new?${query.toString()}`,
+      '_blank'
+    );
   }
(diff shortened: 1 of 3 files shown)


src/auth/DevelopmentLoginMenuItem.js
```diff
@@ -22,11 +22,16 @@ export default class DevelopmentLoginMenuItem extends React.PureComponent {
   onSelect() {
     const query = new URLSearchParams();
 
-    query.set('target', new URL('/login/development', window.location));
+    query.set('callback_url', new URL('/login/development', window.location));
+    query.set('scope', 'xyz'); // defeat the default '*' and require user to enter desired scopes
+    query.set('name', 'tools-devel');
```
helfi92: Should we set the expiry date in the query or simply default to 3 days?

src/auth/DevelopmentLoginMenuItem.js
```diff
@@ -22,11 +22,16 @@ export default class DevelopmentLoginMenuItem extends React.PureComponent {
...
     query.set(
       'description',
-      'Taskcluster Tools offers various way to create and inspect both tasks and task groups.'
+      'Temporary client for use developing and testing taskcluster-tools'
```
helfi92: s/use//

src/views/DevelopmentLogin/index.js
```diff
@@ -11,8 +11,7 @@ export default class DevelopmentLogin extends React.Component {
     this.state = {
       credentials: {
         clientId: search.get('clientId'),
-        accessToken: search.get('accessToken'),
-        certificate: search.get('certificate')
```
helfi92: Out of curiosity, why do we not need certificate anymore?

src/auth/DevelopmentLoginMenuItem.js
djmitche: I'm OK with the default.

src/views/DevelopmentLogin/index.js
djmitche: These are now "permacreds" and not temporary credentials.

helfi92: Sounds good, thanks!

eliperelman:
This is only during development, so no risk. Merge when convenient.

PR #380, opened 2017-11-07
Omit all dependencies when retriggering task

eliperelman:


eliperelman:
Not sure what happened to my branch...

PR #381, opened 2017-11-07
Update client data when resetting access token

helfi92:
If a user decides to reset the access token, the client gets a new token but all other parameters provided in the login query (expires, description, scopes) are presently being neglected. This pull-requests updates the client prior to calling the callback.

### DIFF
--- a/src/views/ClientCreator/ClientCreator.js
+++ b/src/views/ClientCreator/ClientCreator.js
@@ -34,15 +34,32 @@ export default class ClientCreator extends React.PureComponent {
   }
 
   handleResetAccessToken = async () => {
-    try {
-      const client = await this.props.auth.resetAccessToken(
-        this.state.client.clientId
-      );
+    this.setState({ loading: true }, async () => {
+      try {
+        const description =
+          this.state.query.description ||
+          `Client created ${new Date()} for ${this.state.query.callback_url}`;
+        const { scope } = this.state.query;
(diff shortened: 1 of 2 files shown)


src/views/ClientCreator/ClientCreator.js
```diff
@@ -34,15 +34,32 @@ export default class ClientCreator extends React.PureComponent {
...
-      );
+    this.setState({ loading: true }, async () => {
+      try {
+        const description =
+          this.state.query.description ||
+          `Client created ${new Date()} for ${this.state.query.callback_url}`;
+        const { scope } = this.state.query;
+        const scopes = (typeof scope === 'string' ? [scope] : scope).filter(
+          s => s
+        );
```
djmitche: This looks like it could be moved to a utility function :)

eliperelman:
```js
const scopes = (Array.isArray(scope) ? scope : [scope]).filter(Boolean);
```

eliperelman: Put this in utils as `toArray`, similar to Neutrino:

https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino/src/utils.js#L18

You can use Ramda or not, your preference. :)

helfi92: Woah, I love this!

PR #382, opened 2017-11-08
Remove disabled property and use quarantine instead

helfi92:


src/views/Workers/WorkerManager.js
```diff
@@ -235,8 +228,10 @@ export default class WorkerManager extends React.PureComponent {
                       <td>
                         <Label
                           bsSize="sm"
-                          bsStyle={worker.disabled ? 'danger' : 'success'}>
-                          {worker.disabled ? 'disabled' : 'enabled'}
+                          bsStyle={
+                            worker.quarantineUntil ? 'danger' : 'success'
+                          }>
+                          {worker.quarantineUntil ? 'yes' : 'no'}
```
helfi92: I'm open to something different than `yes` and `no`.

helfi92:
Once I get an OK on this pull-request, I will merge https://github.com/taskcluster/taskcluster-queue/pull/222 before merging this one.

src/views/Workers/WorkerManager.js
djmitche: I think this is fine, but maybe a moment.js relative time?

"Quarantined: 13 days"

helfi92:
I will merge after the release of Firefox Quantum to mitigate the risks.

src/views/Workers/WorkerManager.js
helfi92: Great idea! I will add it in a subsequent PR so we can mitigate any weird errors. Right now It's a little bit annoying to test without having https://github.com/taskcluster/taskcluster-queue/pull/222 merged first.

PR #386, opened 2017-11-15
Add a kill all feature to aws-provisioner page

helfi92:


helfi92:
I'm planning to make use of the Notification component once it gets merged to notify the user when if the action was successful. This shouldn't stop the pull-request from being merged, however.

PR #392, opened 2017-11-20
Call the new endpoint to quarantine a worker

helfi92:


src/views/WorkerManager/WorkerManager.js
```diff
@@ -143,17 +143,17 @@ export default class WorkerManager extends React.PureComponent {
...
     try {
-      const worker = await this.props.queue.declareWorker(
+      const worker = await this.props.queue.quarantineWorker(
         provisionerId,
         workerType,
         workerGroup,
         workerId,
         {
-          disabled: !disabled
+          quarantineUntil: quarantineUntil ? null : fromNow('7 days')
```
helfi92: @djmitche I made quarantining a worker default to 7 days. Is that a good number or should it be larger?

src/views/WorkerManager/WorkerManager.js
```diff
@@ -203,13 +203,16 @@ export default class WorkerManager extends React.PureComponent {
...
     } = this.state;
     const { provisionerId, workerType, workerGroup, workerId } = this.props;
-    const disableTooltip = (
+    const quarantineTooltip = (
       <Tooltip id="tooltip">
-        {worker && worker.disabled
+        {worker && worker.quarantineUntil
           ? 'Enabling a worker will resume accepting jobs.'
-          : 'Disabling a worker allows the machine to remain alive but not accept jobs.'}
+          : 'Quarantining a worker allows the machine to remain alive but not accept jobs for 7 days.'}
```
eliperelman: Does this statement make sense if the value for quarantining can be variable?

helfi92: For the user, quarantining is not really a variable when setting it since it currently always defaults to a week. I'm not sure I understand what is not making sense in this statement.

eliperelman: The context is this:

```js
quarantineUntil: quarantineUntil ? null : fromNow('7 days')
```

Which supposedly can come from the worker.

eliperelman: So it's like we always ignore it, so I'm not sure what the purpose is.

src/views/WorkerManager/WorkerManager.js
djmitche: I thought this was going to be user-selectable, so that we can easily quarantine a machine "forever" (until 3017).  I think that expecting buildduty to click this button on a weekly basis for disabled workers is not a good design.

djmitche: Also, it looks like there's no way to update this date without re-enabling the worker.

helfi92:
Sounds good, I will re-work this to allow quarantineUntil to be user-selectable. Thanks for the feedback @eliperelman @djmitche .

djmitche:
A quick option that would make me (but probably not @jonasfj) happy would be to default the quarantine to 1000 years.  That keeps the user from having to type/select a number every time.

helfi92:
Made the quarantine button user selectable.
screen shot 2017-11-21 at 9 34 06 am

eliperelman:
Let's add some white space above the input, and make it label on top of input, with the input the full width.

helfi92:
screen shot 2017-11-21 at 1 29 47 pm

Issue #387, opened 2017-11-16
Expose dependencies short name and current icon status


lissyx:
When I am looking at a task that has dependencies, as of now: 
![screenshot-2017-11-16 deepspeech packages 1](https://user-images.githubusercontent.com/1645737/32909004-a3e78382-cb05-11e7-8d52-ba5ab4120419.png)

We only have a list of taskId. It makes it hard to track the current status of them. It would be cool if we had, in addition:
 - dependency's short name
 - current status of the dep (unscheduled, pending, running, ...)

PR #393, opened 2017-11-22
Add task name status to the list of dependencies

helfi92:
Resolves https://github.com/taskcluster/taskcluster-tools/issues/387

### DIFF
--- a/src/views/UnifiedInspector/TaskDetails.js
+++ b/src/views/UnifiedInspector/TaskDetails.js
@@ -200,7 +200,7 @@ export default class TaskDetails extends React.PureComponent {
             <tr>
               <td>Dependencies</td>
               <td>
-                {this.state.dependenciesLoading || task.dependencies.length ? (
+                {this.state.dependencies.length ? (
                   <Table condensed>
                     <thead>
                       <tr>
--- a/src/views/UnifiedInspector/TaskDetails.js
+++ b/src/views/UnifiedInspector/TaskDetails.js
@@ -15,39 +15,37 @@ export default class TaskDetails extends React.PureComponent {
     task: object
   };
 
-  constructor(props) {
-    super(props);
+  state = {
+    dependencies: [],
+    dependenciesLoading: false
+  };
 
-    this.state = {
-      dependencies: [],
-      dependenciesLoading: false
-    };
-  }
+  async loadDependencies(taskIds) {
+    this.setState({ dependenciesLoading: true });
 
-  loadDependencies(taskIds) {
-    const dependencies = {};
+    try {
+      const dependencies = await Promise.all(
+        taskIds.map(taskId =>
+          Promise.all([
+            this.props.queue.task(taskId),
+            this.props.queue.status(taskId)
+          ])
+        )
+      );
 
-    this.setState({ dependenciesLoading: true }, async () => {
-      try {
-        await Promise.all(
-          taskIds.map(async taskId => {
-            const [task, { status }] = await Promise.all([
-              this.props.queue.task(taskId),
-              this.props.queue.status(taskId)
-            ]);
-
-            dependencies[taskId] = {
-              task,
-              status
-            };
-          })
-        );
-
-        this.setState({ dependencies, dependenciesLoading: false });
-      } catch (error) {
-        this.setState({ error });
-      }
-    });
+      this.setState({
+        dependenciesLoading: false,
+        dependencies: dependencies.reduce(
+          (deps, [task, { status }]) => ({
+            ...deps,
+            [status.taskId]: { task, status }
+          }),
+          {}
+        )
+      });
+    } catch (error) {
+      this.setState({ error });
+    }
   }
 
   componentWillReceiveProps({ task }) {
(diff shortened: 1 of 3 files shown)


helfi92:
screen shot 2017-11-22 at 1 56 55 pm

src/views/UnifiedInspector/TaskDetails.js
```diff
@@ -130,13 +173,49 @@ export default class TaskDetails extends React.PureComponent {
...
+                              {dependenciesLoading ? (
+                                <BarSpinner />
+                              ) : (
+                                <Label
+                                  bsStyle={
+                                    labels[
+                                      dependencies[dependency].status.state
+                                    ]
+                                  }>
+                                  {dependencies[dependency].status.state}
```
helfi92: @djmitche If I want to get the task state of a dependency, do I need to dig into the `runs` property or is it fine to just look at the `state` property inside `status`.

jonasfj:
I think it should be fine...

jonasfj:
This seems pretty awesome, btw :)

src/views/UnifiedInspector/TaskDetails.js
```diff
@@ -14,8 +15,50 @@ export default class TaskDetails extends React.PureComponent {
...
+    super(props);
+
+    this.state = {
+      dependencies: [],
+      dependenciesLoading: false
+    };
+  }
+
+  loadDependencies(taskIds) {
+    const dependencies = {};
```
eliperelman: You can have less async nesting if you do something like this, which also doesn't need an accumulator object in advance:

```js
loadDependencies(taskIds) {
  this.setState({ dependenciesLoading: true });
  
  try {
    const dependencies = await Promise.all(taskIds.map(taskId => Promise.all([
      this.props.queue.task(taskId),
      this.props.queue.status(taskId)
    ])));

    this.setState({
      dependenciesLoading: false,
      dependencies: dependencies.reduce((deps, [task, { status }]) => ({
        ...deps,
        [status.taskId]: { task, status }
      }), {})
    });
  } catch (error) {
    this.setState({ error });
  }
}
```

src/views/UnifiedInspector/TaskDetails.js
```diff
@@ -14,8 +15,50 @@ export default class TaskDetails extends React.PureComponent {
     task: object
   };
 
+  constructor(props) {
```
eliperelman: For future reference, since we have class properties we don't need a constructor just to initialize state, just like static props:

```js
export default class TaskDetails extends React.PureComponent {
  state = {
    dependencies: [],
    dependenciesLoading: false
  };
```

src/views/UnifiedInspector/TaskDetails.js
```diff
@@ -14,8 +15,50 @@ export default class TaskDetails extends React.PureComponent {
...
+        );
+
+        this.setState({ dependencies, dependenciesLoading: false });
+      } catch (error) {
+        this.setState({ error });
+      }
+    });
+  }
+
+  componentWillReceiveProps({ task }) {
```
eliperelman: Since we added a `componentWillReceiveProps`, this will now hang up every time the taskId changes. You may want to look at how the other sub-components update as a reference for here.

src/views/UnifiedInspector/TaskDetails.js
```diff
@@ -130,13 +173,49 @@ export default class TaskDetails extends React.PureComponent {
...
+                      ? task.dependencies.map((dependency, key) => (
+                          <tr key={`task-dependency-${key}`}>
+                            <td>
+                              <Link to={`/tasks/${dependency}`}>
+                                {dependency}
+                              </Link>
+                            </td>
+                            <td>
+                              {dependenciesLoading ? (
+                                <BarSpinner />
```
eliperelman: This has the potential to show a lot of loading icons. We should block this whole section with a loading icon until `dependenciesLoading` is `false`, then show the whole list when `true`.

src/views/UnifiedInspector/TaskDetails.js
helfi92: I love this refactor. Thanks for the suggestion!

src/views/UnifiedInspector/TaskDetails.js
```diff
@@ -130,13 +200,20 @@ export default class TaskDetails extends React.PureComponent {
...
               <td>Dependencies</td>
               <td>
-                {task.dependencies.length
-                  ? task.dependencies.map((dependency, key) => (
-                      <div key={`task-dependency-${key}`}>
-                        <Link to={`/tasks/${dependency}`}>{dependency}</Link>
-                      </div>
-                    ))
-                  : '-'}
+                {this.state.dependenciesLoading || task.dependencies.length ? (
```
eliperelman: I'm not sure I understand the `or` condition here, could you explain it?

helfi92: I agree it's not really intuitive. I noticed it can just be replaced with `this.state.dependencies.length`. Pushed the change.

PR #394, opened 2017-11-22
Add option to filter for "direct ownership" in scope explorer

jonasfj:
A quick hack to make it easier to find clients that should have a scope removed...

Ie. when we have a scope we want to remove... this is a good way to find those that have the scope directly..

### DIFF
--- a/src/views/ScopeInspector/ScopeInspector.js
+++ b/src/views/ScopeInspector/ScopeInspector.js
@@ -30,7 +30,8 @@ export default class ScopeInspector extends React.PureComponent {
       roles: null,
       clients: null,
       scopeSearchTerm: '',
-      entitySearchMode: 'Has Scope'
+      entitySearchMode: 'Has Scope',
+      directEntitySearch: false
     };
   }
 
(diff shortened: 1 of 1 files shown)


src/views/ScopeInspector/ScopeInspector.js
```diff
@@ -222,6 +229,20 @@ export default class ScopeInspector extends React.PureComponent {
                     />{' '}
                     Has Sub-Scope
                   </MenuItem>
+                  <MenuItem key="match-options-divider" divider />
+                  <MenuItem
+                    key="toggle-match-mode"
+                    onClick={() => this.toggleDirectEntitySearch()}>
```
eliperelman: Since `toggleDirectEntitySearch` is an arrow function, you can pass it directly by reference:

```jsx
<MenuItem
  key="toggle-match-mode"
  onClick={this.toggleDirectEntitySearch}>
```

src/views/ScopeInspector/ScopeInspector.js
```diff
@@ -222,6 +229,20 @@ export default class ScopeInspector extends React.PureComponent {
...
+                    <Glyphicon
+                      glyph="ok"
+                      style={
+                        this.state.directEntitySearch
+                          ? {}
```
eliperelman: Pass `null` here instead of an empty object.

Issue #395, opened 2017-11-22
option to sort clients  by `lastDateUsed`


jonasfj:
This would make it easier to kill old unused clients...

helfi92:
I like this idea. I can take care of that soon.

PR #397, opened 2017-11-23
Add feature to sort clients by last used

helfi92:
Resolves https://github.com/taskcluster/taskcluster-tools/issues/395.

### DIFF
--- a/src/views/ClientManager/ClientManager.js
+++ b/src/views/ClientManager/ClientManager.js
@@ -8,7 +8,6 @@ import {
   Table,
   Checkbox
 } from 'react-bootstrap';
-import moment from 'moment';
 import { Link } from 'react-router-dom';
 import Error from '../../components/Error';
 import Spinner from '../../components/Spinner';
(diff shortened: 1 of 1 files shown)


src/views/ClientManager/ClientManager.js
```diff
@@ -158,8 +163,40 @@ export default class ClientManager extends Component {
...
+    }
+
+    if (clientBLastUsed.isBefore(clientALastUsed)) {
+      return -1;
+    }
+
+    return 0;
+  }
+
+  sortClients = (clientA, clientB) => {
```
eliperelman:
```js
sortClients = (clientA, clientB) => this.state.sortByLastUsed
  ? this.sortByLastUsed(clientA, clientB)
  : clientA.clientId.localeCompare(clientB.clientId);
```

src/views/ClientManager/ClientManager.js
```diff
@@ -158,8 +163,40 @@ export default class ClientManager extends Component {
     );
   };
 
+  handleSortChange = ({ target }) => {
+    this.setState({ sortByLastUsed: target.checked });
+  };
+
+  sortByLastUsed(clientA, clientB) {
+    const clientALastUsed = moment(clientA.lastDateUsed);
```
eliperelman: Is `lastDateUsed` a `Date` object? If so, we can just compare the dates instead of turning every one into a moment object. If they are date strings, they can still be passed to the `Date` constructor and then compared:

```js
sortByLastUsed(clientA, clientB) {
  if (clientA.lastDateUsed > clientBLastUsed) {
    return -1;
  } else if (clientA.lastDateUsed < clientBLastUsed) {
    return 1;
  } else {
    return 0;
  }
```

helfi92: Yep, they are strings. I'll pass them to the `Date` constructor. Thanks!

PR #398, opened 2017-11-24
Snackbar: control the presence of toasts in state

helfi92:


src/components/Snackbar/index.js
```diff
@@ -1,28 +1,45 @@
...
+    ),
+    onDismiss: func.isRequired
+  };
+
   constructor(props) {
     super(props);
+
     this.state = {
-      message: '',
+      toasts: [],
```
djmitche: Maybe the terms are obvious to some, but not to me -- some comments on what a "snackbar" is and how it relates to "toast" (which I think of as more of a breakfastbar item) would be useful.

helfi92: @djmitche. Sorry, I should have provided more information. This pull-request will make it easier to transition the notification component to the [Snackbar component](https://react-md.mlaursen.com/components/snackbars) which we plan on using in our refactor. You may want to see https://github.com/taskcluster/taskcluster-tools/pull/376#discussion_r151431195.

djmitche:
Gotcha, thanks!

src/components/Snackbar/index.js
```diff
@@ -1,28 +1,45 @@
...
-export default class Snackbar extends React.PureComponent {
+export default class SnackBar extends React.PureComponent {
+  static propTypes = {
+    toasts: arrayOf(
+      shape({
+        text: string.isRequired,
+        icon: node
+      })
+    ),
+    onDismiss: func.isRequired
```
eliperelman: Should `onDismiss` really be required?

src/components/Snackbar/index.js
```diff
@@ -1,28 +1,45 @@
...
     };
   }
 
-  componentDidUpdate(prevProps, prevState) {
-    if (!prevState.show && this.state.show) {
-      setTimeout(() => {
-        this.setState({ show: false });
-      }, 2000);
+  componentWillReceiveProps({ toasts }) {
+    const length = toasts && toasts.length;
```
eliperelman: How about importing `deepequal` and doing an is equal on the previous and next toasts?

src/components/Snackbar/index.js
```diff
@@ -1,28 +1,45 @@
...
+
+    if (length && length !== this.state.toasts.length) {
+      this.setState({ toasts, show: true }, () =>
```
eliperelman: If we do async work, we don't really need to wait for the `setState` callback in order to start the next action, since `setState` is also async:

```js
this.setState({ toasts, show: true });
setTimeout(() => {
  this.setState({ show: false });
  this.props.onDismiss();
}, 2000);
```

src/components/Snackbar/index.js
helfi92: If we want to handle things using props, then I believe we need a way to clean toasts that have been shown.

```js
this.setState({ toasts: toasts.concat(...) }); // will display first toast
...
this.setState({ toasts: toasts.concat(...) }); // will display first and second toast 
...
this.setState({ toasts: toasts.concat(...) }); // will display first, second, and third toast

...
render() {
  return (
    ...
    <Snackbar toasts={this.state.toasts} />
    ...
  );
}
```

I believe the react-md's [interactive demo](https://react-md.mlaursen.com/components/snackbars#interactive-demo) uses `onDismiss` as well to clean the state. Let me know if you had something else in mind :)

eliperelman: I understand that most components will use `onDismiss` to unregister a collection of toasts. My concern is that `onDismiss` is required. Does the Snackbar need `onDismiss` to actually remove the shown toasts? It *looks* like that's the case, I just want to make sure.

helfi92: It will technically not need `onDismiss` if the component only uses a single toast. Internally, the Snackbar components renders the first element in `toasts` and then calls onDismiss. To make it work for many toasts, you will need to have an onDismiss, otherwise, you will see the same first toast being pushed. Btw, `onDismiss` is also set as required in react-md.

src/components/Snackbar/index.js
helfi92: We're not really checking for deep equality here. All that is required is that the length is different. Every time the length is different, we use the first element of the array as the notification and then we call `this.props.onDismiss`.

Would you like it if I use the `length` function from ramda instead to make it look nicer and less confusing?
```js
import { length } from 'ramda';
...
if (length(toasts) !== length(this.state.toasts)) {
  this.setState({ toasts, show: true });

  setTimeout(() => {
    this.setState({ show: false });
    this.props.onDismiss();
  }, 2000);
}
```

eliperelman: What I'm worried about is the length being the same, but the toasts being different.

helfi92: I don't see how this could happen if we're implementing `onDismiss` and creating a new array using `concat` or destructuring. At the same time, it doesn't hurt to use `deepEqual` :) I'll change the implementation to check for deep equality.

src/components/Snackbar/index.js
eliperelman: I think the most common use case is going to be a single toast instead of multiple. Thoughts?

helfi92: Yes, the most common use case is a single toast, but you still have the scenario where a user performs two or more actions on a worker, worker-type or provisioner. Also, if we are going to use the react-md Snackbar component in the photon rewrite, we might as well keep it required so that it's easier to transition to that library component.

src/components/Snackbar/index.js
```diff
@@ -1,28 +1,40 @@
...
-    };
-  }
+export default class SnackBar extends React.PureComponent {
+  static propTypes = {
+    toasts: arrayOf(
+      shape({
+        text: string.isRequired,
+        icon: node
+      })
+    ),
```
eliperelman: `toasts` should be required, right?

helfi92: `toasts` has its `text` property required. As a result, omitting toasts will throw saying that it requires at the least a toast object with a text property.

eliperelman:
Thanks for being patient with me. 😸

PR #399, opened 2017-11-24
Integrate App component

helfi92:
This commit attaches links to the drawer element, successfully
allowing each item in the drawer to navigate to their respective page.
The components and views folder have not been touched. The content
of `src/components/` and `src/views` are simply a copy of `src-legacy`.


djmitche:
Sorry, I don't know why I assigned reviewers..

djmitche:
Oh, I see this isn't merging to master, which explains why I'm totally confused :)

helfi92:
The last commit changes the account menu section. I'm thinking we could later add small avatars in the top right linking to recent pages a user has navigated.

Before             |  After
:-------------------------:|:-------------------------:
screen shot 2017-11-27 at 11 01 20 am | screen shot 2017-11-27 at 11 05 13 am



The idea was to make it similar to the material design specs where the account menu looks like that:

screen shot 2017-11-27 at 11 08 40 am

eliperelman:
I had the rest of the components in src-legacy because they still contain react-bootstrap components. I'd like the src directory to only contain views and components that we have migrated to material.

helfi92:
@eliperelman The `src` directory is clean now and I removed the commit that erased `src-legacy`. I agree it's easier to work in a clean `src` directory.

helfi92:
Next, I will work on enhancing the sidebar layout.

helfi92:
Refined the sidebar layout. We can iterate on it or keep this layout.

screen shot 2017-11-30 at 11 03 06 am

djmitche:
Hooks and secrets are what we refer to as "core services", so that might be
a good heading for them.  The index is also a core service, too, and to be
honest should only have one too (instead of the indexed task browser and
the indexed artifact browser).
​

eliperelman:
@helfi92 any idea why many of the files still show as modified?

helfi92:
Ah, didn't realize this happened. I think it has to with file mode permissions changed when I deleted the src-legacy folder and put it back. 

Something like `0 src-legacy/App/LegacyRedirect.js 100644 → 100755` means file was changed from 644=`rw-r--r--` to  755=`rwxr-xr-x`. I'll fix this.

helfi92:
Fixed the file mode weirdness.

PR #405, opened 2017-12-05
Possible fix for taskcluster signin

helfi92:
There seems to be a regression. Clicking on "Development Sign-in" redirects to the create new client editor view and appends the "-1" to the clientId. This behavior is not wanted. The "-1" at the end of the clientId should only happen when a user chooses not to reset the token. It's currently hard to debug since the development sign-in redirect you to the tools production website (no longer on localhost). I'm hoping this would fix the issue but it's hard to know without having it merged first.


helfi92:
OK, that seems to have fixed the issue :)
/cc @djmitche

PR #407, opened 2017-12-05
Update to latest tc-client-web

eliperelman:
Fixes issues with the slug functions being non-functional.


djmitche:
Hm, only yarn.lock changed, not package.json?

eliperelman:
I think recent yarn versions won't update package.json for patch semver matches.

PR #411, opened 2017-12-12
Bug 1422968 - Combine Index Browser and Indexed Artifacts Browser

ydidwania:
Compared Index Browser and Indexed Artifacts Browser and added the Artifact viewing capabilities to Index Browser itself. Indexed Artifacts Browser still exists and will remove it (and make necessary changes) once these set of changes are reviewed and accepted. 
Pinging  @djmitche  for review :smiley:


src/views/IndexBrowser/EntryView.js
```diff
@@ -4,6 +4,7 @@ import { Link } from 'react-router-dom';
 import Spinner from '../../components/Spinner';
 import DateView from '../../components/DateView';
 import JsonInspector from '../../components/JsonInspector';
+import ArtifactView from '../IndexedArtifactBrowser/ArtifactView';
```
ydidwania: Will move `ArtifactView.js` to `src/IndexBrowser/` once the changes are approved.

djmitche:
@ydidwania - we're all gathered in Austin, TX this week for lots of meetings, so we may be a little slow to review.  We will get to it!

eliperelman:
These changes look good to me so far. Let's combine this PR with the changes to Indexed Artifacts.

ydidwania:
@eliperelman , great, which PR is that ?
And should I delete the Indexed Artifacts Browser and push another commit?

eliperelman:
@ydidwania yes sorry. The bug this is for specifically says that these 2 views should be combined into 1, so we need to delete the artifacts browser and only have the index browser.

ydidwania:
I moved [ArtifactsView.js](https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/IndexedArtifactBrowser/ArtifactView.js)  from `src/views/IndexedArtifactBrowser` to `src/views/IndexBrowser` . Should I also merge this ArtifactsView.js into [src/views/IndexedBrowser/EntryView.js](https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/IndexBrowser/EntryView.js) ?

eliperelman:
@ydidwania yep! any functionality that exists within the artifacts browser should make its way into the index browser.

ydidwania:
@eliperelman , I have migrated all the functionaliy, it was just a view tbh. 
Currently the artifacts view component resides in a separate file in `src/views/IndexBrowser` . I was thinking if you wanted me to have 1 single EntryView.js instead of two separate files now.

src/views/IndexBrowser/EntryView.js
```diff
@@ -120,6 +123,27 @@ export default class EntryView extends React.PureComponent {
...
+          <dd>
+            <ArtifactView
+              taskId={this.state.task.taskId}
+              namespace={this.state.task.namespace}
+              queue={queue}
+              userSession={userSession}
+            />
+            <br />
+            <div className="alert alert-info" role="alert">
+              <strong>Latest Artifacts</strong>&nbsp; is the artifacts from the
```
eliperelman: Having a `&nbsp;` and a space right after it is putting a double space in the UI:

screen shot 2018-01-17 at 9 42 51 am

Let's remove the `&nbsp;`.

src/App/index.js
```diff
@@ -271,10 +268,6 @@ export default class App extends React.Component {
                   path="/purge-caches"
                   component={CachePurgeInspector}
                 />
-                <PropsRoute
```
eliperelman: With this route now gone, we will need to add a redirect from this old route to the new one. You can either add a new component to handle this, or piggy-back off the `LegacyRouting` component and add a condition there to handle it:

https://github.com/taskcluster/taskcluster-tools/blob/015baa0da3bf2e7f03cb8d81eaa00575e5da2d17/src/App/LegacyRedirect.js

src/views/IndexBrowser/EntryView.js
ydidwania: Removing `&nbsp;` threw up a linting error 
```
 138:74  error  Insert `·last`  prettier/prettier
 139:14  error  Delete `·last`  prettier/prettier
``` 
I removed the extra space after it instead. I hope thats okay

src/App/index.js
ydidwania: Added  this 
```
<PropsRoute
    path="/index/artifacts/:namespace?/:namespaceTaskId?"
    component={IndexBrowser}
/>
```

src/App/LegacyRedirect.js
```diff
@@ -55,6 +55,10 @@ const LegacyRedirect = props => {
       );
     }
 
+    case '/index/artifacts': {
+      return <Redirect to={`/index/${props.location.pathname.slice(17)}`} />;
```
eliperelman: Using a slice of this index is fragile here, we can just replace it:

```jsx
return <Redirect to={props.location.pathname.replace('/artifacts', '')} />;
```

djmitche:
A few notes for further improvement, after this PR lands:

* On the left side, distinguish tasks from namespaces somehow.  At the moment, there's a list of tasks, a little whitespace, and then a list of namespaces.  Maybe by using a hierarchy of "twirldown" arrows like we do in the hooks manager?
* When clicking on a namespace that does not have a task, there's an alert that says "Task not found!", but this expected, so that shouldn't appear.

eliperelman:
Thank you for your hard work on this!

ydidwania:
Thank you too Eli for your guidance!

PR #412, opened 2017-12-13
Add collaborators settings for tc-gh quickstart

imbstack:
People were asking about this.


aneeshusa:
As one of the aformentioned people, thanks for adding this!

PR #413, opened 2017-12-18
Set default GitHub events for Quick Start tool

aneeshusa:
This ensures .taskcluster.yml files generated with the quick start tool
actually cause Taskcluster jobs to be triggered,
instead of seeming to not do anything.

Because the current job discoverability story is not great
(best way to find jobs is via GitHub status checks),
this should help reduce confusion for newcomers to TaskCluster.


aneeshusa:
This PR should help others avoid the confusion I faced in https://github.com/servo/servo/pull/19591#issuecomment-352293192 while figuring out how to enable TaskCluster for Servo.

PR #414, opened 2017-12-18
Avoid automatically changing working dir, git state

aneeshusa:
Having a pre-commit lint hook is great,
but the current setup that automatically changes working directory
files and does a raw git add was surprising to me.

Additionally, this is incompatible with any workflow that uses the git
staging area, which I tend to do a lot.

I was surprised that the hooks for my local git repo were mutated by
husky (via yarn), but using this for linting seems OK for now.
In the future, I would prefer to have a one-line command to
enable/install hooks instead of having them automatically configured.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,7 @@
   },
   "lint-staged": {
     "*.js": [
-      "neutrino lint --fix",
-      "git add"
+      "neutrino lint"
     ]
   },
   "repository": "taskcluster/taskcluster-tools",


aneeshusa:
The current behavior confused me at first appearance and impeded my workflow while working on #413. Hopefully the proposed change keeps most of the benefits while being less surprising for new contributors like me.

djmitche:
+1 to this -- I keep trying to disable the hooks but I haven't figure out how to keep them disabled.

eliperelman:
The point of this setup is to prevent fixable linting errors from being committed to a PR, and `lint-staged` fixes these error prior to the commit, and rejecting the commit if linting fails. Disabling this means having to re-commit linting fixes. After finding out travis failed.

I could go either way on this, but I prefer not having linting fixes guarded before ever having reached travis.

Thoughts @helfi92?

aneeshusa:
I've kept the `neutrino lint` bit, so husky will still prevent you from committing if lint errors are found. What I've changed is to have it not try to automatically fix + commit any lint errors for you.
This way errors are still caught locally before commiting, which I agree is great for tightening the feedback loop.

eliperelman:
> have it not try to automatically fix + commit any lint errors for you

Can you elaborate on why this is a problem for you?

helfi92:
I think the current implementation is fine. I rather have my bad linting errors fixed and added to my commit automatically instead of running an extra command to fix linting errors.

aneeshusa:
In terms of automatically fixing issues, I was surprised to see `git commit` mutate working directory files. 

The main issue I have with the current behavior is that it silently breaks using git's [staging area](https://git-scm.com/about/staging-area). When working on a patchset, I will often make a bunch of local changes, then use `git add -p` to add and commit individual portions to provide a cleaner commit history. This makes reviews and later refactors or rebases easier; it also makes it easy to separate refactors from behavior changes. However, the current behavior of `git add` will add all local changes, even those I didn't want to commit just yet, and is almost silent about this.

@helfi92 I would suggest configuring your editor to autofix lint issues when saving, which would be even faster.
In general I agree with automating mundane things, but the surprise factor was a bit high for me in this case. Maybe make the fix/add bit opt-in?

helfi92:
You do raise a valid point regarding `git add -p`. I can see how this can be a problem.

eliperelman:
For reference: https://github.com/okonet/lint-staged/issues/62

djmitche:
I realize neutrino is opinionated, but this is *really* aggressive and doesn't match my workflow either (which often involves a lot of rebasing).

I'd really like this to be opt-in :)

eliperelman:
I don't want to get off on a tangent, but wanted to make it clear that: Neutrino isn't very opinionated, it's our configuration that is, and this issu has nothing to do with Neutrino, but rather our settings for lint-staged.

Im considering making this opt-in, but I would first like to determine if there is a compromise we can do instead.

PR #418, opened 2018-01-03
Fix issue where dependencies aren't shown

helfi92:
Navigating directly to the details page of a task ID properly shows the
list of dependencies in the UI. However, clicking on one of
those tasks then navigating to the details page may show an empty list
of dependencies.

Presently, the logic behind loading the dependencies sits under
`componentWillReceiveProps`. This works when you load a URL directly
but when you navigate to it from the "Run Logs" tab for example,
`componentWillReceiveProps` will not trigger since we are dealing
with the same task. To solve that issue, when we have the task,
we can take care of the loading part in `componentWillMount`.

### DIFF
--- a/src/views/UnifiedInspector/TaskDetails.js
+++ b/src/views/UnifiedInspector/TaskDetails.js
@@ -49,6 +49,12 @@ export default class TaskDetails extends React.PureComponent {
     }
   }
 
+  componentWillMount() {
+    if (this.props.task && this.props.task.dependencies.length) {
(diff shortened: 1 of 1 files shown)


helfi92:
@djmitche It won't. When you load the details page from the URL, `componentWillMount` will not have `this.props.task` set to anything, hence will not pass the if condition. It will only load the dependencies when the task has been set. On the other hand, when a user navigates to the details page from a different tab, `componentWillReceiveProps` will not be called since the props of that component are not changing.

PR #423, opened 2018-01-17
Bug 1430895 - upgrade hook triggering to support context

djmitche:
This is the frontend for the new parameterized-hooks support @alexandrasp landed today..

### DIFF
--- a/src/views/HooksManager/TriggerButton.js
+++ b/src/views/HooksManager/TriggerButton.js
@@ -1,4 +1,6 @@
 import React from 'react';
+import Ajv from 'ajv';
+import jsonSchemaDefaults from 'json-schema-defaults';
 import { string, object, func } from 'prop-types';
 import { Row, Col, Alert } from 'react-bootstrap';
 import { safeLoad, safeDump } from 'js-yaml';
@@ -16,10 +18,19 @@ export default class TriggerButton extends React.PureComponent {
 
   constructor(props) {
     super(props);
+    this.ajv = new Ajv({ format: 'full', verbose: true, allErrors: true });
+    this.validate = this.ajv.compile(props.schema);
     this.state = {
       context: {},
-      contextValid: true
+      contextValid: true,
+      validityMessage: undefined,
+      initialContext: safeDump(jsonSchemaDefaults(props.schema) || {})
     };
+  }
+
+  componentWillMount() {
+    // initialize context validation
+    this.handleContextChange(this.state.initialContext);
   }
 
   handleSubmit = () => {
@@ -43,6 +54,16 @@ export default class TriggerButton extends React.PureComponent {
       });
     }
 
+    const valid = this.validate(context);
+
+    if (!valid) {
+      return this.setState({
+        context: null,
+        contextValid: false,
+        validityMessage: this.ajv.errorsText(this.validate.errors)
+      });
+    }
+
     this.setState({
       context,
       contextValid: true,
@@ -52,7 +73,7 @@ export default class TriggerButton extends React.PureComponent {
 
   render() {
     const { hookGroupId, hookId, schema } = this.props;
-    const { validityMessage } = this.state;
+    const { initialContext, validityMessage } = this.state;
     const triggerModal = (
       <div>
         Trigger Hook{' '}
@@ -66,7 +87,7 @@ export default class TriggerButton extends React.PureComponent {
             <CodeEditor
               mode="yaml"
               lint={true}
-              value={'{}'}
+              value={initialContext}
               onChange={this.handleContextChange}
             />
             {validityMessage && (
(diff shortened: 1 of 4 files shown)


src/views/HooksManager/TriggerButton.js
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+    this.ajv = new Ajv({ format: 'full', verbose: true, allErrors: true });
+    this.validate = this.ajv.compile(props.schema);
+    this.state = {
+      context: {},
+      contextValid: true,
+      validityMessage: undefined,
```
eliperelman: For state with empty values, we want to initialize with `null` instead of `undefined`. That way when they get reset to `null`, React will overwrite the value instead of skipping it.

src/views/HooksManager/TriggerButton.js
```diff
@@ -0,0 +1,119 @@
...
+import Ajv from 'ajv';
+import jsonSchemaDefaults from 'json-schema-defaults';
+import { string, object, func } from 'prop-types';
+import { Row, Col, Alert } from 'react-bootstrap';
+import { safeLoad, safeDump } from 'js-yaml';
+import Code from '../../components/Code';
+import CodeEditor from '../../components/CodeEditor';
+import ModalItem from '../../components/ModalItem';
+
+export default class TriggerButton extends React.PureComponent {
```
eliperelman: To be safe, this should probably be `Component` instead of `PureComponent` since `schema` may change deeply.

src/views/HooksManager/TriggerButton.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  componentWillMount() {
+    // initialize context validation
+    this.handleContextChange(this.state.initialContext);
+  }
+
+  handleSubmit = () => {
+    if (!this.state.contextValid) {
+      throw new Error('Trigger context is not valid');
```
eliperelman: We need to handle this error somewhere, maybe through `state`:

```js
this.setState({
  error: new Error('Trigger context is not valid')
});
```

src/views/HooksManager/TriggerButton.js
```diff
@@ -0,0 +1,119 @@
...
+    } else {
+      this.props.onTrigger(this.state.context);
+    }
+  };
+
+  handleContextChange = value => {
+    let context;
```
eliperelman: Petty, but this can be re-written for `const` instead of using `let` just to break out of a scope. It also saves some extra lines:

```js
try {
  const context = safeLoad(value);
  const valid = this.validate(context);

  this.setState({
    context: valid ? context : null,
    contextValid: valid,
    validityMessage: valid ? null : this.ajv.errorsTest(this.validate.errors)
  });
} catch (err) {
  this.setState({
    context: null,
    contextValid: false,
    validityMessage: 'Trigger context is not valid YAML'
  });
}
```

src/views/HooksManager/TriggerButton.js
djmitche: I'd really rather schema *couldn't* change, for UI reasons -- what would the user do in that case? But I can just change PureComponent to Component and ignore that issue :)

src/views/HooksManager/TriggerButton.js
djmitche: The ModalItem class handles it and displays it within the modal.  But I do need to change that error message to use validityMessage..

djmitche: Also, the button is disabled in this case, so this error "shouldn't happen".

src/views/HooksManager/TriggerButton.js
djmitche: I find the ordering of your approach to be less readable (it mixes what are really two distinct steps).  But now I'm just being petty too :)

src/views/HooksManager/TriggerButton.js
eliperelman: I'll just say that if the schema can't change, it's the responsibility of the parent component to not pass a new value to the child, and not this child component's responsibility.

src/views/HooksManager/TriggerButton.js
eliperelman: Haha, it's a tradeoff, and I'm fine with either (using an `if/else` and explicit vs. ternaries and succinct). Whatever you like here I'm fine with.

src/views/HooksManager/TriggerButton.js
```diff
@@ -0,0 +1,108 @@
...
+    hookId: string.isRequired,
+    hookGroupId: string.isRequired,
+    schema: object.isRequired,
+    onTrigger: func.isRequired
+  };
+
+  constructor(props) {
+    super(props);
+    this.ajv = new Ajv({ format: 'full', verbose: true, allErrors: true });
+    this.validate = this.ajv.compile(props.schema);
```
eliperelman: As mentioned before, if `props.schema` ever changes, this validate method will be out of date.

PR #424, opened 2018-01-19
Rendering worker type pending tasks differently depending on error or no value

eliperelman:
I randomly faked the errors from some responses so you could see how it would look.

Loading stats (ignore the error stars, they will all have dashes):

screen shot 2018-01-19 at 3 39 23 pm

Stats with errors:

screen shot 2018-01-19 at 3 39 32 pm

### DIFF
--- a/src/views/AwsProvisioner/WorkerTypeRow.js
+++ b/src/views/AwsProvisioner/WorkerTypeRow.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { string, shape, number, bool, func } from 'prop-types';
 import { OverlayTrigger, ProgressBar, Tooltip } from 'react-bootstrap';
+import Icon from 'react-fontawesome';
 
 export default class WorkerTypeRow extends React.Component {
   static propTypes = {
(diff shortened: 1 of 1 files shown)


src/views/AwsProvisioner/WorkerTypeRow.js
```diff
@@ -154,15 +154,17 @@ export default class WorkerTypeRow extends React.Component {
   }
 
   renderState() {
-    if (this.state.error) {
-      return <span />;
+    const { error, pendingTasks } = this.state;
+
+    if (error) {
+      return <span style={{ color: '#f00' }}>*</span>;
```
jhford: Any chance we could make this more descriptive?  `ERROR` would be a lot more likely to trigger investigation than `*` i suspect.  R+ either way.

eliperelman: Sure, I could switch to an error icon with the text in a tool tip.

eliperelman:
Switched to icon:

screen shot 2018-01-22 at 9 15 38 am

Issue #426, opened 2018-01-22
Terminate instances buttons need to use the correct provisioner id


jhford:
In the staging instance of the provisioner UI, all the links for the Terminate buttons use the wrong provisioner base url

`https://tools.taskcluster.net/aws-provisioner-staging/gecko-1-b-linux/resources`

calls

`DELETE https://ec2-manager.taskcluster.net/v1/worker-types/gecko-1-b-linux/resources`

When it should be calling

`DELETE https://ec2-manager-staging.taskcluster.net/v1/worker-types/gecko-1-b-linux/resources`

I'm not entirely sure where this string is coming from, so it might not be in this repository, but the end result here is that this repository is surfacing the issue.  Please refile in a new repository if more appropriate.

cc @helfi92 @eliperelman

PR #427, opened 2018-01-22
Use correct base url in the staging instance of provisioner

helfi92:
Fixes https://github.com/taskcluster/taskcluster-tools/issues/426.


eliperelman:
@jhford this should now be live.

PR #430, opened 2018-01-30
Bug 1427821 - only save when valid

djmitche:


src/views/HooksManager/HookEditor.js
```diff
@@ -234,6 +233,7 @@ export default class HookEditor extends React.PureComponent {
           disabled={!this.validHook()}>
           <Glyphicon glyph="ok" /> Save Changes
         </Button>
+        valid = {JSON.stringify(this.validHook())}
```
eliperelman: Dev artifact?

djmitche: Fixed

djmitche:
..and lint fixed too, sorry.

PR #432, opened 2018-01-30
Pool session credentials across tabs

eliperelman:
@djmitche wanted to get this open for a sanity check. I'm not super-happy with the outcome, but it's a start.

### DIFF
--- a/src/auth/UserSession.js
+++ b/src/auth/UserSession.js
@@ -1,5 +1,7 @@
 import { OIDCCredentialAgent } from 'taskcluster-client-web';
 
+const SESSION_KEY = 'taskcluster-client-pool-credentials';
+
 /**
  * An object representing a user session.  Tools supports a variety of login methods,
  * so this combines them all in a single representation.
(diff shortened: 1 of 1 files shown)


src/auth/UserSession.js
```diff
@@ -82,10 +84,44 @@ export default class UserSession {
...
+      localStorage.setItem(SESSION_KEY, 'pending');
+
+      const credentials = await this.credentialAgent.getCredentials({});
+
+      localStorage.setItem(SESSION_KEY, JSON.stringify(credentials));
+
+      return credentials;
+    } else if (sessionCredentials === 'pending') {
+      return new Promise(resolve => {
+        window.addEventListener('storage', function handler({
```
djmitche: Is there a race here?

* localStorage.getItem() -> 'pending'
* another tab updates, event sent
* this tab adds event listener
* promise never resolves

djmitche: Maybe that could be fixed by just leaving the event listener installed all the time..

eliperelman: I was thinking the same thing; it's what I was afraid of. I need to do some more thinking on this.

Issue #278, opened 2017-08-12
tool to expand scope or set of scopes


jonasfj:
Say I have a role like: `assume:repo:github.com/glandium/git-cinnabar:branch:master` and I want to know what scopes it gets... I need to call `auth.expandScopes` it would be nice to have a tool for this...

Note: `assume:repo:github.com/glandium/git-cinnabar:branch:<branch>` is basically a pattern made up by tc-gh...

jonasfj:
This is just a nice-to-have quick hack.. Nothing important :)

djmitche:
The pattern is actually documented as a Taskcluster convention in the namespaces document, and is also supported by mozilla-taskcluster.

jonasfj:
I probably have no excuse for being too lazy to lookup the right docs...

But yes, namespaces document is where I should look.. still it's nice to be able to expand scopes.

djmitche:
Yes, I was only disagreeing with your second paragraph.  I agree with the first :)

Biboswan:
Can I work on this.

djmitche:
Yes!  I think this would be an additional view which can take a list of scopes and expand it.

Biboswan:
I 'm trying ```` await this.props.auth.expandScopes({ scopes: this.state.scopes }) ```` where scopes is an array

Getting error:
````
TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.
    at s (VM42094 290:2)
    at setTimeout (VM42094 290:2
````
https://docs.taskcluster.net/reference/platform/taskcluster-auth/references/api#expand-scopes

djmitche:
Uh, wow.  So that API call isn't valid HTTP.  That was really stupid of me.

djmitche:
So it sounds like you're on the right track!  Once the change linked above lands, we'll need to upgrade taskcluster-client-web in tools, and then the call you're making should work properly.

Biboswan:
I see only taskcluster-client-web/src/clients/Auth.js needs an update.

Biboswan:
Anything else?

djmitche:
https://github.com/taskcluster/taskcluster-client-web#updating-built-in-apis shows how to update the client and make a PR.  I'll try to get to that today, but if you beat me to it and make a PR that's great too :)

djmitche:
OK, if you `yarn add taskcluster-client-web@5.1.0` you should get a working
expandScopes.

[...]

PR #433, opened 2018-01-31
Expandscopes tool

Biboswan:
#278


djmitche:
Looks good to me.  I'll defer to Eli for any detailed issues.

package.json
```diff
@@ -75,7 +75,7 @@
     "react-vnc-display": "^1.1.0",
     "shebang-loader": "0.0.1",
     "slugid": "^1.1.0",
-    "taskcluster-client-web": "^5.0.0",
+    "taskcluster-client-web": "5.1.0",
```
eliperelman: Hmm, looks like we lost the caret range on this dependency.

src/App/index.js
```diff
@@ -263,6 +266,10 @@ export default class App extends React.Component {
                   path="/auth/grants/:pattern?"
                   component={ScopeGrants}
                 />
+                <PropsRoute
+                  path="/auth/scopeset-expand"
```
eliperelman: Since we already have a `/auth/scopes` route, we could probably make this `/auth/scopes/expansions`.

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import { ButtonToolbar, Button, Glyphicon, Row, Col } from 'react-bootstrap';
+import equal from 'deep-equal';
+import HelmetTitle from '../../components/HelmetTitle';
+import ScopeEditor from '../../components/ScopeEditor';
+
+export default class ScopesetExpander extends React.PureComponent {
+  constructor(props) {
```
eliperelman: Since nothing here is relying on props, we can make state a class field:

```js
export default class ScopesetExpander extends React.PureComponent {
  state = {
    scopes: [],
    expandedScopes: null,
    error: null
  }
```

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,91 @@
...
+    super(props);
+
+    this.state = {
+      scopes: [],
+      expandedscopes: null,
```
eliperelman: Use camel case for joined works, e.g. `expandedScopes`.

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,91 @@
...
+  scopesUpdated = scopes => {
+    this.setState({ scopes });
+  };
+
+  renderExpandedScopes() {
+    const { expandedscopes } = this.state;
+
+    return (
+      <div>
+        {expandedscopes ? (
```
eliperelman: Looks like we render an empty `div` here, even when there are no expanded scopes. We can early return `null`:

```jsx
const { expandedScopes } = this.state;

if (!expandedScopes) {
  return null;
}
```

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,91 @@
...
+    );
+  }
+
+  cleanScopesetInput = () => {
+    this.setState({ scopes: [] });
+  };
+
+  fetchExpandedScopes = async () => {
+    try {
+      const expandedscopes = await this.props.auth.expandScopes({
```
eliperelman: You can save a variable declaration here by inlining the fetch:

```js
try {
  this.setState({
    expandedScopes: await this.props.auth.expandScopes({
      scopes: this.state.scopes
    }),
    error: null
  });
} catch (err) {
```

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  };
+
+  render() {
+    return (
+      <Row style={{ marginTop: 10 }}>
+        <HelmetTitle title="Scopeset Expander" />
+        <Col md={6}>
+          <ButtonToolbar style={{ marginBottom: 7 }}>
+            <Button bsStyle="info" onClick={this.fetchExpandedScopes}>
```
eliperelman: The idiomatic way of naming event props and handlers is: `on{Event}={this.handle{Event}}`. Let's rename these events:

```jsx
onClick={this.handleClickFetchExpanded}

onClick={this.handleClickCleanScopes}
```

As a follow-up, I need to go back and get this ESLint rule enabled.

Biboswan:
@eliperelman  I think I forgot to ping you that I have made the changes. :)

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,85 @@
...
+  };
+
+  render() {
+    return (
+      <Row style={{ marginTop: 10 }}>
+        <HelmetTitle title="Scopeset Expander" />
+        <Col md={6}>
+          <ButtonToolbar style={{ marginBottom: 7 }}>
+            <Button bsStyle="info" onClick={this.handleClickFetchExpanded}>
+              <Glyphicon glyph="plus" />Expand Scopes
```
eliperelman: Put a space between the icon and the text:

screen shot 2018-02-05 at 11 46 35 am

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,85 @@
...
+            </Button>
+            <Button bsStyle="warning" onClick={this.handleClickCleanScopes}>
+              <Glyphicon glyph="remove" />Clean
```
eliperelman: Put a space between the icon and the text:

screen shot 2018-02-05 at 11 47 06 am

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+  };
+
+  render() {
+    return (
+      <Row style={{ marginTop: 10 }}>
+        <HelmetTitle title="Scopeset Expander" />
+        <Col md={6}>
+          <ButtonToolbar style={{ marginBottom: 7 }}>
+            <Button bsStyle="info" onClick={this.handleClickFetchExpanded}>
```
eliperelman: When I click this button with nothing entered in the scopes field, I see a heading pop in with nothing there. We should either give them a message on the right saying something about there not being any scopes, or prevent the button from being clicked until something valid is in the field, preferably the latter.

screen shot 2018-02-05 at 11 45 49 am

Biboswan:
take an example like assume:repo:hg.mozilla.org/integration/mozilla-inbound:cron:*
You can test more from https://tools.taskcluster.net/auth/roles/ and clicking on any roleId, its scopes and  expanded scopes appear. 
Note here, however the role id gets also displayed under the expanded scopes.

src/views/ScopesetExpander/ScopesetExpander.js
Biboswan: Scope editor component was modified to check for valid scopes. I need to use some functionality of codemirror.If somehow I can access return value of **CodeMirror.registerHelper** maybe that will do. Or may be something like **doc.getMode()**...  
Any suggestion. Abit confused with codemirror docs

Biboswan:
hopefully things seem better now.

src/components/ScopeEditor/index.js
```diff
@@ -61,13 +61,18 @@ export default class ScopeEditor extends React.Component {
...
+  }
+
   /** Render scopes and associated editor */
   renderScopeEditor() {
     const { scopeText } = this.state;
 
     return (
       <div>
         <CodeEditor
+          ref={editorRef => this.captureCodeEditor(editorRef)}
```
eliperelman: If you switch `captureCodeEditor` to an arrow function, then you can pass it by reference here:

```jsx
captureCodeEditor = (ref) => {
  this.codeEditor = ref;
};

ref={this.captureCodeEditor}
```

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -20,6 +21,16 @@ export default class ScopesetExpander extends React.PureComponent {
     }
   }
 
+  componentDidUpdate() {
+    if (this.scopeEditor) {
+      this.setState({
```
eliperelman: We need to guard `this.setState` from constantly causing this component to re-render by ensuring that the validity hasn't changed:

```js
componentDidUpdate(prevProps, { scopeIsValid }) {
  if (!this.scopeEditor) {
    return;
  }

  const isValid = this.scopeEditor.codeEditor.codeMirror.doc.cm.state.lint.marked.length === 0;
  
  if (isValid !== scopeIsValid) {
    this.setState({ scopeIsValid: isValid });
  }
}
```

src/views/ScopesetExpander/ScopesetExpander.js
```diff
@@ -65,14 +76,20 @@ export default class ScopesetExpander extends React.PureComponent {
...
+                this.state.scopes.length === 0 || !this.state.scopesIsValid
+              }>
               <Glyphicon glyph="plus" /> Expand Scopes
             </Button>
             <Button bsStyle="warning" onClick={this.handleClickCleanScopes}>
               <Glyphicon glyph="remove" /> Clean
             </Button>
           </ButtonToolbar>
           <ScopeEditor
+            ref={editorref => (this.scopeEditor = editorref)}
```
eliperelman: We can move this into a class property just like we do in `ScopeEditor`.

```jsx
registerEditor = (ref) => {
  this.scopeEditor = ref;
};

<ScopeEditor
  ref={this.registerEditor}
```

PR #435, opened 2018-02-02
Update scope editor to warn you when you are entering invalid scopes

imbstack:
I'm not sure that I did any of this in an optimal manner but it does seem to work locally.

cc @djmitche / @jonasfj for inspiration and for probable tweaks.

screen shot 2018-02-01 at 5 32 42 pm
screen shot 2018-02-01 at 5 32 48 pm
screen shot 2018-02-01 at 5 31 51 pm


src/components/ScopeEditor/index.js
```diff
@@ -14,63 +19,68 @@ export default class ScopeEditor extends React.Component {
...
-
-    const scopes = textarea.value
-      .split(/[\r\n]+/)
-      .map(s => s.trim())
-      .filter(s => s !== '');
-
-    return Array.from(new Set(scopes)).sort();
-  };
+  parseScopes(scopeText) {
+    return Array.from(
```
eliperelman: You can also use array spread if you fancy: `[...new Set`

src/components/ScopeEditor/index.js
```diff
@@ -14,63 +19,68 @@ export default class ScopeEditor extends React.Component {
...
+      new Set(
+        scopeText
+          .split(/[\r\n]+/)
+          .map(s => s.trim())
+          .filter(s => s !== '')
```
eliperelman: For consistency with where we do this elsewhere, just filter on truthy values:


```js
.filter(Boolean)
```

src/components/ScopeEditor/index.js
```diff
@@ -14,63 +19,68 @@ export default class ScopeEditor extends React.Component {
...
+          placeholder="new-scope:for-something:*"
         />
       </div>
     );
   }
 
   /** Render a list of scopes */
   renderScopes() {
-    const scopes = clone(this.props.scopes || []).sort();
+    const scopes = this.props.scopes || [];
```
eliperelman: This also doesn't need a default.

src/components/ScopeEditor/scopemode.js
```diff
@@ -0,0 +1,34 @@
+import CodeMirror from 'codemirror';
+import 'codemirror/addon/lint/lint';
+
+CodeMirror.registerHelper('lint', 'scopemode', text =>
```
eliperelman: I hate relying on imports for side effects, but haven't thought of a better alternative to this yet.

imbstack: Agreed. Apparently someone has created a react-codemirror2 project that is being actively maintained and might have a better story for this sort of thing? We can look into that soon. It was a bit too much work for right now to switch it out.

src/components/CodeEditor/index.js
```diff
@@ -29,6 +29,7 @@ export default class CodeEditor extends React.PureComponent {
 
   register(ref) {
     this.codeMirror = ref ? ref.getCodeMirror() : null;
+    this.codeMirrorInstance = ref ? ref.getCodeMirrorInstance() : null;
```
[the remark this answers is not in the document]
imbstack: This is some strange intricacy of react-codemirror or codemirror itself. I'm not 100% sure but they definitely are different. Neither one has both of the sets of functionality we need.

src/components/ScopeEditor/scopemode.js
eliperelman: > Apparently someone has created a react-codemirror2

Yep! I planned to switch to it eventually, but this is fine for now.

src/components/CodeEditor/index.js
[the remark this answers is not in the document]
eliperelman: That's unfortunate. Maybe this will go away with codemirror2 in the future.

eliperelman:
Feel free to merge after you've made your changes.

PR #438, opened 2018-02-07
Add better UX for the Secrets view

helfi92:
2 Features:

- Use `hanson` to parse json. Writing a long secret and
forgeting to double quote a property will throw an error and make you
lose access to the secret you just wrote. Being able to have a library parse
unquoted property names would be awesome.

- Fixes [Bug 1389530 - saving a secret doesn't refresh the view](https://bugzilla.mozilla.org/show_bug.cgi?id=1389530)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "font-awesome": "^4.7.0",
     "font-awesome-svg-png": "^1.1.5",
     "fscreen": "^1.0.2",
-    "hanson": "^1.2.0",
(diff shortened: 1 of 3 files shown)


djmitche:
Let's use YAML instead of HanSON.  We do that other places, such as in the task creator.

djmitche:
Also, if the input is invalid JSON (or YAML) then the submit button should be dimmed out..

PR #439, opened 2018-02-09
Bug 1437209 - GitHub QuickStart is not populating the YAML file for Rust

cynthiapereira:


eliperelman:
Thank you for the patch!

PR #442, opened 2018-02-14
First attempt at getting netlify to run tools

eliperelman:


netlify.toml
```diff
@@ -0,0 +1,42 @@
...
+  status = 200
+  force = false
+
+[[headers]]
+  for = "/index.html"
+  [headers.values]
+    Cache-Control = "no-store, no-cache, must-revalidate"
+    Content-Security-Policy = "frame-ancestors 'self'"
+    Referrer-Policy = "origin"
+    Strict-Transport-Security = "max-age=31536000; includeSubDomains; always;"
```
edmorley: `always` isn't a valid `Strict-Transport-Security` directive; I'm guessing it came from a config that used the nginx `always` directive ([here](https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header))

eliperelman: Haha good catch.

eliperelman:
Will follow up elsewhere.

PR #444, opened 2018-02-15
Bug 1385307 - Show artifact lock icon for logged users

cynthiapereira:
This will replace the `getIconFromMime()` function with the `lock icon`. 

Referred: [Bug 1385307](https://bugzilla.mozilla.org/show_bug.cgi?id=1385307)

### DIFF
--- a/src/components/ArtifactList/index.js
+++ b/src/components/ArtifactList/index.js
@@ -89,6 +89,7 @@ export default class ArtifactList extends React.PureComponent {
         if (userSession) {
           return {
             name,
+            locked: true,
             icon: getIconFromMime(contentType),
             url: isNil(runId)
               ? await queue.buildSignedUrl(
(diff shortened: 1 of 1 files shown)


djmitche:
That's the 'one line' I thought it would be :)

cynthiapereira:
I would love to discuss how to improve this and show both cases :)

djmitche:
@eliperelman there was some discussion in the bug about making this more useful for the user -- what do you think?

cynthiapereira:
@eliperelman I updated the PR.

I rollback to display the mime type for private artifacts again and added the lock icon.
Here it will be:
screen shot 2018-02-22 at 16 47 00

eliperelman:
You could add a `locked` property somewhere around here:

https://github.com/taskcluster/taskcluster-tools/pull/444/files#diff-a0dae82bc49a92f8161e2c44ef25b71aR91

cynthiapereira:
You're right about an artifact does not have a `userSession` property, sorry!

So, I made the requested changes, added a `locked: true` property inside the `userSession` conditional for a signed URL, removed the unnecessary part and then take advantage of `map` to render the locked icon.

PR #446, opened 2018-02-15
Error tracking with rollbar

eliperelman:
I already have the env vars set in Heroku.

### DIFF
--- a/src/App/index.js
+++ b/src/App/index.js
@@ -3,6 +3,7 @@ import { BrowserRouter, Route, Switch } from 'react-router-dom';
 import { object } from 'prop-types';
 import { Helmet, link } from 'react-helmet';
 import { Grid } from 'react-bootstrap';
+import * as Rollbar from 'rollbar';
 import PropsRoute from '../components/PropsRoute';
 import Navigation from '../components/Navigation';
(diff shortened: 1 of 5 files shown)


src/App/index.js
```diff
@@ -110,13 +111,27 @@ export default class App extends React.Component {
...
+    });
 
     this.state = {
       userSession: null,
       authReady: false
     };
   }
 
+  componentDidCatch(err) {
+    if (this.rollbar) {
```
djmitche: why the conditional if this is set unconditionally in the constructor?

src/auth/AuthController.js
```diff
@@ -114,10 +114,8 @@ export default class AuthController {
       }
     } catch (err) {
       this.setUserSession(null);
-      // usually this is just "user interaction required" or the like, but just
-      // in case the user wants to peek, put it in the console
-      /* eslint-disable no-console */
-      console.error('Could not renew login:', err);
+      // throw the error so it can be picked up by rollbar
```
djmitche: I don't think we want this particular error handled this way.  In most cases, it is expected.

eliperelman:
You're right, let me look into Sentry.

PR #450, opened 2018-02-18
remove multiple spinner

fiennyangeln:
Bug 1438852

### DIFF
--- a/src/views/HooksManager/HookBrowser.js
+++ b/src/views/HooksManager/HookBrowser.js
@@ -3,7 +3,6 @@ import { string, func, object } from 'prop-types';
 import TreeView from 'react-treeview';
 import 'react-treeview/react-treeview.css';
 import Error from '../../components/Error';
-import Spinner from '../../components/Spinner';
 
 export default class HookBrowser extends Component {
   static propTypes = {
@@ -28,9 +27,9 @@ export default class HookBrowser extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    // if (nextProps.group !== this.props.group) {
-    this.loadHooksList(nextProps);
-    // }
+    if (nextProps.group !== this.props.group) {
+      this.loadHooksList(nextProps);
+    }
   }
 
   async loadHooksList({ hooks, group, onLoadHooksList }) {
@@ -58,15 +57,15 @@ export default class HookBrowser extends Component {
   handleClick = () => this.loadHooksList(this.props);
 
   render() {
-    const { group, hookGroupId } = this.props;
+    const { group, hookGroupId, showList } = this.props;
     const { hooksList, error } = this.state;
 
     if (error) {
       return <Error error={error} />;
     }
 
-    if (!hooksList) {
-      return <Spinner />;
+    if (!showList || !hooksList) {
+      return null;
     }
 
     // For nodeLabel: forward clicks on the label on to the TreeView's handleClick (undocumented..)
--- a/src/views/HooksManager/HookBrowser.js
+++ b/src/views/HooksManager/HookBrowser.js
@@ -33,7 +33,7 @@ export default class HookBrowser extends Component {
     // }
   }
 
-  async loadHooksList({ hooks, group, callbackChildLoaded }) {
+  async loadHooksList({ hooks, group, onLoadHooksList }) {
     const { hooks: hooksList } = await hooks.listHooks(group);
 
     try {
@@ -48,7 +48,7 @@ export default class HookBrowser extends Component {
       });
     }
 
-    callbackChildLoaded(group);
+    onLoadHooksList(group);
   }
 
   registerChild = ref => (this.treeView = ref);
--- a/src/views/HooksManager/HookBrowser.js
+++ b/src/views/HooksManager/HookBrowser.js
@@ -33,7 +33,7 @@ export default class HookBrowser extends Component {
     // }
   }
 
-  async loadHooksList({ hooks, group }) {
+  async loadHooksList({ hooks, group, callbackChildLoaded }) {
     const { hooks: hooksList } = await hooks.listHooks(group);
 
     try {
@@ -47,6 +47,8 @@ export default class HookBrowser extends Component {
         error: err
       });
     }
+
+    callbackChildLoaded(group);
   }
 
   registerChild = ref => (this.treeView = ref);
--- a/src/views/HooksManager/HookBrowser.js
+++ b/src/views/HooksManager/HookBrowser.js
@@ -64,34 +64,33 @@ export default class HookBrowser extends Component {
       return <Error error={error} />;
     }
 
-    if (!showList || !hooksList) {
-      return null;
-    }
-
     // For nodeLabel: forward clicks on the label on to the TreeView's handleClick (undocumented..)
     return (
-      <TreeView
-        key={group}
-        nodeLabel={<code onClick={this.handleLabelClick}>{group}</code>}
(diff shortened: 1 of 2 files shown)


src/views/HooksManager/HooksManager.js
```diff
@@ -76,6 +77,15 @@ export default class HooksManager extends React.PureComponent {
     }
   };
 
+  callbackChildLoaded(loadedGroup) {
+    this.setState({
+      groupsLoaded: [
```
eliperelman: Can you explain what the intent is here? From first glance it appears that we filter the list to exclude the loaded group, then move the loaded group back to the list at the end. What are we trying to achieve here?

src/views/HooksManager/HooksManager.js
```diff
@@ -97,6 +107,7 @@ export default class HooksManager extends React.PureComponent {
             group={group}
             selectHook={this.selectHook}
             hookGroupId={hookGroupId}
+            callbackChildLoaded={this.callbackChildLoaded.bind(this)}
```
eliperelman: We typically name event props in the format of `on{Event}={this.handle{Event}}`. So this would be written as something like:

```jsx
onLoadHooksList={this.handleLoadHooksList}
```

src/views/HooksManager/HooksManager.js
```diff
@@ -76,6 +77,15 @@ export default class HooksManager extends React.PureComponent {
     }
   };
 
+  callbackChildLoaded(loadedGroup) {
```
eliperelman: Rename this to `handleLoadHooksList` and make it an arrow function to avoid the need to re-bind in `render`:

```js
handleLoadHooksList = (loadedGroup) => {
  // ...
};
```

src/views/HooksManager/HookBrowser.js
```diff
@@ -47,6 +47,8 @@ export default class HookBrowser extends Component {
         error: err
       });
     }
+
+    callbackChildLoaded(group);
```
eliperelman: Rename to `onLoadHooksList`.

src/views/HooksManager/HooksManager.js
fiennyangeln: It is used to prevent duplicate inside the groupsLoaded.

fiennyangeln:
hi @eliperelman I changed it. pls check

src/views/HooksManager/HooksManager.js
eliperelman: Ah ok. For that, let's use a `Set`, which only contains unique values:

```js
groupsLoaded: [...new Set([
  ...this.state.groupsLoaded,
  loadedGroup
])]
```

eliperelman:
Looks good so far! Just one request for the `Set` and I think we can move forward. 👍

fiennyangeln:
@eliperelman ok modified!

fiennyangeln:
hi @helfi92, I modified it pls check

src/views/HooksManager/HooksManager.js
```diff
@@ -76,6 +77,19 @@ export default class HooksManager extends React.PureComponent {
     }
   };
 
+  handleLoadHooksList = loadedGroup => {
+    this.setState({
+      groupsLoaded: [...this.state.groupsLoaded, loadedGroup]
+    });
+  };
+
+  isAllGroupLoaded = () => {
```
helfi92: Can we use a different name? Maybe `groupsReady`? :)

src/views/HooksManager/HooksManager.js
```diff
@@ -76,6 +77,19 @@ export default class HooksManager extends React.PureComponent {
...
+    const { groups } = this.state;
+    const groupsLoadedSet = new Set(this.state.groupsLoaded);
+
+    return groups.filter(group => groupsLoadedSet.has(group)).length > 0;
```
helfi92: If one group is loaded (not all of them), then this line would return `true` since filtering would give an array of length 1. What do you think of something like:

```js
groupsReady = () =>
  this.state.groups.every(group => this.state.groupsLoaded.includes(group));
```

fiennyangeln: @helfi92 oh yeah.. Thanks for the catch!! changed as per your request, but I use .has instead of includes since it is for Set.

fiennyangeln:
@helfi92 changed! pls review:)

src/views/HooksManager/HookBrowser.js
```diff
@@ -56,15 +57,15 @@ export default class HookBrowser extends Component {
...
     const { hooksList, error } = this.state;
 
     if (error) {
       return <Error error={error} />;
     }
 
-    if (!hooksList) {
-      return <Spinner />;
+    if (!showList || !hooksList) {
+      return null;
```
helfi92: Let's inline this into the jsx:

```js
render() {
  // ...
  
  return (
    showList &&
    hooksList && (
      <TreeView
         ...
      </TreeView>
  );
}
```

src/views/HooksManager/HooksManager.js
```diff
@@ -76,6 +77,19 @@ export default class HooksManager extends React.PureComponent {
...
 
+  handleLoadHooksList = loadedGroup => {
+    this.setState({
+      groupsLoaded: [...this.state.groupsLoaded, loadedGroup]
+    });
+  };
+
+  groupsReady = () => {
+    const { groups } = this.state;
+    const groupsLoadedSet = new Set(this.state.groupsLoaded);
```
helfi92: Can you explain why we need a `Set` here? Are you sure this list will have duplicates? Do we even care if there are duplicates if we have it as:

```js
groupsReady = () =>
  this.state.groups.every(group => this.state.groupsLoaded.includes(group));
```

eliperelman: I said to use a `Set` in order to ensure no duplicates were put in here based on a previous review.

helfi92: @eliperelman I see. Would it still be necessary to have it here? We're only checking if the items in `groups` are included in the loaded list. It doesn't really matter if we have duplicates or not. Also, I'm not sure why we would have duplicates in the first place.

eliperelman: Actually, it looks like something got a little messed up here. The `Set` should be created in `handleLoadHooksList` instead of here.

```js
handleLoadHooksList = loadedGroup => {
  this.setState({
    groupsLoaded: [...new Set([...this.state.groupsLoaded, loadedGroup])]
  });
};
```

That would then allow this method to be:

```js
groupsReady = () =>
  this.state.groups.every(group => this.state.groupsLoaded.includes(group));
```

Like you propose. And yes, we want to remove duplicates, even if they have no impact, to avoid confusion in the future.

@fiennyangeln Can you revise based on this comment? We need to change the place this set is created at.

fiennyangeln: ok changed @eliperelman @helfi92 thanks for the guides!

src/views/HooksManager/HookBrowser.js
helfi92: @fiennyangeln I think this change still needs to be done.

fiennyangeln: Ok thanks for the guides!

fiennyangeln:
@helfi92 ok i modified the return inside the hooks browser.

fiennyangeln:
Thanks @helfi92 @eliperelman :)

PR #451, opened 2018-02-18
Bug 1438879 - Remove unnecessary spinner icons in Index Browser

JQuinnie:
When [IndexBrowser](https://github.com/taskcluster/taskcluster-tools/tree/master/src/components/IndexBrowser) loads, two spinners show up, one for `namespace` and one for `tasks`.

Create a function to capture the change in state of each as a single item. This will allow the [Spinner](https://github.com/taskcluster/taskcluster-tools/blob/master/src/components/Spinner/index.js) to load once only.

Reference: [Bug1438879](https://bugzilla.mozilla.org/show_bug.cgi?id=1438879)

### DIFF
--- a/src/components/IndexBrowser/index.js
+++ b/src/components/IndexBrowser/index.js
@@ -100,12 +100,6 @@ export default class IndexBrowser extends React.PureComponent {
     );
   };
 
-  renderSpinner() {
-    if (!this.state.namespaces && !this.state.tasks) {
-      return <Spinner />;
-    }
-  }
-
   renderBreadcrumbs() {
     const { namespace, urlRoot } = this.props;
     const parents = namespace.split('.');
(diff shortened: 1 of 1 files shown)


src/components/IndexBrowser/index.js
```diff
@@ -255,6 +252,7 @@ export default class IndexBrowser extends React.PureComponent {
               </div>
             </FormGroup>
           </form>
+          {this.renderSpinner()}
           {this.renderTasks()}
           {this.renderNamespaces()}
```
helfi92: `renderSpinner` and `renderTasks` no longer have:

```js
if (!this.state.namespaces) {
  return <Spinner />;
}
```

This means 

```js
  {this.renderSpinner()}
  {this.renderTasks()}
  {this.renderNamespaces()}
```

will always try to display the tasks and namespaces even if it's not ready to be rendered. How about something like this:

```js
{this.state.tasks && this.state.namespaces ? (
  <div>
    {this.renderTasks()}
    {this.renderNamespaces()}
  </div>
) : (
  <Spinner />
)}
```

JQuinnie: Ah Hassan,

I see what you are saying. I will go ahead and remove the `renderSpinner()` and make the `<Spinner />` a part of the return. Thank you for the advice!

Will update later today.

JQuinnie: Ah Hassan,

I see what you are saying. I will go ahead and remove the `renderSpinner()` and make the `<Spinner />` a part of the `return`. Thank you for the advice!

Will update later today.

JQuinnie:
Hi Hassan, I have updated the change request, please review when convenient. Thanks!

PR #452, opened 2018-02-18
Bug 1438862-Refresh button doesn't seem to be working in the client manager

Biboswan:


Biboswan:
@helfi92  previously `this` was becoming null inside loadClients .So refresh button wasn't loading clients again. Now its done as you said i think.

src/views/ClientManager/ClientManager.js
```diff
@@ -67,23 +68,24 @@ export default class ClientManager extends Component {
   }
 
   async loadClients() {
```
helfi92: Could you change this line to:

```js
loadClients = async () => {
```

Using arrow functions  has the benefit of not changing the meaning of `this`, so things just work automatically.

helfi92: You will then be able to delete  `this.loadClients = this.loadClients.bind(this);`.

src/views/ClientManager/ClientManager.js
```diff
@@ -66,25 +66,26 @@ export default class ClientManager extends Component {
...
-        error: null
-      });
-    } catch (err) {
-      this.setState({
-        clients: null,
-        error: err
-      });
-    }
-  }
+  loadClients = async () => {
```
eliperelman: This method isn't doing anything async, so we can remove that from the function definition.

```js
loadClients = () => {
```

helfi92: Agreed. @Biboswan let's remove it :)

PR #454, opened 2018-02-21
bug 1317289 scope error displayed after signing in under certain circumstances

hashi93:
# Fix error message not clearing and GET call not invoking on user session change
## Technical findings for Bug 1317289

1. The root cause for the bug is in the “componentWillReceiveProps” method in the “WorkerTypeView” component found in “src/views/AwsProvisioner/WorkerTypeView.js”

2. This is the root cause for “loadWorkerType” GET call for the worker type is not executed when a user session changes. Also, note that the GET call is only executed if “provisionerId” or “workerType” type changes. Moreover, there is no clearing of the error in “this.state”.

## Fix details

1. We can use “UserSession.userChanged(this.props.userSession, nextProps.userSession)” to check if the user session changes. We can add this as an additional OR condition for the if check in “componentWillReceiveProps” of “WorkerTypeView”. This ensures the GET call is fired when the user session changes.
2. To ensure the state error is cleared, we need to add “this.setState({ error: null });”. However, clearing this every time would cause an additional re-render of the “WorkerTypeView” component. We only need to clear the error if it is not null already. If state error is present, we first clear it allowing the spinner to be shown while the GET call is run, and in the callback of setState we invoke “loadWorkerType” GET call. If state error is not present we just invoke the “loadWorkerType” GET call without calling setState.

### DIFF
--- a/src/views/AwsProvisioner/WorkerTypeView.js
+++ b/src/views/AwsProvisioner/WorkerTypeView.js
@@ -8,6 +8,7 @@ import Code from '../../components/Code';
 import WorkerTypeEditor from './WorkerTypeEditor';
 import WorkerTypeResources from './WorkerTypeResources';
 import WorkerTypeStatus from './WorkerTypeStatus';
+import UserSession from '../../auth/UserSession';
 
 export default class WorkerTypeView extends React.PureComponent {
   static propTypes = {
@@ -37,9 +38,14 @@ export default class WorkerTypeView extends React.PureComponent {
   componentWillReceiveProps(nextProps) {
     if (
       nextProps.workerType !== this.props.workerType ||
-      nextProps.provisionerId !== this.props.provisionerId
+      nextProps.provisionerId !== this.props.provisionerId ||
+      UserSession.userChanged(this.props.userSession, nextProps.userSession)
     ) {
-      this.loadWorkerType(nextProps);
+      if (this.state.error) {
+        this.setState({ error: null }, () => this.loadWorkerType(nextProps));
+      } else {
+        this.loadWorkerType(nextProps);
+      }
     }
   }
 


src/views/AwsProvisioner/WorkerTypeView.js
```diff
@@ -37,9 +38,14 @@ export default class WorkerTypeView extends React.PureComponent {
   componentWillReceiveProps(nextProps) {
     if (
       nextProps.workerType !== this.props.workerType ||
-      nextProps.provisionerId !== this.props.provisionerId
+      nextProps.provisionerId !== this.props.provisionerId ||
+      UserSession.userChanged(this.props.userSession, nextProps.userSession)
     ) {
-      this.loadWorkerType(nextProps);
+      if (this.state.error) {
+        this.setState({ error: null }, () => this.loadWorkerType(nextProps));
```
helfi92: `loadWorkerType` will try to load the worker type and set `this.state.error` to null if successful. This is done here: https://github.com/hashi93/taskcluster-tools/blob/e53af97a2c409aa83a26b118e631abc73d596c8c/src/views/AwsProvisioner/WorkerTypeView.js#L61. To that end, I don't think we need the if statement anymore.

hashi93: image

1. As seen above, the setstate in "loadWorkerType" will not be complete until the async calls are complete and the parameter object to setState is created. Therefore, the error will not be cleared until the next GET calls are complete. I checked locally and if we clear the error here, no Spinner is shown in between the GET calls.

2. To my best of knowledge, we should first clear the state error, render the component, then invoke "loadWorkerType", this way we will see a Spinner while the data is fetched. Clearing state error inside"loadWorkerType" has no visual effect. I did not remove that line in the PR because we are setting state with other data anyway there and to keep the lines of impacted code to a minimum.

3. Please advice me on which approach to take. We can remove "error: null," in the setstate inside "loadWorkerType" (Recomended) or keep it and just call "loadWorkerType" inside "componentWillReceiveProps" without clearing the error first (no Spinner between GET calls, the previous error will be there until next GET is complete).

helfi92: > we should first clear the state error, render the component, then invoke loadWorkerType, this way we will see a Spinner while the data is fetched

Agreed. The user experience is important here :) I am really happy with your findings.

helfi92:
I'll proceed with the merge next week. Friday is not the best day to merge stuff hehe :)

PR #455, opened 2018-02-22
Bug 1436115 solved - Scope editor should not highlight '*' if it is n…

kanikasaini:
Bug 1436115 solved - Scope editor should not highlight '*' if it is not the last character in the scope.

### DIFF
--- a/src/components/ScopeEditor/scopemode.js
+++ b/src/components/ScopeEditor/scopemode.js
@@ -23,7 +23,7 @@ CodeMirror.registerHelper('lint', 'scopemode', text =>
 
 CodeMirror.defineSimpleMode('scopemode', {
   start: [
-    { regex: /\*/, token: 'star' },
+    { regex: /\*($|(\s+))/, token: 'star' },
     { regex: /:/, token: 'sep' },
     { regex: /\//, token: 'slash' },
     { regex: /<\.>/, token: 'mistake-param' },


helfi92:
This is what I am currently seeing:

screen shot 2018-02-22 at 10 51 16 am

kanikasaini:
No, Hassan.
Here we wanted to highlight the '*' only when it appears at the end of a word, like abc*, so that it depicts the word has a special meaning. And it should not be highlighted when the '*' is a part of the word itself like abc*def, as it has no special meaning then.

helfi92:
Ah I see, my mistake. Will review the pull-request now :)

djmitche:
Spaces [are allowed in scopes](https://github.com/taskcluster/taskcluster-auth/blob/44dc61eafc3454c145ec45b5b3201a76b4d2589f/schemas/create-client-request.yml#L33) (`\x20` is a space character).  So that highlighting would be misleading.

src/components/ScopeEditor/scopemode.js
```diff
@@ -23,7 +23,7 @@ CodeMirror.registerHelper('lint', 'scopemode', text =>
 
 CodeMirror.defineSimpleMode('scopemode', {
   start: [
-    { regex: /\*/, token: 'star' },
+    { regex: /\*($|(\s+))/, token: 'star' },
```
djmitche: If the `$` pattern matches the end of a line, then `\*$` would be correct here.

kanikasaini: Yes, Dustin. If that's the case, then only \*$ should be there. I wasn't clear about what we actually wanted. Will fix it. Thank you!

PR #456, opened 2018-02-22
Bug 1440455 Clear the right panel in the scope expander view when the…

blindacai:
Bug 1440455: Clear the right panel in the scope expander view when the clean button is clicked


helfi92:
Hi @blindacai. This fixes the problem :) However, this bug was assigned to Aditi and so we have to go with her pull-request. Sorry about that.

blindacai:
@helfi92 Thanks, I understand that. I suppose I should close this pull request?

PR #457, opened 2018-02-23
Bug 1440455 resolved- Clear the right panel in the scope expander view when the clean button is clicked

Aditi574:
Clear the right panel in the scope expander view when the clean button is clicked

### DIFF
--- a/src/views/ScopesetExpander/ScopesetExpander.js
+++ b/src/views/ScopesetExpander/ScopesetExpander.js
@@ -55,7 +55,12 @@ export default class ScopesetExpander extends React.PureComponent {
   }
 
   handleClickCleanScopes = () => {
-    this.setState({ scopes: [] });
+    this.setState({
+      scopes: [],
+      expandedScopes: null,
+      error: null,
+      scopesIsValid: false
+    });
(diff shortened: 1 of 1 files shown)


helfi92:
@Aditi574 I see that `handleClickCleanScopes` is setting the state to the the initial state. To mitigate duplication, could we do:

```js
export default class ScopesetExpander extends React.PureComponent {
  initialState = {
    scopes: [],
    expandedScopes: null,
    error: null,
    scopesIsValid: false
  };
 
  state = this.initialState;

  //   ...

  handleClickCleanScopes = () => {
    this.setState(this.initialState);
  };

  // ...
```

Aditi574:
@helfi92 yes you're right! Let me make the changes.

helfi92:
@Aditi574 I think your change was only pushed to your fork but not here.

Aditi574:
@helfi92 I created a new PR #458 after making the changes. Is that incorrect?

helfi92:
In the future, it's better to push your changes to the same pull-request. I'll close this pull-request for now and we can carry on in #458.

PR #458, opened 2018-02-23
Bug 1440455 resolved- Clear the right panel in the scope expander view when the clean button is clicked 

Aditi574:
Shorter code than previous one.


Aditi574:
@helfi92 Thanks. I will keep the pointers you mentioned in mind. :)

PR #460, opened 2018-02-24
Bug 1440374 solved - Remove infinite spinner when editor is empty

kanikasaini:


kanikasaini:
@eliperelman When there's no task in the editor, the empty editor remains on the screen.

kanikasaini:
@helfi92 Thank you for pointing that out. :) 
However, it didn't generate an error when the editor was empty but now it looks cleaner. Both the buttons - Create Task and Update Timestamps are disabled when there's no task in the editor.

PR #461, opened 2018-02-24
bug 1440764: add refresh button in the worker types explorer

YinYue33:
I added a button to rerender the whole page on  https://tools.taskcluster.net/provisioners/aws-provisioner-v1/worker-types

### DIFF
--- a/src/views/WorkerTypes/WorkerManager.js
+++ b/src/views/WorkerTypes/WorkerManager.js
@@ -82,6 +82,8 @@ export default class WorkerManager extends React.PureComponent {
   handleLayoutChange = () =>
     this.setQuery({ layout: this.state.layout === 'grid' ? 'table' : 'grid' });
 
+  handleRefresh = () => window.location.reload();
+
   handleLastActiveClick = () =>
     this.setQuery({ lastActive: !this.state.lastActive, orderBy: null });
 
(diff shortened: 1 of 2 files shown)


YinYue33:
Sure, I will work on this today

YinYue33:
I added a props to WorkerTypeTable to keep track of clicking. Now workerTypeTable will update either when provisionerID changes or when refresh button is clicked.

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -32,6 +32,7 @@ export default class WorkerManager extends React.PureComponent {
       orderBy: null,
       actions: [],
       error: null,
+      clickState: false,
```
helfi92: Can we rename this to something like `refreshFlag`.

src/views/WorkerTypes/WorkerManager.js
```diff
@@ -218,6 +223,9 @@ export default class WorkerManager extends React.PureComponent {
                 </ToggleButton>
               </ToggleButtonGroup>
             </ButtonToolbar>
+            <Button onClick={this.handleRefresh} bsSize="sm">
```
helfi92: Having the button outside `<ButtonToolbar>` will have the refresh button displayed on a new line. I would like to see the refresh button be aligned with the rest of the buttons.

src/views/WorkerTypes/WorkerManager.js
eliperelman: I think using a boolean here also gives a false impression about needing to refresh or not. Let's use a numeric increment instead:

```js
handleRefresh = () => {
  this.setState({ refreshId: ++this.state.refreshId });
};
```

YinYue33: Good idea. But will this potentially exceed largest integer as the server runs?

PR #462, opened 2018-02-25
Bug 1440448 : Task & Group Inspector enter key to search

akshitac8:
Do share your reviews.

### DIFF
--- a/src/views/UnifiedInspector/SearchForm.jsx
+++ b/src/views/UnifiedInspector/SearchForm.jsx
@@ -184,3 +184,4 @@ export default class SearchForm extends React.PureComponent {
     );
   }
 }
+


src/views/UnifiedInspector/SearchForm.js
```diff
@@ -61,18 +62,16 @@ export default class SearchForm extends React.PureComponent {
   };
 
   handleSubmit = e => {
-    e.preventDefault();
-
-    const isValid = this.validate().form;
-
-    if (!isValid) {
-      return;
-    }
```
helfi92: What's the reason behind the removal of L64 - L70?

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -61,18 +62,16 @@ export default class SearchForm extends React.PureComponent {
...
-
     this.props.onSearch({
       taskGroupId: this.state.taskGroupIdInput || '',
       taskId: this.state.taskIdInput || ''
     });
+    e.preventDefault();
```
helfi92: Can you put `e.preventDefault();` at the beginning of the method, not at the end so that we make it clear the default action should not be taken.

src/views/UnifiedInspector/SearchForm.js
```diff
@@ -34,6 +34,7 @@ export default class SearchForm extends React.PureComponent {
       taskGroupIdInput: props.taskGroupId || '',
       taskIdInput: props.taskId || ''
     };
+    this.handleSubmit = this.handleSubmit.bind(this);
```
helfi92: This is not needed, you're already using the arrow notation to define a method.

akshitac8: Yes, I have removed it

src/views/UnifiedInspector/SearchForm.js
akshitac8: We need to keep it at the end of the function because it is preventing the link to open on form submission on enter according to what i understood

akshitac8: Sir, To be more accurate e.preventdefault() should be called only when validation call is false.
I have updated this in my recent pull request. Sorry for the inconvenience and multiple pull request.

src/views/UnifiedInspector/SearchForm.js
akshitac8: Sorry,That was due to slow network commit. I executed a commit on the latest file and it took last updated file. 
I have made 2 last pull request with changes in the validation function.

akshitac8:
@eliperelman  I am working on this but i found while googling that i need to use keydown event. Is that so?

eliperelman:
@akshitac8 no, you shouldn't need to use that. Every time you type in the search field, the `onChange` handler will set the new value of the field in state. When you hit enter or click the inspect button, that should trigger the `onSubmit` handler. We are only concerned with the second case.

Have you verified whether or not the `onSubmit` handler gets triggered if you hit enter in the search field? See my previous comment.

akshitac8:
Yes, I understood it should automatically trigger on enter key but it is not. I am looking might get it working by tomorrow.

eliperelman:

akshitac8:
Hello,
I need to make few changes and then I will make a new PR. Kindly consider that for Review.
Sorry for the inconvience.

helfi92:
Continued in https://github.com/taskcluster/taskcluster-tools/pull/482.

PR #465, opened 2018-02-26
Bug 1440342 - scary warning for any callback_url but localhost

djmitche:
This service is really *only* intended for use with `taskcluster login`, so let's make everything else throw up a big, scary warning.

### DIFF
--- a/src/views/ClientCreator/ClientCreator.js
+++ b/src/views/ClientCreator/ClientCreator.js
@@ -207,7 +207,7 @@ export default class ClientCreator extends React.PureComponent {
             bsStyle="danger"
             onClick={() =>
               this.setState({ unknownCallbackAcknowledged: true })}>
-            I am Sure
+            Proceed
           </Button>
         </ButtonToolbar>
       </Alert>
--- a/src/views/ClientCreator/ClientCreator.js
+++ b/src/views/ClientCreator/ClientCreator.js
@@ -181,6 +181,15 @@ export default class ClientCreator extends React.PureComponent {
     </div>
   );
 
+  renderInvalidQuery = () => (
+    <div>
+      <Alert bsStyle="danger">
+        This tool must be invoked with query parameters{' '}
+        <code>callback_url</code> and <code>name</code>.
+      </Alert>
+    </div>
+  );
+
   renderUnknownCallback = () => (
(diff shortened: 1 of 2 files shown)


src/views/ClientCreator/ClientCreator.js
```diff
@@ -174,19 +175,55 @@ export default class ClientCreator extends React.PureComponent {
...
     if (this.state.loading) {
       return <Spinner />;
     }
 
+    if (
+      !this.isWhitelistedCallback(this.state.query.callback_url) &&
+      !this.state.unknownCallbackAcknowledged
+    ) {
+      return this.renderUnknownCallback();
+    }
```
helfi92: Could we put this check at the beginning of the `render` method? We don't want to end up in a scenario where the spinner is shown instead of the error view.

djmitche:
Done!

src/views/ClientCreator/ClientCreator.js
```diff
@@ -174,19 +175,55 @@ export default class ClientCreator extends React.PureComponent {
...
+        <code>taskcluster signin</code> operation.
+        <br />
+        Granting access to another URL might expose your credentials to an
+        attacker that controls that URL.
+        <br />
+        Are you sure you want to proceed?
+        <ButtonToolbar className={styles.flexRight}>
+          <Button
+            bsStyle="danger"
+            onClick={() =>
```
eliperelman: Let's also move this to its own method.

src/views/ClientCreator/ClientCreator.js
```diff
@@ -174,19 +175,55 @@ export default class ClientCreator extends React.PureComponent {
...
+              this.setState({ unknownCallbackAcknowledged: true })}>
+            I am Sure
```
eliperelman: Let's make the prose of the question match the prose of the answer by making this button say "Proceed".

helfi92:
@djmitche I think the error view would show if the URL doesn't contain a callback_url, but I don't think we should worry about this edge case right?

djmitche:
That's a good point - the error message will look funny.  I'll add another error for that situation :)

djmitche:
@eliperelman updated

PR #466, opened 2018-02-28
Bug 1441655: Error box showing in Roles Manager when deleting a role

blindacai:
Bug 1441655: Error box showing in Roles Manager when deleting a role

### DIFF
--- a/src/components/ModalItem/index.js
+++ b/src/components/ModalItem/index.js
@@ -40,14 +40,10 @@ export default class ModalItem extends React.PureComponent {
       e.preventDefault();
     }
 
-    if (this.state.error) {
-      this.setState({ error: null });
-    }
-
     this.setState({ show: true });
   };
 
-  handleClose = () => this.setState({ show: false });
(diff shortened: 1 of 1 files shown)


src/components/ModalItem/index.js
```diff
@@ -40,6 +40,10 @@ export default class ModalItem extends React.PureComponent {
       e.preventDefault();
     }
 
+    if (this.state.error) {
+      this.setState({ error: null });
+    }
+
```
blindacai: on second thought, I think it may be better to reset error to null in handleClose:
`handleClose = () => this.setState({ show: false, error: null });`
What do you think?

blindacai: and not sure how necessary the if conditional check is, but it can be moved down to handleClose() too

eliperelman: I think that would be a good solution. You could always reset the error and show to null. Let's go with that. :)

blindacai:
Thanks for your reviews. An update is added

PR #467, opened 2018-02-28
Bug 1441652: Rename 'task-graphs' to 'task groups' in tools homepage

blindacai:


blindacai:
I've been having this problem with prettier/prettier.
I am on Windows 10 using IDEs like VS code or Atom, and have installed some prettier package on them, but apparently it still causes problems.

Is there a recommended dev environment? I am thinking code prettify should be an auto process?

eliperelman:
@blindacai It is built into the repo, so you can run `yarn lint --fix` and it will be fix anything it can. Give it a shot here, then push up the commit. Thanks!

blindacai:
@eliperelman Thanks, that indeed fixed the format problem. 
At first, running lint seemed to have modified every file due to a different way of line ending in Windows, but they were eventually ignored.

PR #470, opened 2018-03-01
Bug 1441615 - Hooks Editor: clicking on create hook while logged out

kanikasaini:
Hooks Editor: clicking on create hook while logged out will make you lose your form.

Whenever User session is changed, React does forceupdate() of all the components but earlier error status wasn't getting updated upon re-rendering. This change allows error to be reset to null on user login while other fields in the editor retain their data. 

As this cannot be checked by running it on the local server, I request the maintainers to please try review the change.

### DIFF
--- a/src/views/HooksManager/HookEditor.jsx
+++ b/src/views/HooksManager/HookEditor.jsx
@@ -1,6 +1,6 @@
 import { PureComponent } from 'react';
 import { string, bool, object, func } from 'prop-types';
-import { Button, Glyphicon, ButtonToolbar } from 'react-bootstrap';
+import { Alert, Button, Glyphicon, ButtonToolbar } from 'react-bootstrap';
 import clone from 'lodash.clonedeep';
 import equal from 'deep-equal';
 import { assocPath } from 'ramda';
@@ -45,10 +45,12 @@ export default class HookEditor extends PureComponent {
     currentHookId: string,
     currentHookGroupId: string,
     hook: object,
+    error: object,
     isCreating: bool,
     onCreateHook: func.isRequired,
     onUpdateHook: func.isRequired,
-    onDeleteHook: func.isRequired
+    onDeleteHook: func.isRequired,
+    onError: func.isRequired
   };
 
   constructor(props) {
@@ -279,6 +281,16 @@ export default class HookEditor extends PureComponent {
   render() {
     const { isCreating, hookGroupId, hookId } = this.props;
     const { hook } = this.state;
+
+    if (this.props.error) {
+      return (
+        <Alert bsStyle="danger" onDismiss={this.props.onError}>
+          <strong>Error executing operation</strong>
+          <br />
+          {this.props.error.toString()}
+        </Alert>
+      );
+    }
 
     if (!hook) {
       return null;
--- a/src/views/HooksManager/HookEditView.jsx
+++ b/src/views/HooksManager/HookEditView.jsx
@@ -35,7 +35,7 @@ export default class HookEditView extends PureComponent {
       nextProps.hookId !== this.props.hookId
     ) {
       this.loadHook(nextProps);
-    }
+    } else this.dismissError();
   }
 
   async loadHook({ hookGroupId, hookId, hooks }) {
--- a/src/views/HooksManager/HookEditView.jsx
+++ b/src/views/HooksManager/HookEditView.jsx
@@ -1,7 +1,6 @@
 import { PureComponent } from 'react';
 import { string, func } from 'prop-types';
 import { omit } from 'ramda';
-import { Alert } from 'react-bootstrap';
 import Spinner from '../../components/Spinner';
 import HookEditor from './HookEditor';
 import HookDisplay from './HookDisplay';
@@ -35,7 +34,9 @@ export default class HookEditView extends PureComponent {
       nextProps.hookId !== this.props.hookId
     ) {
       this.loadHook(nextProps);
-    } else this.dismissError();
+    } else {
+      this.handleDismissError();
+    }
   }
 
   async loadHook({ hookGroupId, hookId, hooks }) {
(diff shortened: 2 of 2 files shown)


kanikasaini:
@djmitche  Since the latest change doesn't affect HookEditView.js, I believe the solution remains the same. :)

kanikasaini:
@helfi92 Thank you. 
It now works fine when the alert window is closed.

src/views/HooksManager/HookEditView.js
```diff
@@ -23,6 +22,8 @@ export default class HookEditView extends React.PureComponent {
       editing: true,
       error: null
     };
+
+    this.dismissError = this.dismissError.bind(this);
```
eliperelman: This is unnecessary since `dismissError` is already defined as an arrow function.

src/views/HooksManager/HookEditView.js
```diff
@@ -35,7 +36,7 @@ export default class HookEditView extends React.PureComponent {
       nextProps.hookId !== this.props.hookId
     ) {
       this.loadHook(nextProps);
-    }
+    } else this.dismissError();
```
eliperelman: For block statements, always use curly braces:

```js
} else {
  this.dismissError();
}
```

src/views/HooksManager/HookEditView.js
```diff
@@ -176,6 +167,7 @@ export default class HookEditView extends React.PureComponent {
           createHook={this.createHook}
           updateHook={this.updateHook}
           deleteHook={this.deleteHook}
+          error={this.state.error}
```
eliperelman: Why are we pushing the error down into the child instead of handling it where it happens, in this current component?

src/views/HooksManager/HookEditor.js
```diff
@@ -1,12 +1,13 @@
...
 import { string, bool, object, func } from 'prop-types';
-import { Button, Glyphicon, ButtonToolbar } from 'react-bootstrap';
+import { Alert, Button, Glyphicon, ButtonToolbar } from 'react-bootstrap';
 import clone from 'lodash.clonedeep';
 import equal from 'deep-equal';
 import { assocPath } from 'ramda';
 import Icon from 'react-fontawesome';
 import ModalItem from '../../components/ModalItem';
 import CodeEditor from '../../components/CodeEditor';
+
```
eliperelman: Delete this extra line.

src/views/HooksManager/HookEditor.js
```diff
@@ -59,15 +61,19 @@ export default class HookEditor extends React.PureComponent {
       newScheduleValue: '',
       hook: null,
       hookValidJson: false,
-      triggerSchemaValidJson: false
+      triggerSchemaValidJson: false,
+      error: this.props.error
     };
+
+    this.dismissError = this.dismissError.bind(this);
```
eliperelman: This isn't necessary since `dismissError` is defined as an arrow function.

src/views/HooksManager/HookEditor.js
```diff
@@ -59,15 +61,19 @@ export default class HookEditor extends React.PureComponent {
...
   }
 
   componentWillMount() {
     this.setState({
       hook: clone(this.props.isCreating ? initialHook : this.props.hook),
       hookValidJson: true,
-      triggerSchemaValidJson: true
+      triggerSchemaValidJson: true,
+      error: this.props.error
```
eliperelman: I am still not comfortable passing errors as props to children. There needs to be good justification for why it needs to be done this way as opposed to handling in the parent.

src/views/HooksManager/HookEditor.js
```diff
@@ -280,6 +291,16 @@ export default class HookEditor extends React.PureComponent {
     const { isCreating, hookGroupId, hookId } = this.props;
     const { hook } = this.state;
 
+    if (this.props.error && this.state.error) {
```
eliperelman: If we decide to go this route, this would probably need to be logical OR `||` instead of logical AND `&&`.

src/views/HooksManager/HookEditor.js
```diff
@@ -280,6 +291,16 @@ export default class HookEditor extends React.PureComponent {
...
+      return (
+        <Alert bsStyle="danger" onDismiss={this.dismissError}>
+          <strong>Error executing operation</strong>
+          <br />
+          {this.props.error.toString()}
```
eliperelman: Here we only handle if error belongs to props and not state.

src/views/HooksManager/HookEditView.js
kanikasaini: As we were trying to maintain the form fields when we dismiss the error (which was then being handled by the parent), we required the child props to change so that 'componentWillRecieveProps' should be called (to use prevState to set the current state) whenever the props are updated.

For this to happen, we had to keep the HTML code for alert window (upon dismissing which the props change ) and the call to the HookEditView component in the same return block in render. 
Doing this would compel us to show the alert window even when there's no error. 

The only way I could think of solving it was to render the alert box when the error is passed to HookEditView, else display the normal editor form. 
Can you please help if this is not the optimal solution to this problem?

eliperelman:

src/views/HooksManager/HookEditView.js
kanikasaini: @eliperelman Can you please approve this way which involves passing the error prop to the child?
I will create a rebased PR after that only with desired changes.

kanikasaini:
@eliperelman Can you please approve this way which involves passing the error prop to the child?
I will create a rebased PR after that only with desired changes.

kanikasaini:
@eliperelman @helfi92 Please review whenever convenient.

src/views/HooksManager/HookEditor.jsx
```diff
@@ -45,10 +45,12 @@ export default class HookEditor extends PureComponent {
     currentHookId: string,
     currentHookGroupId: string,
     hook: object,
+    error: object,
```
eliperelman: So this is what I was asking about. If we already have the error in the component, why do we need to pass it to the parent, which passes it back to the child, to render in the child?

I think this boils down to how we re-structured this component when we split the views. Instead of calling these methods in the parent, they should now live in the child:

```js
onCreateHook
onDeleteHook
onUpdateHook
```

If we take them out of the parent, and move them into the child, when they have an error, we can set their state in the editor and show the error there as well.

kanikasaini:
@eliperelman @helfi92 Thank you! :)

PR #476, opened 2018-03-02
Upgrade preset and project to Neutrino v8, fix associated linting errors

eliperelman:


package.json
```diff
@@ -62,7 +59,7 @@
     "react-dom": "^15.5.4",
     "react-fontawesome": "^1.6.1",
     "react-helmet": "^5.2.0",
-    "react-hot-loader": "3.0.0-beta.6",
+    "react-hot-loader": "^3.0.0",
```
helfi92: HMR is not working. How hard would it be to switch to v4?

src/components/ArtifactList/index.jsx
```diff
@@ -18,7 +18,8 @@ export default class ArtifactList extends React.PureComponent {
 
   static defaultProps = {
     menu: false,
-    style: {}
+    style: {},
+    artifacts: null
```
helfi92: I'm surprised this is allowed. If the type of `artifacts` is an array, then I wouldn't expect to handle another type `null`.  Not a big deal though.

src/views/ScopeGrants/ScopeGrants.jsx
```diff
@@ -162,7 +161,7 @@ export default class ScopeGrants extends React.PureComponent {
 
     await Promise.all([
       ...toCreate.map(role =>
-        this.props.auth.createRole(role, {
+        this.props.auth.handleCreateRole(role, {
```
helfi92: This should probably stay `this.props.auth.createRole` :)

eliperelman: Ugh, I thought I caught all these. Webstorm's refactor tool messed this up.

src/components/ArtifactList/index.jsx
eliperelman: If the prop is required, it will be, this just initializes it when it's not provided. This is how all props work when not validated.

package.json
eliperelman: Not too hard, maybe I can bake this into the Neutrino preset.

eliperelman: To be fair, we didn't have HMR before this patch either. 🤣

eliperelman:
We are going to have to merge this to unblock a contribution, but there will be some fallout for PR rebasing. 😿

PR #478, opened 2018-03-03
Bug 1442896 - make roles manager have multiple views

kanikasaini:


blindacai:
Since I am working on one of the sibling bugs (make Scope Grants have multiple views), I am digging around and see how people implement theirs.

One thing I noticed in RoleManager.js:
renderolerow
selectedRoleId is not one of the initial states, so it seems `this.state.selectedRoleId` is never defined and isSelected is always false, thus 'info' attribute is never applied and the row will not be highlighted when selected.

and after this bug is resolved, we don't even need the select -> highlight feature. I think we can remove those lines and merge `renderRoleRow` to `renderRolesTable` since it only returns some html

but this is just to make the code look better and won't change user experience. Not sure if you want a PR for it?

eliperelman:
@blindacai we would always welcome a PR that improves something, even if it's only to improve the code's readability or maintainability. 😃

kanikasaini:
@blindacai Thank you for pointing out `selectedRoleId`.  :)
That is not being used in this case as upon clicking a row we're redirected to its editing page. Therefore, I have removed these unused lines in the latest commit. 
However, I think merging `renderRolesTable` into `renderRoleRow` only makes it tougher to understand as it's a method which is being called for every role in the roleslist. 

@eliperelman May decide better.

kanikasaini:
@eliperelman I have edited the files as required. :) Please review.

blindacai:
@eliperelman Cool.
@kanikasaini Sounds good. I guess it's not a big deal whether to merge `renderRolesTable` or not and I don't think it will make the code hard to read if we do so:
renderrolestable
Not bad, right?

kanikasaini:
@blindacai Looks good! @eliperelman Should I make these changes in this PR itself?

src/views/RoleManager/index.js
```diff
@@ -5,10 +5,12 @@ import WithUserSession from '../../components/WithUserSession';
 import RoleManager from './RoleManager';
 
 const View = ({ history, match }) => {
-  const roleId = window.location.hash.slice(1);
+  const [roleId, create] = window.location.hash.slice(1).split('/');
```
eliperelman: This hash is a carry-over from when we used to use hash-based routing and the role ID would be in the hash. This would not be the case for new routes, like `create`. I'm pretty sure we can remove these changes.

src/views/RoleManager/RoleManager.js
```diff
@@ -117,16 +112,23 @@ export default class RoleManager extends React.PureComponent {
...
+        history={this.props.history}
+        currentRoleId={this.props.roleId === 'create' ? '' : this.props.roleId}
+        deleteRole={this.deleteRole}
+        navigate={this.navigate}
+      />
+    );
+  }
+
+  render() {
+    return !this.props.roleId ? this.renderRoles() : this.renderRoleEditor();
```
eliperelman: If you invert the ternary expression here, you can write this without having to negate the condition:

```js
return this.props.roleId ? this.renderRoleEditor() : this.renderRoles();
```

src/views/RoleManager/RoleManager.js
```diff
@@ -52,26 +52,24 @@ export default class RoleManager extends React.PureComponent {
...
-        <td>
-          <Link to={`/auth/roles/${encodeURIComponent(role.roleId)}`}>
-            <code>{role.roleId}</code>
-          </Link>
-        </td>
-      </tr>
-    );
-  };
+  renderRoleRow = (role, index) => (
+    <tr key={index}>
```
eliperelman: Can you change this to use a different key?

```jsx
<tr key={`role-row-${role.id}`}>
```

src/views/RoleManager/RoleManager.js
```diff
@@ -52,26 +52,24 @@ export default class RoleManager extends React.PureComponent {
 
   navigate = roleId => {
     this.load();
-    this.props.history.replace(
-      `/auth/roles/${roleId ? encodeURIComponent(roleId) : ''}`
+    const { history } = this.props;
+
+    history.push(
```
eliperelman: Wouldn't this always force the user to either be on the role edit or role create page? How would you get back to the list using this method?

eliperelman:

kanikasaini:
@eliperelman Thank you for the rebasing tutorial! It made things go smoothly. :)
I have made changes as suggested.

src/views/RoleManager/index.js
kanikasaini: Removed. :)

src/views/RoleManager/RoleManager.js
kanikasaini: Done! Thank you.

src/views/RoleManager/RoleManager.js
kanikasaini: Please check if I have understood what you meant correctly or not.
Are we taking the key as the difference of roleRow and role.Id?

kanikasaini:
@eliperelman Please review. Thank you.

src/views/RoleManager/RoleManager.jsx
```diff
@@ -96,17 +92,16 @@ export default class RoleManager extends PureComponent {
...
         <HelmetTitle title="Role Manager" />
-        <Col md={5}>
+        <Col md={12}>
           {this.renderRolesTable()}
           <ButtonToolbar>
             <Button
               bsStyle="primary"
-              href="/auth/roles"
-              disabled={this.props.roleId === ''}>
+              onClick={() => this.props.history.push(`/auth/roles/create`)}>
```
helfi92: Could you use `react-router-dom` to perform the route change:

```js
import { Link } from 'react-router-dom';

// ...
<Link to="/auth/roles/create">
  <Button bsStyle="primary" ...
</Link>

```

src/views/RoleManager/index.jsx
```diff
@@ -4,7 +4,7 @@ import WithUserSession from '../../components/WithUserSession';
 import RoleManager from './RoleManager';
 
 const View = ({ history, match }) => {
-  const roleId = window.location.hash.slice(1);
+  const [roleId] = window.location.hash.slice(1).split('/');
```
helfi92: See how we handle routing in https://github.com/taskcluster/taskcluster-tools/blob/master/src/App/index.jsx#L174-L323.

This should then just be `match.params.roleId`.

kanikasaini: Hey, changing the code to `const [roleId] = match.params.roleId` generates the following error on refreshing the page. (after I have added the `<Link>` to the button)
image

helfi92: Hm, what do you get if you do:

```
console.log(match.params)
```

kanikasaini: Okay, so it prints the expected roleId. I think that error resolves when I directly add match.params.roleId in the if condition rather than assigning it to a variable, like 
` if (match.params.roleId) {
    return <Redirect to={'/auth/roles/${match.params.roleId}'} />;
  }`

The previous error is no more there but the pages still don't load and there are some new warnings. 
image

src/views/RoleManager/RoleManager.jsx
```diff
@@ -50,28 +50,24 @@ export default class RoleManager extends PureComponent {
...
-  handleNavigate = roleId => {
-    this.handleLoad();
-    this.props.history.replace(
-      `/auth/roles/${roleId ? encodeURIComponent(roleId) : ''}`
-    );
+  navigate = roleId => {
+    this.load();
+    const { history } = this.props;
+
+    history.push(`/auth/roles/${roleId ? encodeURIComponent(roleId) : ''}`);
```
helfi92: I think you want to push `/auth/roles` and not `/auth/roles/` when `roleId` is not defined.

src/views/RoleManager/index.jsx
helfi92: Is it possible you're already on `/auth/roles/create` and trying to navigate to `/auth/roles/create`? That would explain the error.

helfi92: Did you create the `auth/roles/create` route in `src/App/index.js`?

```js
<PropsRoute
  path="/auth/roles/create"
 // ...
/>
```

src/views/RoleManager/RoleManager.jsx
kanikasaini: Fixed this :). The warnings remain the same.
However, I didn't create the auth/roles/create route in src/App/index.js as create becomes the roleId in this implementation.
` <RoleEditor
        auth={this.props.auth}
        history={this.props.history}
        currentRoleId={this.props.roleId === 'create' ? '' : this.props.roleId}
       ...
      />`

helfi92: I don't think we want `create` to be the `roleId` here. That will be confusing to the reader reading the code. @eliperelman can maybe confirm that.

eliperelman:
Apologies, but we are going to hold off on merging any PRs that have to deal with splitting views until we can ensure that they are done consistently and using the techniques that we lay out. Thank for your patience here.

PR #481, opened 2018-03-06
Bug 1441950 add ability to filter through the list of roles

hashi93:
# Added Search Panel
## Fix Detail
The search panel has been added to the RoleManager view.Furthermore, its functionality has been implemented.


src/views/RoleManager/RoleManager.js
```diff
@@ -89,18 +99,42 @@ export default class RoleManager extends React.PureComponent {
...
             .sort((a, b) => a.roleId.localeCompare(b.roleId))
             .map(this.renderRoleRow)}
         </tbody>
       </Table>
     );
   }
 
+  renderTypeInput() {
+    return (
+      <div className="form-group form-group-sm">
```
helfi92: Let's rely on react-bootstrap to generate the form instead of manually adding the class names (i.e., `form-group`, `from-group-sm`, etc...). For inspiration, you can look at https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/WorkerTypes/SearchForm.jsx#L48-L62.

After we merge this pull-request, I'm thinking you can take `src/views/WorkerTypes/SearchForm.jsx` and move it to the `src/components` folder and have both views use this component instead of having duplicates. We can create a separate bug for that later.

src/views/RoleManager/RoleManager.js
```diff
@@ -73,6 +74,15 @@ export default class RoleManager extends React.PureComponent {
     );
   };
 
+  setRoleId = e => this.setState({ roleIdContains: e.target.value });
+
+  enterRoleId = e => {
```
helfi92: If you wrap `renderTypeInput` with a `<form>` element, then this won't be needed anymore.

src/views/RoleManager/RoleManager.js
```diff
@@ -89,18 +99,42 @@ export default class RoleManager extends React.PureComponent {
         </thead>
         <tbody>
           {this.state.roles
+            .filter(role => role.roleId.includes(this.state.roleIdContains))
```
eliperelman: Can we use `localeCompare` here as well to filter regardless of case sensitivity?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

eliperelman:
I think the latest merge overwrote our review comments here.

src/views/RoleManager/RoleManager.js
hashi93: @eliperelman Isn't `localeCompare` used to sort strings, How are we going to use that to perform filtering regardless of case sensitivity?

src/views/RoleManager/RoleManager.js
hashi93: @helfi92 used `SearchForm` component to implement the search panel as you requested

src/views/RoleManager/RoleManager.js
hashi93: @helfi92 removed `enterRoleId ` and refactored `setRoleId` to `handleSetRoleId`

hashi93:
@eliperelman the merge with the master was required to resolve conflicts

PR #482, opened 2018-03-06
Bug 1440448 : Task & Group Inspector enter key to search

akshitac8:
Hello,
This PR is in reference with PR <https://github.com/taskcluster/taskcluster-tools/pull/462>
Final changes made to the file with invisible button hack.
Referred : [Bug 1440448](https://bugzilla.mozilla.org/show_bug.cgi?id=1440448)

### DIFF
--- a/src/views/UnifiedInspector/SearchForm.jsx
+++ b/src/views/UnifiedInspector/SearchForm.jsx
@@ -8,7 +8,8 @@ import {
   FormControl,
   ButtonGroup,
   DropdownButton,
-  MenuItem
+  MenuItem,
+  Button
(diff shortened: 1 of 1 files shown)


akshitac8:
So, I should just keep the form submission with enter key?

helfi92:
We definitely don't want to manually take care of the enter key. I would just use something like `<form onSubmit={...}>` and make sure to prevent the default behavior. It still looks good however if you just attach an `onClick` handler to the button and call `e.preventDefault()`. See how we make use that method in other places in the codebase.

akshitac8:
Ohkay @helfi92  I will implement the changes.

PR #483, opened 2018-03-07
Bug1442237 Make Scope Grants have multiple views

blindacai:
Link to [Bug 1442237](https://bugzilla.mozilla.org/show_bug.cgi?id=1442237)

Some thoughts and concerns:
1. Features
* have 3 separate views for Scope Grants as stated in the bug description
* properly display view/delete page when users navigate to it by typing in url (I understand that we want to eliminate page reload in single page app, but we have no control over users' behavior I guess esp. when child urls here are not that long like other type of edit pages)
* derived from the above one: display an error when the page users search for does not exist
2. Testing
* I couldn't actually test delete/create grants part as it seems I don't have enough credentials
* Not sure how it will do when there are more supported scope patterns with more than 1 params and grants
3. Error handling
* Now when users search for a non-exist organization: /auth/grants/authorize-github-organizations/aaa, the errors displayed are `Cannot read property '[object Array]' of undefined` with method stacks, not quite user friendly. 
It's thrown by `this.props.auth.listRoles()` of which I couldn't find its implementation in this project and am curious where it is.

Also, as for variable/function namings, I am trying to follow the existing style, and do suggest better ones if you like. Thanks

### DIFF
--- a/src/views/ScopeGrants/ScopeGrants.jsx
+++ b/src/views/ScopeGrants/ScopeGrants.jsx
@@ -67,8 +67,18 @@ export default class ScopeGrants extends PureComponent {
 
   handleGrantsRedirect = () => this.props.history.push(`/auth/grants`);
   handleNewGrant = () => this.setState({ selected: null, args: {} });
+  select = org => {
+    const { history, pattern } = this.props;
 
-  renderGrants() {
+    if (org === 'create') {
+      this.handleNewGrant();
+      history.push(`/auth/grants/${pattern}/create`);
+    } else {
+      history.push(`/auth/grants/${pattern}/${encodeURIComponent(org)}`);
+    }
+  };
+
+  renderGrants(pattern) {
     if (this.state.error) {
       return <Error error={this.state.error} />;
     }
@@ -77,7 +87,6 @@ export default class ScopeGrants extends PureComponent {
       return <Spinner />;
     }
 
-    const pattern = PATTERNS.find(p => p.name === this.props.pattern);
     const instantiatedArgs = instantiatedArguments(pattern, this.state.roles);
     const params = Object.keys(pattern.params);
 
@@ -100,7 +109,7 @@ export default class ScopeGrants extends PureComponent {
           </Row>
           <hr />
           <Row>
-            <Col md={6}>
+            <Col md={12}>
               <Table hover>
                 <thead>
                   <tr>
@@ -111,15 +120,14 @@ export default class ScopeGrants extends PureComponent {
                 </thead>
                 <tbody>
                   {instantiatedArgs.map((args, index) => (
-                    <tr
-                      key={`args-${index}`}
-                      style={{ cursor: 'pointer' }}
-                      className={
-                        equals(args, this.state.selected) ? 'info' : null
-                      }
-                      onClick={() => this.setState({ selected: args })}>
+                    <tr key={`args-${index}`} style={{ cursor: 'pointer' }}>
                       {params.map((param, index) => (
-                        <td key={`params-${index}`}>
+                        <td
+                          key={`params-${index}`}
+                          onClick={() => {
+                            this.setState({ selected: args });
+                            this.select(args[param]);
+                          }}>
                           <code>{args[param]}</code>
                         </td>
                       ))}
@@ -128,16 +136,11 @@ export default class ScopeGrants extends PureComponent {
                 </tbody>
               </Table>
               <hr />
-              <Button onClick={this.handleNewGrant}>
+              <Button onClick={() => this.select('create')}>
                 <Glyphicon glyph="plus" /> New Grant
               </Button>
               <br />
               <br />
-            </Col>
-            <Col md={6}>
-              {this.state.selected
-                ? this.renderPatternInstance(pattern, this.state.selected)
-                : this.renderCreateForm(pattern)}
             </Col>
           </Row>
         </Col>
@@ -375,6 +378,9 @@ export default class ScopeGrants extends PureComponent {
   }
 
   render() {
+    const pattern = PATTERNS.find(p => p.name === this.props.pattern);
+    const { organization } = this.props;
+
     if (
       !this.props.pattern ||
       !PATTERNS.some(p => p.name === this.props.pattern)
@@ -387,6 +393,14 @@ export default class ScopeGrants extends PureComponent {
       );
     }
 
-    return this.renderGrants();
+    if (organization) {
+      if (organization === 'create') {
+        return this.renderCreateForm(pattern);
+      }
+
+      return this.renderPatternInstance(pattern, this.state.selected);
+    }
+
+    return this.renderGrants(pattern);
   }
 }
--- a/src/views/ScopeGrants/ScopeGrants.jsx
+++ b/src/views/ScopeGrants/ScopeGrants.jsx
@@ -67,12 +67,14 @@ export default class ScopeGrants extends PureComponent {
 
   handleGrantsRedirect = () => this.props.history.push(`/auth/grants`);
   handleNewGrant = () => this.setState({ selected: null, args: {} });
-  selectGrant = org => {
+  handleNavigate = org => {
     const { history, pattern } = this.props;
 
     if (org === 'create') {
       this.handleNewGrant();
       history.push(`/auth/grants/${pattern}/create`);
+    } else if (org === 'pattern') {
+      history.push(`/auth/grants/${pattern}`);
     } else {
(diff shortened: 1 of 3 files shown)


src/views/ScopeGrants/ScopeGrants.jsx
```diff
@@ -67,8 +67,20 @@ export default class ScopeGrants extends PureComponent {
...
+  handleNavigate = org => {
+    const { history, pattern } = this.props;
+
+    if (org === 'create') {
+      this.handleNewGrant();
+      history.push(`/auth/grants/${pattern}/create`);
+    } else if (org === 'pattern') {
+      history.push(`/auth/grants/${pattern}`);
+    } else {
+      history.push(`/auth/grants/${pattern}/${encodeURIComponent(org)}`);
```
helfi92: For routing, we use [`react-router`](https://reacttraining.com/react-router/web/guides/philosophy). It should look more like https://github.com/taskcluster/taskcluster-tools/blob/master/src/App/index.jsx#L235-L245.

src/views/ScopeGrants/ScopeGrants.jsx
```diff
@@ -128,17 +139,12 @@ export default class ScopeGrants extends PureComponent {
                 </tbody>
               </Table>
               <hr />
-              <Button onClick={this.handleNewGrant}>
+              <Button onClick={() => this.handleNavigate('create')}>
```
helfi92: For transitioning between views with `react-router`, you will most likely want to use the `Link` component. See https://github.com/taskcluster/taskcluster-tools/blob/e8b6d45f10e7520f717b7a9f5db87d550c74d15e/src/components/ScopeEditor/index.jsx#L94-L96 for an example.

blindacai: I removed `handleNavigate` function and replaced `history.push` with the `Link` component in 4 places. I like it this way too. 
Now the only 2 places with `history.replace` are at form `onComplete` handlers. Wish I could test create/delete grants locally

src/views/ScopeGrants/ScopeGrants.jsx
blindacai: I looked at the link you posted here.
For Scope Grants, we only have path: `/auth/grants/:pattern?/:organization?` mapped to one view component that includes all 3 views (list, create, delete)
The navigation seems to work fine for now. Also each view doesn't have too many elements in them and they need the same load function pre-mounting.
but is it still better to extract them to 3 separate view components like the workerManager?

src/views/ScopeGrants/ScopeGrants.jsx
```diff
@@ -358,23 +380,30 @@ export default class ScopeGrants extends PureComponent {
...
+                // key={`pattern-${index}`}
+                header={
+                  <h3>
+                    <Icon name={icon} /> {title}
+                  </h3>
+                }>
+                <Markdown>{description}</Markdown>
+              </ListGroupItem>
+            </Link>
           ))}
```
blindacai: I just noticed that after applying `<Link>` on `<ListGroupitem>`, we lose the background color change on mouse hovering. I tried to bring the effect back but haven't found a way.
The interesting thing is it won't affect `<Button>` on this effect

blindacai:
@helfi92 Some updates and comments are ready for review. Thanks

src/views/ScopeGrants/ScopeGrants.jsx
helfi92: Give [`LinkContainer`](https://github.com/react-bootstrap/react-router-bootstrap) a try. We use it in multiple places in the project.

blindacai: Indeed, how I missed it. Replaced 4 places with `LinkContainer` as they are all non-text bootstrap components.

src/views/ScopeGrants/ScopeGrants.jsx
```diff
@@ -183,6 +184,9 @@ export default class ScopeGrants extends PureComponent {
     const { args } = this.state;
 
     this.setState({ selected: null, roles: null, args: {} });
+    this.state.history.replace(
```
helfi92: This should be `this.props.history.replace`.

src/views/ScopeGrants/ScopeGrants.jsx
```diff
@@ -253,6 +257,7 @@ export default class ScopeGrants extends PureComponent {
 
   handleReload = () => {
     this.setState({ selected: null, roles: null });
+    this.state.history.replace(`/auth/grants/${this.props.pattern}`);
```
helfi92: `this.props.history.replace(...);`

blindacai: Ah, thanks for catching that. I guess the replace of history creates the illusion that it's changing like a state

src/views/ScopeGrants/ScopeGrants.jsx
```diff
@@ -387,6 +418,14 @@ export default class ScopeGrants extends PureComponent {
       );
     }
 
-    return this.renderGrants();
+    if (organization) {
+      if (organization === 'create') {
+        return this.renderCreateForm(pattern);
+      }
+
```
blindacai: As a note, I also had this concern that 'create' fills in the place as '/:organization' for quite a while when I started on this bug. If there's an organization in the future named 'create', it's not able to be edited. I hadn't thought of a better way. Maybe 'create' can be processed as special characters.

helfi92: Can we have `create` be defined as a separate view in `src/App/index.js`?

blindacai: Maybe we can.
At first all I thought was that
`<PropsRoute
    path="/auth/grants/:pattern?/create"`
`<PropRoute
    path="/auth/grants/:pattern?/:organization?`
makes no difference

But after looking at some existing routes, I think it can be:
`<PropsRoute
    path="/auth/grants/:pattern?/:organization?/edit"`
`<PropRoute
    path="/auth/grants/:pattern?/create`
so that '/create' can be sure to be mapped to a different view

blindacai: Actually, `<PropsRoute path="/auth/grants/:pattern?/:organization?/edit"` with those question marks will cause problems as '/pattern' and '/organization' cannot be mapped
maybe change it to something like:
grantscopes-routes

helfi92: I think `<PropRoute path="/auth/grants/:pattern?/create` will work :)

blindacai: Wait, you mean
`<PropsRoute path="/auth/grants/:pattern?/create"`
`<PropRoute path="/auth/grants/:pattern?/:organization?` ?
Doesn't it have the same issue that if an organization is named 'create', the user clicks on it will be taken to create new grants page instead of edit page?

eliperelman: Yes, it would still suffer from that same problem.

blindacai: Alright, I kinda get a feeling that I mixed two issues here. One is the anti-intuitive way of having `if (organization === 'create')` in the code which seems to be @helfi92 's original concern, and your suggestion is to have a different view component for `renderCreateForm`.
The other one is some peculiar organization named 'create'

blindacai: There seems some trade-offs here. A new view component to createGrants versus a not so good if conditional check. Hmm...
If you made the final decision of a separate view, I can try it.
I guess it also depends on how 'create new grant' page may evolve. If it gets more complex, you would want to separate it out too

helfi92: I think creating a new view component will be the "right" way to do it here. We also do it in other parts of the tools website (e.g., https://github.com/taskcluster/taskcluster-tools/blob/master/src/App/index.jsx#L251-L258). I'll let @eliperelman have the final verdict here :)

eliperelman: The most "correct" way to do this would be to make the scopes' top-level component also use a `Switch` from react-router:

```jsx
<Switch>
  <Route /* route for create  */ />
  <Route /* route for edit  */ />
  <Route /* route for list  */ />
</Switch>
```

But this may not be a trivial amount of work. Find a balance with what you are comfortable with and we can go from there.

eliperelman:
Apologies, but we are going to hold off on merging any PRs that have to deal with splitting views until we can ensure that they are done consistently and using the techniques that we lay out. Thank for your patience here.

PR #487, opened 2018-03-08
Add missing allowed characters in client ID

helfi92:


helfi92:
https://github.com/taskcluster/taskcluster-login/blob/master/schemas/constants.yml#L2

helfi92:
Users will not be able to create clients with the new tc-login changes deployed a couple of minutes ago. I will go ahead with the merge as this is kind of needed right now.

PR #489, opened 2018-03-08
Pool login credentials across tabs

eliperelman:
This upgrades taskcluster-client-web to use the new credentials-pooling feature, but creating a localStorage mutex against login credentials. I also upgraded auth0-js here to protect against the latest vulnerability.


djmitche:
I don't really know what "embedded login" means in Auth0, but it mentions hidden iframes in https://auth0.com/docs/libraries/auth0js/v9/migration-v8-v9 and we do use those for *renewals*.  So I'm not sure if we'll be affected.  But there's one sure way to find out :)

PR #497, opened 2018-03-13
Allow datestamps to be copied as ISO timestamps

ihsavru:
Fix: https://bugzilla.mozilla.org/show_bug.cgi?id=1439109

out-6

### DIFF
--- a/src/components/DateView/index.jsx
+++ b/src/components/DateView/index.jsx
@@ -8,21 +8,52 @@ export default class DateView extends PureComponent {
   constructor(props) {
     super(props);
     this.state = {
+      didCopy: false,
       id: v4()
     };
   }
 
+  registerChild = ref => {
+    this.textareaRef = ref;
+  };
+
+  handleMouseLeave = () => {
+    if (this.state.didCopy) {
+      this.setState({ didCopy: false });
+    }
+  };
+
+  handleCopyToClipboard = () => {
+    const textarea = this.textareaRef;
+
+    textarea.select();
+    document.execCommand('copy');
+    this.setState({ didCopy: true });
+  };
+
   render() {
-    const { date, since, placement, format } = this.props;
+    const { date, since, placement } = this.props;
     const tooltip = (
-      <Tooltip id={this.state.id}>{moment(date).format(format)}</Tooltip>
+      <Tooltip id={this.state.id}>
+        {moment(date).toISOString()}
+        <br />
+        {this.state.didCopy ? '(Copied!)' : '(Click to copy)'}
+      </Tooltip>
     );
 
     return (
       <OverlayTrigger placement={placement} overlay={tooltip}>
-        <span id={this.state.id}>
+        <span
+          id={this.state.id}
+          onClick={this.handleCopyToClipboard}
+          onMouseLeave={this.handleMouseLeave}>
           {moment(date).fromNow()}{' '}
           {since && `(${moment(date).from(since, true)} later)`}
+          <textarea
+            ref={this.registerChild}
+            style={{ width: 0, height: 0, opacity: 0 }}>
+            {moment(date).toISOString()}
+          </textarea>
         </span>
       </OverlayTrigger>
     );
@@ -37,6 +68,5 @@ DateView.propTypes = {
 };
 
 DateView.defaultProps = {
-  format: 'Do of MMMM YYYY, H:mm:ss',
   placement: 'top'
 };


src/components/DateView/index.jsx
```diff
@@ -8,19 +8,52 @@ export default class DateView extends PureComponent {
   constructor(props) {
     super(props);
     this.state = {
+      copy_status: '(Click to copy)',
```
eliperelman: Use camelCasing for names:

```js
copyStatus: '(Click to copy)',
```

src/components/DateView/index.jsx
```diff
@@ -8,19 +8,52 @@ export default class DateView extends PureComponent {
...
       id: v4()
     };
+    this.copyToClipboard = this.copyToClipboard.bind(this);
```
eliperelman: You don't need to create these bindings since we are using class fields for events. Simply declare the event handler as an arrow function within the class. See other comment for specifics.

src/components/DateView/index.jsx
```diff
@@ -8,19 +8,52 @@ export default class DateView extends PureComponent {
...
+    this.handleMouseLeave = this.handleMouseLeave.bind(this);
+  }
+
+  handleMouseLeave() {
+    this.setState({ copy_status: '(Click to copy)' });
+  }
+
+  copyToClipboard(date) {
+    const textarea = document.createElement('textarea');
```
eliperelman: Can we keep a hidden textarea in the rendering instead of generating it every time? Then maybe reference it with a `ref`?

```jsx
import { findDOMNode } from 'react-dom';

registerChild = (ref) => {
  this.textareaRef = ref;
};

handleCopyToClipboard = () => {
  const textarea = findDOMNode(this.textareaRef);

  textarea.select();
  document.execCommand('copy');
  this.setState({ copyStatus: '(Copied!)' });
};

<OverlayTrigger placement={placement} overlay={tooltip}>
  <textarea
    ref={this.registerChild}
    style={{ display: 'none' }}>
    {moment(date).toISOString()}
  </textarea>
```

Do you think something like this would work?

src/components/DateView/index.jsx
```diff
@@ -8,19 +8,52 @@ export default class DateView extends PureComponent {
...
+        {this.state.copy_status}
+      </Tooltip>
     );
 
     return (
       <OverlayTrigger placement={placement} overlay={tooltip}>
-        <span id={this.state.id}>
+        <span
+          id={this.state.id}
+          onClick={() => {
```
eliperelman: Pass a reference to the event handler directly, when it is declared as an arrow function:

```jsx
onClick={this.handleCopyToClipboard}
onMouseLeave={this.handleMouseLeave}
```

src/components/DateView/index.jsx
```diff
@@ -8,19 +8,52 @@ export default class DateView extends PureComponent {
...
     super(props);
     this.state = {
+      copy_status: '(Click to copy)',
       id: v4()
     };
+    this.copyToClipboard = this.copyToClipboard.bind(this);
+    this.handleMouseLeave = this.handleMouseLeave.bind(this);
+  }
+
+  handleMouseLeave() {
```
eliperelman: Declare as an arrow function so you can skip binding and pass as a reference directly to event props:

```js
handleMouseLeave = () => {
  this.setState({ copyStatus: '(Click to copy)' });
};
```

ihsavru:
@eliperelman made the requested changes :)

src/components/DateView/index.jsx
```diff
@@ -8,21 +8,50 @@ export default class DateView extends PureComponent {
...
       id: v4()
     };
   }
 
+  registerChild = ref => {
+    this.textareaRef = ref;
+  };
+
+  handleMouseLeave = () => {
+    this.setState({ didCopy: false });
```
eliperelman: Just to keep the component from unnecessarily re-rendering, could you guard this by checking if `didCopy` is already `true`?

```js
handleMouseLeave = () => {
  if (this.state.didCopy) {
    this.setState({ didCopy: false });
  }
};
```

src/components/DateView/index.jsx
```diff
@@ -8,21 +8,50 @@ export default class DateView extends PureComponent {
   constructor(props) {
     super(props);
     this.state = {
+      didCopy: false,
       id: v4()
     };
   }
 
+  registerChild = ref => {
```
eliperelman: You declared this `ref` handler, but didn't use it in the JSX in `render`.

src/components/DateView/index.jsx
```diff
@@ -8,21 +8,50 @@ export default class DateView extends PureComponent {
...
+    this.textareaRef = ref;
+  };
+
+  handleMouseLeave = () => {
+    this.setState({ didCopy: false });
+  };
+
+  handleCopyToClipboard = () => {
+    const textarea = this.node;
```
eliperelman: This should be a reference to the `ref` set in `registerChild`, `this.textareaRef`.

src/components/DateView/index.jsx
```diff
@@ -8,21 +8,50 @@ export default class DateView extends PureComponent {
...
       <OverlayTrigger placement={placement} overlay={tooltip}>
-        <span id={this.state.id}>
+        <span
+          id={this.state.id}
+          onClick={this.handleCopyToClipboard}
+          onMouseLeave={this.handleMouseLeave}>
           {moment(date).fromNow()}{' '}
           {since && `(${moment(date).from(since, true)} later)`}
+          <textarea
+            ref={node => (this.node = node)}
```
eliperelman: Here is where you can pass the `ref` handler:

```jsx
ref={this.registerChild}
```

src/components/DateView/index.jsx
ihsavru: I removed this function because it isn't needed now. Previously, I was using it together with findDOMNode but findDOMNode was raising an error.

ihsavru:
@eliperelman made further changes

src/components/DateView/index.jsx
eliperelman: Technically you're correct; it isn't *necessary*, but I would still like to see this in its own method. Please move the `node => this.node = node` to its own method called `registerChild`:

```js
registerChild = (ref) => {
  this.textareaRef = ref;
};
```

ihsavru:
@eliperelman Thankyou for your time in reviewing this pull request, I've made some final changes to extract the ref handler to a class function.

eliperelman:
Great job, thanks for your patience here!

PR #498, opened 2018-03-13
Algolia docs search in navigation

eliperelman:
screen shot 2018-03-13 at 1 39 56 pm

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,6 @@
+ALGOLIA_APP_ID="latency"
+ALGOLIA_API_KEY="a85bf4b3df84bee712cefa9e60046a80"
+ALGOLIA_INDEX_NAME="taskcluster"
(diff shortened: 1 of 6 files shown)


djmitche:
Hm, should we think twice about running Algolia's (hosted?) JS in a context where we have LocalStorage full of powerful credentials?  I don't know how best to evaluate that..

eliperelman:
Right now trying to get feature parity with docs, so if you're cool with not having docs search, we can delete this.

djmitche:
It seems safer - @imbstack, thoughts?

imbstack:
Yeah, let's remove it for now I guess. It was never all that useful and we can add it back later if we want.

eliperelman:
Works for me.

PR #501, opened 2018-03-14
Load docs from markdown files

eliperelman:
@helfi92 this is still buggy, but wanted to get you something that you could build off of for loading markdown. If you could take a look at the routing, that would be helpful. 😃


eliperelman:
This works pretty good now, so I'm going to land this and try copy-pasting the other directories now.

PR #504, opened 2018-03-15
Add reference manual to docs

helfi92:


src/docs/ReferenceSidebar.jsx
```diff
@@ -0,0 +1,850 @@
...
+          <ul>
+            <li>
+              <Link exact to="/docs/reference/platform/taskcluster-auth/">
+                taskcluster-auth
+              </Link>
+              <div className="inline-navs">
+                (
+                <Link
+                  exact
+                  to="/docs/reference/platform/taskcluster-auth/references/api">
```
helfi92: Is that a directory or a page? If it's the former, then I will need to add a trailing slash.

PR #505, opened 2018-03-15
Make docs responsive

helfi92:
- Add hamburger menu when using a small screen
- Make docs mobile friendly
- Alter line length for better text legibility

### DIFF
--- a/src/docs/index.jsx
+++ b/src/docs/index.jsx
@@ -1,21 +1,19 @@
 import { Component } from 'react';
 import { Link } from 'react-router-dom';
 import resolve from 'resolve-pathname';
-import { Grid, Row, Col, Button, Collapse } from 'react-bootstrap';
+import { Grid, Row, Col } from 'react-bootstrap';
 import 'prismjs/themes/prism.css';
 import HelmetTitle from '../components/HelmetTitle';
 import Error from '../components/Error';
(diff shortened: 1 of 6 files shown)


src/docs/index.jsx
```diff
@@ -61,13 +63,43 @@ export default class Documentation extends Component {
     );
   };
 
+  handleMenuClick = () =>
+    this.setState({ showSidebar: !this.state.showSidebar });
+
   renderSidebar() {
     const { pathname } = this.props.location;
+    let TableOfContent;
```
eliperelman: This can be a `const` with a ternary. Rename to `TableOfContents`.

eliperelman: Also, let's move it to a separate component now. 🤣

src/docs/index.jsx
```diff
@@ -61,13 +63,43 @@ export default class Documentation extends Component {
...
 
     if (pathname.includes('/docs/manual')) {
-      return <ManualSidebar />;
+      TableOfContent = ManualSidebar;
     } else if (pathname.includes('/docs/reference')) {
-      return <ReferenceSidebar />;
+      TableOfContent = ReferenceSidebar;
+    }
+
+    if (TableOfContent) {
```
eliperelman: We can put this whole method in a separate component.

PR #506, opened 2018-03-16
Bug 1441966:  Add ability to filter through the list of secrets

akshitac8:
This PR is in reference to [Bug 1441966](https://bugzilla.mozilla.org/show_bug.cgi?id=1441966).
- Added Search field


src/views/SecretsManager/SearchForm.jsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = { value: '' };
+  }
+
+  componentWillMount() {
+    if (this.props.default) {
```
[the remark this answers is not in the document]
eliperelman: So reading the parent component, it appears that this value is also controlled by the parent state. That means you can remove all state handling in this component and delegate it to the parent by making `handleSubmit` call `onSearch`, and passing a new `onChange` function.

src/views/SecretsManager/SearchForm.jsx
```diff
@@ -0,0 +1,62 @@
+import { PureComponent } from 'react';
```
helfi92: I see that this is the same file as ` src/views/RoleManager/SearchForm.jsx`. Could we create `src/components/SearchForm/index.jsx` and `src/components/SearchForm/styles.module.css` and have the RoleManager and SecretsManager use that instead?

src/views/SecretsManager/SecretsManager.jsx
```diff
@@ -104,6 +117,7 @@ export default class SecretsManager extends PureComponent {
       <Row>
         <HelmetTitle title="Secrets Manager" />
         <Col md={5}>
+          {this.renderTypeInput()}
```
helfi92: To make the search form work, you will need to add a filter method in `renderSecretsTable` similar to https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/RoleManager/RoleManager.jsx#L96.

akshitac8:
Sorry for the inconvenience,
I was working on another issue within the same repo and I forgot that this one is still under review.

akshitac8:
@helfi92 @eliperelman Implemented the changes:)

src/components/SearchForm/index.jsx
```diff
@@ -1,53 +1,52 @@
...
 
   render() {
     return (
       <div>
-        <Form onSubmit={this.handleSubmit} className={styles.typeInput}>
+        <Form onSubmit={this.handleSubmit} className={typeInput}>
           <FormGroup>
             <InputGroup bsSize="sm">
-              <InputGroup.Addon>Role Id Containing</InputGroup.Addon>
+              <InputGroup.Addon>Search</InputGroup.Addon>
```
helfi92: Let's have this as a prop `label`. That way we can change the label based on where we render it:

```js
static propTypes = {
  onSearch: func.isRequired,
  label: string
};

static defaultProps = {
  label: 'String Containing'
};
```

Then in RoleManger:

```js
<SearchForm label="Role Id Containing" ... />
```

src/views/RoleManager/RoleManager.jsx
```diff
@@ -1,4 +1,4 @@
-import { PureComponent } from 'react';
+import { Component } from 'react';
```
helfi92: Any reason why we are changing this to `Component`?

src/views/SecretsManager/SecretsManager.jsx
```diff
@@ -80,22 +82,26 @@ export default class SecretsManager extends PureComponent {
...
-              key={`secret-row-${index}`}
-              className={secretId === id ? 'info' : null}>
-              <td>
-                <Link to={`/secrets/${encodeURIComponent(id)}`}>
-                  <code>{id}</code>
-                </Link>
-              </td>
-            </tr>
-          ))}
+          {this.state.secrets
```
helfi92: `secrets` is defined above so we can just use `secrets` instead of `this.state.secrets`.

src/views/SecretsManager/SecretsManager.jsx
```diff
@@ -104,6 +110,10 @@ export default class SecretsManager extends PureComponent {
       <Row>
         <HelmetTitle title="Secrets Manager" />
         <Col md={5}>
+          <SearchForm
+            default={this.state.secretIdContains}
```
helfi92: I don't see `default` defined in `SearchForm`. Am I missing something? :)

akshitac8:
@helfi92  Implemented further changes:)

src/components/SearchForm/index.jsx
```diff
@@ -1,53 +1,52 @@
-import { PureComponent } from 'react';
+import { Component } from 'react';
```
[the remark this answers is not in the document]
eliperelman: I think it's better to err on the side of working vs. strange side effects, so for contributor code I don't mind using Component, since it will always render, but maybe not the most efficiently.

That said, we should probably only be using PureComponent when props and state are shallow, or we only need referential equality.

src/components/SearchForm/index.jsx
```diff
@@ -1,53 +1,57 @@
...
 } from 'react-bootstrap';
-import styles from './styles.module.css';
+import { typeInput } from './styles.module.css';
 
-export default class SearchForm extends PureComponent {
+export default class SearchForm extends Component {
   static propTypes = {
-    onSearch: func.isRequired
+    onSearch: func.isRequired,
+    label: String
```
eliperelman: This should be `string` from `prop-types`, not the `String` constructor.

src/components/SearchForm/index.jsx
```diff
@@ -1,53 +1,57 @@
...
   };
 
-  constructor(props) {
-    super(props);
+  static defaultProps = {
+    label: 'String Containing'
```
eliperelman: Some users may not know what a "string" is, so make this "value" containing.

src/views/RoleManager/RoleManager.jsx
eliperelman: I think this one is OK since `roles` is an array.

akshitac8:
@eliperelman I have done the changes as told :smiley:

akshitac8:
Are there any further changes to be made?

src/views/SecretsManager/SecretsManager.jsx
```diff
@@ -80,22 +82,26 @@ export default class SecretsManager extends PureComponent {
...
-                  <code>{id}</code>
-                </Link>
-              </td>
-            </tr>
-          ))}
+          {secrets
+            .filter(id => id.includes(this.state.secretIdContains))
+            .map((id, index) => (
+              <tr
+                key={`secret-row-${index}`}
```
eliperelman: Since we have added a `filter` into the mix, using the `index` in the key can produce unwanted rendering results, since `secret-row-1` may be filtered away and be re-used by another `index`. This should be driven by the ID.

src/components/SearchForm/index.jsx
```diff
@@ -1,53 +1,57 @@
...
-                onChange={this.handleChange}
                 value={this.state.value}
+                onChange={this.handleChange}
                 type="text"
               />
+              <Button
+                type="submit"
+                style={{
+                  visibility: 'hidden',
+                  display: 'none'
```
eliperelman: I *think* using `display: none` makes this submit trick not work properly in some browsers, hence why we were using `visibility: hidden`.

akshitac8: Yeah, Actually we don't even need this. Thank you for the correction :)

akshitac8:
@eliperelman  Implemented further changes:)

src/views/SecretsManager/SecretsManager.jsx
```diff
@@ -80,22 +82,24 @@ export default class SecretsManager extends PureComponent {
...
-                <Link to={`/secrets/${encodeURIComponent(id)}`}>
-                  <code>{id}</code>
-                </Link>
-              </td>
-            </tr>
-          ))}
+          {secrets
+            .filter(id => id.includes(this.state.secretIdContains))
+            .map(id => (
+              <tr key={id} className={secretId === id ? 'info' : null}>
```
eliperelman: Can we put the `secret-row` portion back in?

```js
<tr key={`secret-row-${id}`}
```

akshitac8:
Hello,
sorry for asking again. Is this one done?

eliperelman:
Hello, I was waiting for @helfi92 to return so we could give this another pass.

akshitac8:
Ohkay @eliperelman 
Thank you :)

src/views/SecretsManager/SecretsManager.jsx
```diff
@@ -80,22 +82,26 @@ export default class SecretsManager extends PureComponent {
...
+                  </Link>
+                </td>
+              </tr>
+            ))}
         </tbody>
       </Table>
     );
   }
 
+  handleSetSecretId = value => this.setState({ secretIdContains: value });
```
helfi92: Can we rename this to `handleSearchChange`?

src/components/SearchForm/index.jsx
```diff
@@ -1,51 +1,47 @@
...
 
-  componentWillMount() {
-    if (this.props.default) {
-      this.setState({ value: this.props.default });
-    }
-  }
+  state = { value: '' };
 
-  handleChange = e => this.setState({ value: e.target.value });
+  handleChange = e => {
```
helfi92: I would rename this to `handleSearchChange` to make it more explicit what is changing.

akshitac8:
Hello @helfi92 changed name in the following files :)

src/components/SearchForm/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+    label: string
+  };
+
+  static defaultProps = {
+    label: 'Value Containing'
+  };
+
+  state = { value: '' };
+
+  handleChange = e => {
```
helfi92: This should be `handleSearchChange`.

src/components/SearchForm/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+    this.setState({ value: e.target.value });
+  };
+
+  handleSearchChange = () => this.props.onSearch(this.state.value);
```
helfi92: This should be `handleSearch`.

src/components/SearchForm/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+  render() {
+    return (
+      <div>
+        <Form onSubmit={this.handleSubmit} className={typeInput}>
+          <FormGroup>
+            <InputGroup bsSize="sm">
+              <InputGroup.Addon>{this.props.label}</InputGroup.Addon>
+              <FormControl
+                value={this.state.value}
+                onChange={this.handleChange}
```
helfi92: `this.handleSearchChange`

src/components/SearchForm/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+                type="text"
+              />
+              <InputGroup.Addon onClick={this.handleSearchChange} role="button">
```
helfi92: `this.handleSearch`

helfi92:
Also, travis is failing. You can try `yarn lint --fix` or `npm run lint --fix` to see if that makes travis happy.

akshitac8:
Done some changes :)
But was not able to make Travis happy :slightly_frowning_face:

akshitac8:
Thank you for the approval :)

PR #507, opened 2018-03-18
Bug1382594 :  "Follow the log" should follow after completion

akshitac8:
This PR is in reference with the issue [Bug 1382594](https://bugzilla.mozilla.org/show_bug.cgi?id=1382594)
Added else case to automatic  check the checkbox when scrolled back down.
Thank you:)

### DIFF
--- a/src/views/RoleManager/RoleManager.jsx
+++ b/src/views/RoleManager/RoleManager.jsx
@@ -12,6 +12,7 @@ import equal from 'deep-equal';
 import Spinner from '../../components/Spinner';
 import RoleEditor from '../../components/RoleEditor';
 import HelmetTitle from '../../components/HelmetTitle';
+import SearchForm from './SearchForm';
 
 export default class RoleManager extends PureComponent {
   constructor(props) {
@@ -19,7 +20,8 @@ export default class RoleManager extends PureComponent {
 
     this.state = {
       roles: null,
-      error: null
+      error: null,
+      roleIdContains: ''
     };
   }
 
@@ -73,6 +75,8 @@ export default class RoleManager extends PureComponent {
     );
   };
 
+  handleSetRoleId = value => this.setState({ roleIdContains: value });
+
   renderRolesTable() {
     const { roles } = this.state;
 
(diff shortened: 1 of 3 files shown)


src/views/UnifiedInspector/LogView.jsx
```diff
@@ -115,6 +115,11 @@ export default class LogView extends PureComponent {
   handleScroll = ({ scrollTop, scrollHeight, clientHeight }) => {
     if (this.state.follow && scrollHeight - scrollTop !== clientHeight) {
       this.setState({ follow: false });
+    } else if (
+      !this.state.follow &&
+      scrollHeight - scrollTop === clientHeight
+    ) {
+      this.setState({ follow: true });
```
djmitche: It looks like this will automatically start following if you scroll to the bottom of the log.  That's an interesting idea for a feature, but not what this particular bug is about.

akshitac8:
The changes made will keep the checkbox check when scrolled at the beginning of the log and checks it back again when reached the bottom of the log.
Can I get little suggestion on how to proceed further with the bug?
Thank you:)

djmitche:
You can see the bug in action by using the Task Creator tool to create a new task that prints some output for a while before exiting -- maybe it prints "hello" and the time every second for 600 seconds.

When you run that task, you can follow the log and see the latest messages.  But once the task exits, you'll see the log "reload" and show you the beginning of the logfile, not the end.  That is the thing to fix.

akshitac8:
Hello @djmitche 
I understood the bug and tried various techniques like componentwillupdate() just before componentdidupdate() to save the scroll position and also change calculation for handylazyviewerheight() but then added this modified. Is this the required change needed or I am moving in the wrong direction?
Thank you:)

akshitac8:
Hello @djmitche 
I have implemented few changes:)

akshitac8:
Hello @eliperelman 
Yes, There was but now I have created another branch and rebased the code according to the master.
sorry for the inconvenience :slightly_frowning_face:

eliperelman:
Hmm, I'm still not seeing the rebase here.

akshitac8:
There was some proper with rebase. but now that is sorted and created a new PR for more clarity.
Thank you:)

Issue #475, opened 2018-03-02
Support provisioning health and error reporting in aws-provisioner tools pages


jhford:
We've talked about this in an email thread, but I wanted to get the issue on file.  The supporting work in the EC2-Manager service has been completed, awaiting deployment to our production instance.  All URLs are pointing at our staging instance.  I'm also working on figuring out why we're not getting API references to publish for the EC2-Manager.  These are the changes that I think we should make:

1. Split worker type details into its own page
1. On the overall view of the AWS-Provisioner page, have a table that shows the latest errors encountered by this provisioner.  The supporting API is [this](https://ec2-manager-staging.taskcluster.net/v1/errors)
1. On the overall view of the AWS-Provisioner page, have a table that shows the 'health' of each region/az/instanceType.  The supporting API is [this](https://ec2-manager-staging.taskcluster.net/v1/health).  The values in this API response are approximate stats on the configuration.  How they're displayed isn't critical to me, but I suspect highlighting rows which have non-zero error values as warnings and rows where those counts are >40% of the success rate as errors would be useful.  Ideally, it could be viewed by Region, AZ and Instance Type.  The idea is to be able to get an overview of an entire Region, AZ or Instance Type.
1. On the worker type detail, add a new tab to the view called 'Health' that contains the Errors for that worker type.  The supporting API is [this](https://ec2-manager-staging.taskcluster.net/v1/worker-types/gecko-1-b-linux/errors).  Basically, a worker type specific view of the other errors page
1. On the worker type 'Health' tab, add a worker type specific overview of Region, AZ and Instance Type.  The supporting API is [this](https://ec2-manager-staging.taskcluster.net/v1/worker-types/gecko-1-b-linux/health)

This will allow Sheriffs and others interested to see what is going on inside the provisioner.  This should reduce the impact on the Taskcluster team when there are problems inside the EC2 system as well as highlight when worker types are incorrectly configured.

whimboo:
Those details would be great to see. We just had a situation when lots of workers didn't get provisioned and it was not that clear what's going on by just looking at https://tools.taskcluster.net/aws-provisioner/. So links to health and error pages would be very helpful. Thanks!

grenade:
+1 yes please! this will be super useful for diagnosing windows/occ ami issues too.

petemoore:
[...]

PR #508, opened 2018-03-20
Add aws-provisioner recent errors page

helfi92:
This should cover the second point of https://github.com/taskcluster/taskcluster-tools/issues/475#issue-301802685.

### DIFF
--- a/src/views/AwsProvisioner/WorkerTypeTable.jsx
+++ b/src/views/AwsProvisioner/WorkerTypeTable.jsx
@@ -255,9 +255,8 @@ export default class WorkerTypeTable extends PureComponent {
 
   handleRecentErrorsClick = () => {
     const { push, location } = this.props.history;
-    const normalize = path => path.replace('//', '/');
 
-    push(normalize(`${location.pathname}/recent-errors`));
+    push(`${location.pathname.replace('//', '/')}/recent-errors`);
   };
 
   render() {
--- a/src/views/AwsProvisioner/WorkerTypeTable.jsx
+++ b/src/views/AwsProvisioner/WorkerTypeTable.jsx
@@ -7,7 +7,14 @@ import WorkerTypeRow from './WorkerTypeRow';
 import Spinner from '../../components/Spinner';
 import Error from '../../components/Error';
 import HelmetTitle from '../../components/HelmetTitle';
-import { workerTypes } from './styles.module.css';
+import {
+  workerTypes,
+  workerTypeViewContainer,
+  workerTypeCreatorContainer,
+  stateTable,
+  createWorkerTypeButton,
+  recentErrorsButton
+} from './styles.module.css';
 import UserSession from '../../auth/UserSession';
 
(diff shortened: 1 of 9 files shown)


src/views/AwsProvisioner/WorkerTypeTable.jsx
```diff
@@ -246,6 +253,13 @@ export default class WorkerTypeTable extends PureComponent {
     );
   }
 
+  handleRecentErrorsClick = () => {
+    const { push, location } = this.props.history;
+    const normalize = path => path.replace('//', '/');
+
+    push(normalize(`${location.pathname}/recent-errors`));
```
eliperelman: Why not just:

```js
push(`${location.pathname.replace('//', '/')}/recent-errors`);
```

eliperelman: Does `match.url` give us a more normalized location?

src/views/AwsProvisionerErrors/AwsProvisionerErrors.jsx
```diff
@@ -0,0 +1,132 @@
...
+              <th>Type</th>
+              <th>Instance Type</th>
+              <th>Code</th>
+              <th>Region</th>
+              <th>Time</th>
+              <th>Message</th>
+            </tr>
+          </thead>
+          <tbody>
+            {recentErrors &&
```
eliperelman: `recentErrors` is guaranteed to exist because of the check on L85.

src/views/AwsProvisionerErrors/AwsProvisionerErrors.jsx
```diff
@@ -0,0 +1,132 @@
...
+    if (sortBy === 'Time') {
+      return moment(a.time).diff(moment(b.time)) < 0 ? 1 : -1;
+    }
+
+    return a[camelCase(sortBy)].localeCompare(b[camelCase(sortBy)]);
+  };
+
+  renderRecentError(error, key) {
+    return (
+      <tr key={`recent-error-${key}`}>
```
eliperelman: Since we are sorting this array, using `index` as `key` can give us incorrect renderings. Do the errors have IDs, or can we use a `time` or a combination of `time` and some other property as the key?

helfi92: I used a combination of `time`, `type` and `workerType`. The browser didn't complain about duplicate keys.

src/views/AwsProvisionerErrors/AwsProvisionerErrors.jsx
helfi92: Agreed.

src/views/AwsProvisioner/WorkerTypeTable.jsx
helfi92: `match.url` doesn't give us a more normalized location, but I like what you suggested in your first comment :)

PR #509, opened 2018-03-22
Bug1382594 :  "Follow the log" should follow after completion

akshitac8:
This PR is in reference to the issue [Bug 1382594](https://bugzilla.mozilla.org/show_bug.cgi?id=1382594) and closed [PR](https://github.com/taskcluster/taskcluster-tools/pull/507)
Thank you:)

### DIFF
--- a/src/views/UnifiedInspector/LogView.jsx
+++ b/src/views/UnifiedInspector/LogView.jsx
@@ -43,7 +43,7 @@ export default class LogView extends PureComponent {
     fscreen.addEventListener(
       'fullscreenchange',
       this.handleFullscreenChange,
-      true
+      false
     );
     window.addEventListener('resize', this.handleLazyViewerHeight);
   }
@@ -69,10 +69,8 @@ export default class LogView extends PureComponent {
         this.props.log &&
         nextProps.log.name !== this.props.log.name)
     ) {
-      // const streaming = this.isStreaming(nextProps.status);
-
       this.setState({
-        follow: this.prefersFollow(),
+        follow: true,
         isFullscreen: false
       });
     }
@@ -133,9 +131,8 @@ export default class LogView extends PureComponent {
   handleLazyViewerHeight = () => {
     if (this.lazylog) {
       const lazyViewerHeight =
-        window.scrollHeight -
-          Math.round(this.lazylog.getBoundingClientRect().top) ===
-        window.clientHeight;
+        window.innerHeight -
+        Math.round(this.lazylog.getBoundingClientRect().top);
 
       if (lazyViewerHeight > VIEWER_HEIGHT_MIN) {
         this.setState({ lazyViewerHeight });
--- a/src/views/UnifiedInspector/LogView.jsx
+++ b/src/views/UnifiedInspector/LogView.jsx
@@ -39,6 +39,7 @@ export default class LogView extends PureComponent {
       lineNumber: ''
     };
   }
+
   componentWillMount() {
     fscreen.addEventListener(
       'fullscreenchange',
@@ -69,8 +70,11 @@ export default class LogView extends PureComponent {
         this.props.log &&
         nextProps.log.name !== this.props.log.name)
     ) {
+      const streaming = this.isStreaming(nextProps.status);
+
       this.setState({
-        follow: true,
+        streaming,
+        follow: streaming || this.prefersFollow(),
         isFullscreen: false
       });
     }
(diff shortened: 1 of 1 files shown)


src/views/UnifiedInspector/LogView.jsx
```diff
@@ -39,12 +39,11 @@ export default class LogView extends PureComponent {
       lineNumber: ''
     };
   }
-
```
eliperelman: Add this blank line back in.

src/views/UnifiedInspector/LogView.jsx
```diff
@@ -39,12 +39,11 @@ export default class LogView extends PureComponent {
...
   componentWillMount() {
     fscreen.addEventListener(
       'fullscreenchange',
       this.handleFullscreenChange,
-      false
+      true
```
eliperelman: Can you explain why this was changed?

src/views/UnifiedInspector/LogView.jsx
```diff
@@ -70,11 +69,10 @@ export default class LogView extends PureComponent {
         this.props.log &&
         nextProps.log.name !== this.props.log.name)
     ) {
-      const streaming = this.isStreaming(nextProps.status);
+      // const streaming = this.isStreaming(nextProps.status);
```
eliperelman: Please remove commented code from commits.

Can you explain why we don't want to check if this is a streaming log?

src/views/UnifiedInspector/LogView.jsx
```diff
@@ -130,8 +133,9 @@ export default class LogView extends PureComponent {
   handleLazyViewerHeight = () => {
     if (this.lazylog) {
       const lazyViewerHeight =
-        window.innerHeight -
-        Math.round(this.lazylog.getBoundingClientRect().top);
+        window.scrollHeight -
```
eliperelman: This statement confuses me. Before this returned a numeric value, and now it's evaluating to be a boolean. Can you explain?

src/views/UnifiedInspector/LogView.jsx
```diff
@@ -115,6 +113,11 @@ export default class LogView extends PureComponent {
   handleScroll = ({ scrollTop, scrollHeight, clientHeight }) => {
     if (this.state.follow && scrollHeight - scrollTop !== clientHeight) {
       this.setState({ follow: false });
+    } else if (
+      !this.state.follow &&
```
eliperelman: Can you explain how switching from checking `follow` to `!follow` fixes this bug?

akshitac8:
hello @eliperelman Removed few unnecessary changes:)

src/views/UnifiedInspector/LogView.jsx
```diff
@@ -70,11 +69,8 @@ export default class LogView extends PureComponent {
         this.props.log &&
         nextProps.log.name !== this.props.log.name)
     ) {
-      const streaming = this.isStreaming(nextProps.status);
-
       this.setState({
-        streaming,
```
eliperelman: This now never checks if we are using a streaming log, and always follows. This is not the intent of the bug.

src/views/UnifiedInspector/LogView.jsx
eliperelman: I am looking for a resolution for the bug. Extra features should be in their own PR.

src/views/UnifiedInspector/LogView.jsx
akshitac8: I thought If I could stop the log from reloading it will stay at its last position. I will correct the changes...

akshitac8:
Hello @eliperelman I have added the changes. The changes made now make the log go to the last line after reload in both the cases:
- [x] Follow log
- [ ] Follow log  

I wanted to ask, should it behave differently for the second case, like stay at the same position before and after reload.

src/views/UnifiedInspector/LogView.jsx
```diff
@@ -87,10 +87,15 @@ export default class LogView extends PureComponent {
       // eslint-disable-next-line react/no-did-update-set-state
       this.setState({ jump: false });
     }
+
+    if (this.state.follow) {
```
eliperelman: This section doesn't do anything. If `this.state.follow` is true, you just set it to true again.

src/views/UnifiedInspector/LogView.jsx
```diff
@@ -87,10 +87,15 @@ export default class LogView extends PureComponent {
...
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ follow: true });
+    }
   }
 
   prefersFollow() {
-    return localStorage.getItem('follow-log') === 'true';
+    return localStorage.getItem('follow-log') === 'true' ? ' ' : 'true';
```
eliperelman: This used to return a boolean but now returns a string?

helfi92:
Closing for inactivity.

PR #513, opened 2018-04-01
Add a diff view before saving changes in the ScopeEditor

MynahMarie:
In this PR I have:

- added a 4th button to the `ButtonToolbar` which toggles between "**Review**" and "**Edit**"
- created a new rendering function that shows the changes the user made if he/she wishes to review them before saving.
- created a customized version of the `react-diff` module in order to create a component that renders those changes
- created a new directory in the project's root to store the custom `react-diff` module I've created

Relates Bugzilla 1306494


custom_modules/react-diff-custom/dist/react-diff-custom.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
```
eliperelman: We already have a `components` directory inside `src`, and this can go there. Some things to change:

- Use ES `import` and `export` instead of CommonJS `require` and `module.exports`.
- Use `const` over `var`, which the linter will enforce once this is moved into `src`.
- `React.createClass` isn't used in this codebase, as ES classes are used. See all other components in this repo for examples.
- React isn't going to have `PropTypes` on it, as we are importing the `prop-types` package.
- Use JSX.

custom_modules/react-diff-custom/dist/react-diff-custom.js
```diff
@@ -0,0 +1,59 @@
...
+
+var React = require('react');
```
eliperelman: What is different about this file from the original that couldn't be fixed via different props?

custom_modules/react-diff-custom/package.json
```diff
@@ -0,0 +1,39 @@
+{
```
eliperelman: This file isn't needed once the component lives in `src`.

custom_modules/react-diff-custom/dist/react-diff-custom.js
djmitche: Another option here is to try to contribute these changes to the upstream package or, if that doesn't work or the changes are so massive that it doesn't make sense, create a new package on npm for this.

MynahMarie: @eliperelman So the first thing I tried was to add it in the src folder but there was 1 linting error I just didn't manage to get rid of... I will try again with more intense re-writing and see if I can get it to work.

@djmitche yes I thought about that also... I looked at the react-diff project's repo and not much activity is happening there and there is an issue of the owner saying that he is looking for someone to maintain it. I think I will try and see if I can make it work as Eli suggested and if that doesn't work I will try to contact the owner of the module directly.

custom_modules/react-diff-custom/dist/react-diff-custom.js
MynahMarie: The only things I changed from the original file was to change the styling and to add the option of using the `diffLines` method from the `diff` module. Using `diffLines` allows to pass it the `newlineIsToken: true` option which I felt was the most appropriate one for rendering the kind of diff we needed...

MynahMarie:
@eliperelman I moved the `react-diff-custom` to the components folder in `src` and deleted the `package.json` associated to it. I also refactored using `jsx` and ES6 syntax, please let me know if this is an improvement and if there is any other changes I should do.

src/components/ScopeEditor/index.jsx
```diff
@@ -8,12 +8,14 @@ import 'codemirror/addon/mode/simple';
 import CodeEditor from '../CodeEditor';
 import './scopemode';
 import './styles.css';
+import Diff from './react-diff-custom';
```
eliperelman: For consistency, this file should be in PascalCase like all other components. You can just name it `Diff`.

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
+// This implementation is a customization of the 'react-diff' module.
```
eliperelman: Can you move these comments to below the imports?

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+// https://github.com/cezary/react-diff
+
+import { Component, PropTypes } from 'react';
```
eliperelman: We don't use `PropTypes` from the react module as they have been removed/deprecated. For consistency, see how we import prop types in all other components: by importing from the `prop-types` package.

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export default class Diff extends Component {
+  static propTypes = {
+    inputA: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+    inputB: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+    type: PropTypes.oneOf(['chars', 'words', 'sentences', 'json', 'lines'])
+  };
+
+  getDefaultProps = () => ({
```
eliperelman: You don't need a function to get the default props, and additionally I don't see this function being used. Declare a static default props property:

```js
static defaultProps = {
  inputA: '',
  inputB: '',
  type: 'chars'
};
```

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+  chars: jsdiff.diffChars,
+  words: jsdiff.diffWords,
+  sentences: jsdiff.diffSentences,
+  json: jsdiff.diffJson,
+  lines: jsdiff.diffLines
+};
+
+export default class Diff extends Component {
+  static propTypes = {
+    inputA: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
eliperelman: These variables names could be improved to convey the intent behind them. For example, instead of `inputA`, we could name it `original`, and `inputB` could be named `updated`.

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  getDefaultProps = () => ({
+    inputA: '',
+    inputB: '',
+    type: 'chars'
+  });
+
+  render() {
+    let diff = fnMap[this.props.type](this.props.inputA, this.props.inputB);
```
eliperelman: This can be a ternary expression which would lend itself to being a `const`-preferred declaration:

```js
const diff = this.props.propTypes === 'lines'
  ? fnMap[this.props.type](
      this.props.inputA,
      this.props.inputB[{ newlineIsToken: true }]
    )
  : fnMap[this.props.type](this.props.inputA, this.props.inputB);
```

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+    // If the type equals our new 'lines' type, change the 'diff' assignement
+    // to pass an option to InputB.
+    if (this.props.propTypes === 'lines') {
+      diff = fnMap[this.props.type](
+        this.props.inputA,
+        this.props.inputB[{ newlineIsToken: true }]
+      );
+    }
+
+    // Change the original styling.
```
eliperelman: What was the original styling?

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+        backgroundColor = 'mistyrose';
+      }
+
+      const spanStyle = {
+        color,
+        backgroundColor
+      };
+
+      return (
+        <span key={index} style={spanStyle}>
```
eliperelman: We cannot use the `index` to drive the keys here as we are doing line sorting, which means we can end up in a situation where two lines of different content end up with the same key and cause rendering issues. You'll need to come up with a way to derive a key based on the content of the line.

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+        this.props.inputB[{ newlineIsToken: true }]
+      );
+    }
+
+    // Change the original styling.
+    const result = diff.map((part, index) => {
+      let color = 'grey';
+      let backgroundColor = null;
+
+      if (part.added) {
```
eliperelman: These CSS styles can be extracted into class names using CSS modules. See other uses of CSS files in this repo for some examples of how to do that, as opposed to using inline styles.

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+      };
+
+      return (
+        <span key={index} style={spanStyle}>
+          {part.value}
+        </span>
+      );
+    });
+
+    return <pre className="diff-result">{result}</pre>;
```
eliperelman: I don't see this class name used anywhere. Does it have some purpose with the library, or is it just added here for identification?

src/components/ScopeEditor/react-diff-custom.jsx
```diff
@@ -0,0 +1,68 @@
...
+// Add a new 'lines' type to fnMap that was not present in the original module.
+const fnMap = {
+  chars: jsdiff.diffChars,
+  words: jsdiff.diffWords,
+  sentences: jsdiff.diffSentences,
+  json: jsdiff.diffJson,
+  lines: jsdiff.diffLines
+};
+
+export default class Diff extends Component {
```
eliperelman: As mentioned elsewhere, since this component is named `Diff`, the filename should be `Diff.jsx`. Also, this doesn't seem to be specific to scopes, so we can move it to the `components` directory as a top-level item as `components/Diff/index.jsx`. That would give us the ability to use diff viewers in more places than this in the future.

src/components/ScopeEditor/index.jsx
```diff
@@ -83,6 +86,16 @@ export default class ScopeEditor extends Component {
     );
   }
 
+  renderDiff() {
+    return (
+      <Diff
+        inputA={this.state.oldScopes.join('\n')}
```
eliperelman: In the original bug we discussed the need for sorting the scopes prior to rendering them or showing them in the diff viewer so the diff would be accurate. Is that not needed or helpful?

MynahMarie:
@eliperelman Hi, I am really sorry for not continuing with this sooner, I had a contract that needed all my attention and had not time... I just recently got back to working on it, I've addressed most of the requested changes (the comments I gave a thumbs up are the changes I've addressed, still missing a few). Is it ok if I continue to work on this or is this PR now too outdated?

eliperelman:
> Is it ok if I continue to work on this or is this PR now too outdated?

No, feel free to continue. 😃

helfi92:
@MynahMarie Let us know when this is ready to be reviewed :)

Issue #271, opened 2017-08-09
two 3000+ task taskgroup inspectors hang nightly


escapewindow:
I'm hoping to be able to open 2-3 active nightly graphs without having a hung browser. It's possible this is a nightly issue, and possible it's a tools.tc.net issue, and certainly possible that both could use improvements.

```
11:27 <aki> two tabs with taskgraphs of 3000+tasks + treeherder seem to hang nightly
11:31 <aki> i wonder if we can lazy load the tasks in a task group, kind of like how gmaps doesn't load the entire world's tiles, just the ones in the area you're looking at
11:32 — aki waves hands vaguely
11:33 <@dustin> sounds like a good PR for you to make :)
11:33 <@dustin> I think it only loads the taskgraph, not fetching each individual task
11:34 <aki> i do have the use case where i'm following a single task, and it's keeping track of the taskgraph
```

We just unified the two views (task inspector / taskgroup inspector) which is an overall win for UX. Hanging when I'm viewing a single task + a second taskgroup isn't ideal, though; I'm wondering if there's anything we can do on the tools.tc.net side to make this better.

eliperelman:
Here's the code for the loading of tasks, which is a recursive call:

https://github.com/taskcluster/taskcluster-tools/blob/65f7f1237cd701873ce350327c3c9df1cea8ef68/src/views/UnifiedInspector/Inspector.js#L134

Upon looking at this, I do notice a couple problems:

- The `updateLocalHistory` and `createGroupListener` is done every iteration, and should probably only be done if there is no `token`.
- We only load 100 tasks at a time, we could probably raise this or make it a little smarter.
- All this code is being `await`ed. We could probably put each iteration in a `requestIdleCallback` and try not to hang as much.

There is probably more we could do, too.

escapewindow:
Filed on the Fx side as [bug 1399229](https://bugzilla.mozilla.org/show_bug.cgi?id=1399229) .

JohanLorenzo:
I've been noticing such an issue for about a month, but I hadn't thought this is because of Task Group inspector, until last Friday. I'm on Linux, I thought this was a system issue, until I realize a text editor didn't show the symptom below:

## Symptoms
Type some in text in a text field of whatever page and notice the input stops from time to times. From times to times, the input hangs. Sometimes because of this hang, the key gets repeated more than 10 times (as in "buggggggggggggggggggg").

## Data set

[...]

PR #514, opened 2018-04-04
Optimize tasks rendering

helfi92:
Refer to https://github.com/taskcluster/taskcluster-tools/issues/271#issuecomment-378257496 for details.

### DIFF
--- a/src/views/UnifiedInspector/Inspector.jsx
+++ b/src/views/UnifiedInspector/Inspector.jsx
@@ -4,7 +4,7 @@ import { LinkContainer } from 'react-router-bootstrap';
 import { Row, Col, Nav, NavItem, Button } from 'react-bootstrap';
 import Icon from 'react-fontawesome';
 import { WebListener, request } from 'taskcluster-client-web';
-import { isNil } from 'ramda';
+import { isNil, equals } from 'ramda';
 import Helmet from 'react-helmet';
 import PropsRoute from '../../components/PropsRoute';
 import Error from '../../components/Error';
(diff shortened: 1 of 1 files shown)


src/views/UnifiedInspector/GroupDetailsTaskRow.jsx
```diff
@@ -0,0 +1,34 @@
+import { PureComponent } from 'react';
```
eliperelman: I don't understand what extracting this into a separate component achieves. Any insight?

src/views/UnifiedInspector/Inspector.jsx
```diff
@@ -207,7 +207,7 @@ export default class Inspector extends PureComponent {
               ({ status }) => status.taskId === task.status.taskId
             );
 
-            if (index > -1) {
+            if (index > -1 && !equals(reduction[index], task)) {
```
eliperelman: I'm not sure this is going to work as expected. This says "if the item is a match AND if they aren't equal". If either of these end of being false, then we add it onto the end of the array. I think the equality check is supposed to be **inside** the index match check. Does that make sense?

helfi92: Yes, makes a lot of sense. Thank you.

src/views/UnifiedInspector/GroupDetailsTaskRow.jsx
helfi92: With the current setup, I don't think it achieves anything. In the future, if there is a prop that changes in `GroupDetailsTaskRow`, then this won't re-render `GroupDetails`. FWIW, we should probably remove it because this is all premature optimization.

PR #516, opened 2018-04-05
We should show when task is created in index browser

fiennyangeln:
Done by creating taskDetail as state to get the created field

Bug 1212940

@djmitche please review:D

### DIFF
--- a/src/views/IndexBrowser/EntryView.jsx
+++ b/src/views/IndexBrowser/EntryView.jsx
@@ -47,14 +47,14 @@ export default class EntryView extends PureComponent {
       return this.setState({ error: null, task: {} });
     }
 
-    const task = await props.index.findTask(
-      props.namespaceTaskId
-        ? `${props.namespace}.${props.namespaceTaskId}`
-        : props.namespace
-    );
-    const taskDetail = await props.queue.task(task.taskId);
-
     try {
+      const task = await props.index.findTask(
+        props.namespaceTaskId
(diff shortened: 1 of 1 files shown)


src/views/IndexBrowser/EntryView.jsx
```diff
@@ -75,9 +80,11 @@ export default class EntryView extends PureComponent {
   }
 
   renderTask() {
-    const { error, task } = this.state;
+    const { error, task, taskDetail } = this.state;
 
     if (error && error.response.status === 404) {
+      console.log(error);
```
djmitche: let's remove this :)

src/views/IndexBrowser/EntryView.jsx
```diff
@@ -47,19 +47,24 @@ export default class EntryView extends PureComponent {
       return this.setState({ error: null, task: {} });
     }
 
+    const task = await props.index.findTask(
+      props.namespaceTaskId
+        ? `${props.namespace}.${props.namespaceTaskId}`
+        : props.namespace
+    );
+    const taskDetail = await props.queue.task(task.taskId);
```
djmitche: These both need to be in the inside the `try` block, otherwise we get a spinner forever when there's no task.

fiennyangeln:
@djmitche done!

djmitche:
Looks lke some lint issues -- otherwise, if @eliperelman is happy, I'm happy..

fiennyangeln:
@djmitche ok done!

Issue #517, opened 2018-04-06
Status tab on the worker type detail is empty


helfi92:
Example: https://tools.taskcluster.net/aws-provisioner/gecko-1-b-linux/

This must be some sort of regression.

PR #519, opened 2018-04-09
Fix status tab not showing content

helfi92:
Fixes https://github.com/taskcluster/taskcluster-tools/issues/517.

### DIFF
--- a/src/views/AwsProvisioner/WorkerTypeStatus.jsx
+++ b/src/views/AwsProvisioner/WorkerTypeStatus.jsx
@@ -1,6 +1,7 @@
 import { Component } from 'react';
 import { object, shape, arrayOf } from 'prop-types';
 import { Table } from 'react-bootstrap';
+import { emptyStatusTab } from './styles.module.css';
 
 export default class WorkerTypeStatus extends Component {
   static propTypes = {
@@ -26,10 +27,6 @@ export default class WorkerTypeStatus extends Component {
         inst.zone === availabilityZone
     ).length;
 
-    if (running + pending === 0) {
-      return;
-    }
-
     return (
       <tr key={`${instTypeDef.instanceType}:${availabilityZone}`}>
         <td>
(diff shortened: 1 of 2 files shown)


src/views/AwsProvisioner/WorkerTypeStatus.jsx
```diff
@@ -39,10 +36,14 @@ export default class WorkerTypeStatus extends Component {
           <code>{availabilityZone}</code>
         </td>
         <td>
-          {running * instTypeDef.capacity} ({running} instances)
+          {running
```
eliperelman: This just outputs a zero. How is this preferable to what was here before?

src/views/AwsProvisioner/WorkerTypeStatus.jsx
```diff
@@ -56,6 +57,12 @@ export default class WorkerTypeStatus extends Component {
       ])
     ];
 
+    if (!availabilityZones.length) {
+      return (
+        <div className={emptyStatusTab}>Availability zones not available</div>
```
eliperelman: Is this messaging accurate? In what situation would there be no instances and no requests?

djmitche: When there's no work to do, and the existing instances have terminated, there will be no instances and no requests

eliperelman: OK, maybe we can have better messaging to reflect that, such as "No running availability zones" or something. This message makes me think that something went wrong.

djmitche: Perhaps "no running instances"?

src/views/AwsProvisioner/WorkerTypeStatus.jsx
helfi92: Having multiple cell entries with `0 (0 instances)` made my eyes bleed.

PR #525, opened 2018-04-25
Bug 1415868 - support actions with kind=hook

djmitche:


src/views/UnifiedInspector/ActionsMenu.jsx
```diff
@@ -407,33 +422,57 @@ export default class ActionsMenu extends PureComponent {
...
+    const hookPayload = jsone(action.hookPayload, context);
+    const { hookId, hookGroupId } = action;
+    // verify that the decision task has the appropriate in-tree:action-hook:.. scope
+    const expansion = await this.props.auth.expandScopes({
+      scopes: this.props.decision.scopes
     });
+    const expression = `in-tree:hook-action:${hookGroupId}/${hookId}`;
 
-    await actionsQueue.createTask(ownTaskId, newTask);
+    if (!satisfiesExpression(expansion.scopes, expression)) {
```
tomprince: If I'm reading this correctly, this just controls what can be done via the UI, but the hook can still be triggered by somebody with the appropriate hook scopes manually?

src/views/UnifiedInspector/ActionsMenu.jsx
```diff
@@ -84,6 +88,10 @@ export default class ActionsMenu extends PureComponent {
             validate
           };
 
+          if (!knownKind(action.kind)) {
```
jonasfj: Sweet! :)

src/views/UnifiedInspector/ActionsMenu.jsx
```diff
@@ -70,6 +73,7 @@ export default class ActionsMenu extends PureComponent {
       const groupActions = [];
       const actionInputs = {};
       const actionData = {};
+      const knownKind = kind => kind.match(/^(task|hook)$/);
```
jonasfj: nit: An array is probably more appropriate than a regex... as you would always want to match exact strings.
(just a comment, no need to fix)

PR #526, opened 2018-04-30
Bug 1457772 - Refresh button getting stuck

helfi92:


helfi92:
`loadWorkerTypes` fetches the list of worker types and puts them in `this.state.workerTypeSummaries`.

`loadPendingTasks` maps over `this.state.workerTypeSummaries` and extends the latter with the `pendingTask` property.

Clicking on `refresh` was triggering both `loadPendingTasks` and `loadPendingTasks` concurrently. As a result, `loadWorkerTypes` would finish last and overwrite the result from `loadPendingTasks`. What this PR does is make sure the worker types have been loaded before asynchronously loading the pending tasks.

helfi92:
Pushed a cleaner fix that doesn't require the removal of `componentDidUpdate`.

PR #528, opened 2018-05-11
Bug 1460345 - Sort payload and extras on task inspector

Gabbyjose:
Created a 'sort' function on the TaskDetail component where you pass in the payload obj, sort it alphabetically by keys, and send that through JSON.stringify. Please let me know if you have any questions or any fixes I could make!


src/views/UnifiedInspector/TaskDetails.jsx
```diff
@@ -93,15 +93,21 @@ export default class TaskDetails extends PureComponent {
...
   render() {
     const { status, task } = this.props;
-
     if (!status || !task) {
       return null;
     }
-
     const { error } = this.state;
-    const { metadata } = task;
+    const { metadata, payload } = task;
```
helfi92: Instead of calling `sortPayload` twice below, you can sort here and store the result in a variable.

Gabbyjose:
Hi Eli,

Thanks so much for the quick feedback! I’ll work on these changes now and let you know if I run into any trouble.

Best,
Gabby

src/views/UnifiedInspector/TaskDetails.jsx
```diff
@@ -102,6 +103,8 @@ export default class TaskDetails extends PureComponent {
 
     const { error } = this.state;
     const { metadata } = task;
+    const payload = deepSortObject(task.payload);
+    const extra = deepSortObject(task.extra);
```
helfi92: Oops, I thought `deepSortObject` was operating on the same parameter. In that case, inlining it inside the render method works too. I am okay with both ways.

Gabbyjose:
Thank you!!

PR #531, opened 2018-05-29
 Redeployable tools site base, support for TASKCLUSTER_ROOT_URL

eliperelman:
I have added the relevant environment variables for Heroku. I will also handle the client-web stuff in a follow-up.


src/auth/DevelopmentLoginMenuItem.jsx
```diff
@@ -10,12 +10,9 @@ export default class DevelopmentLoginMenuItem extends PureComponent {
...
     query.set(
       'description',
-      'Temporary client for developing and testing taskcluster-tools'
-    );
-    window.open(
-      `https://tools.taskcluster.net/auth/clients/new?${query.toString()}`,
-      '_blank'
+      'Temporary client for developing and testing this repository'
     );
+    window.open(`/auth/clients/new?${query.toString()}`, '_blank');
```
djmitche: I think we'll want to work on this -- maybe an env var to specify the rootUrl to get the credentials from (DEVELOPMENT_LOGIN_ROOT_URL).  But that can happen in a followup.

src/utils.js
```diff
@@ -120,3 +121,5 @@ export const loadable = loader =>
     loading: Loading,
     loader
   });
+
+export const urls = withRootUrl(process.env.TASKCLUSTER_ROOT_URL);
```
djmitche: very cool :)

.env
```diff
@@ -0,0 +1,10 @@
+APPLICATION_NAME="Taskcluster"
+AUTH0_DOMAIN="auth.mozilla.auth0.com"
+AUTH0_CLIENT_ID="TBD"
+AUTH0_AUDIENCE="login.taskcluster.net"
+AUTH0_SCOPE="taskcluster-credentials openid profile"
+AUTH0_RESPONSE_TYPE="token id_token"
+OIDC_PROVIDER="mozilla-auth0"
+PORT=9000
+SIGN_IN_METHODS="development"
+TASKCLUSTER_ROOT_URL="taskcluster.net"
```
djmitche: TASKCLUSTER_ROOT_URL should be `https://taskcluster.net`.

src/auth/DevelopmentLoginMenuItem.jsx
djmitche: Hm, really we will want to use `TASKCLUSTER_ROOT_URL`, since we'll want to get the development login from the same taskcluster instance as we're calling APIs against.  So I tihnk this would be something like `urls.ui('/auth/clients/new?${query.toString()}')`

PR #540, opened 2018-06-29
Bug 1472025 - Blank pages while exploring workers

helfi92:
Some tasks only have `firstClaim`, `workerGroup`, and `workerId` property fields because I guess they weren't claimed. This pull-request makes sure all workers returned from the response are rendered.

### DIFF
--- a/src/views/Workers/WorkerManager.jsx
+++ b/src/views/Workers/WorkerManager.jsx
@@ -267,71 +267,72 @@ export default class WorkerManager extends PureComponent {
           <tbody>
             {!loading &&
               workers &&
-              workers.workers.map(
-                (worker, index) =>
-                  worker.state && (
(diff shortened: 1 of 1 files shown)


djmitche:
I don't understand from the description what the issue was.  Is this a bug in the queue implementation, or something we should add to the API docs in the queue?

helfi92:
Sorry, I should have been clearer. For every worker in a given worker-type we make a call to `loadStatus` to get more information about the latest task that ran in that worker. If a worker has no latest task then we don't do anything.

The reason why some workers were not being rendered on the screen is because of https://github.com/taskcluster/taskcluster-tools/pull/540/files#diff-77ab64cd679262c812edf592fc5bed0fL272. If a worker has no latest task then we don't augment the worker with its latest task state, therefore not render the worker.

PR #541, opened 2018-07-02
Use 'purge-cache' as the special-case name for purging caches

djmitche:
"purge" is a bit general - what are we purging? :)


djmitche:
@helfi92 please merge if you're OK with this :)

PR #543, opened 2018-07-06
Bug 1472792 - select the lowest-ordered action of a given name

djmitche:
https://github.com/taskcluster/taskcluster-docs/pull/270

We'll need to do the same for treeherder.

This can be tested on https://treeherder.mozilla.org/#/jobs?repo=try&revision=0e08ac79027a3395abf12cfd6fff0f6d4a442913&selectedJob=186904294 which has both a mochitest (that should get a mochitest-specific retrigger action) and other tasks (that should get a "regular" retrigger)

### DIFF
--- a/src/views/UnifiedInspector/ActionsMenu.jsx
+++ b/src/views/UnifiedInspector/ActionsMenu.jsx
@@ -77,6 +77,29 @@ export default class ActionsMenu extends PureComponent {
 
       if (actions && actions.actions) {
         actions.actions.forEach(action => {
+          if (!knownKinds.includes(action.kind)) {
+            return;
+          }
+
+          // if an action with this name has already been selected,
+          // don't consider this version
+          if (actionData[action.name]) {
+            return;
+          }
+
+          // add the action if it is a match; otherwise bail out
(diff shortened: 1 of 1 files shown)


src/views/UnifiedInspector/ActionsMenu.jsx
```diff
@@ -76,22 +77,18 @@ export default class ActionsMenu extends PureComponent {
...
-
-          actionInputs[action.name] = safeDump(
-            jsonSchemaDefaults(schema) || {}
-          );
-          actionData[action.name] = {
-            action,
-            validate
-          };
-
+        sortBy(actions.actions, action => action.order).forEach(action => {
```
helfi92: @djmitche Logging `action.order` to the console shows `undefined`. Is this expected?

djmitche:
Oh, interesting, it's not even defined in [the schema](https://github.com/taskcluster/taskcluster-references/blob/master/schemas/action-schema-v1.yml).

djmitche:
OK, updated to just use the order of the array in `actions.json`.

PR #544, opened 2018-07-09
Upgrade pulse-inspector to tc-events v1

ydidwania:
The client implementation is almost ready. As of now there are issues with CORS headers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "taskcluster-client-web": "^8.0.4",
     "taskcluster-lib-scopes": "^1.9.0",
     "taskcluster-lib-urls": "^10.0.0",
-    "urlencode": "^1.1.0",
     "ws-shell": "0.1.5",
     "www-authenticate": "0.6.2",
     "xml2js": "^0.4.19"
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "deepmerge": "^1.5.2",
     "docker-exec-websocket-server": "^1.3.1",
     "dotenv": "^5.0.1",
-    "eventsource": "^1.0.5",
     "font-awesome": "^4.7.0",
     "font-awesome-svg-png": "^1.1.5",
     "fscreen": "^1.0.2",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "deepmerge": "^1.5.2",
     "docker-exec-websocket-server": "^1.3.1",
     "dotenv": "^5.0.1",
+    "eventsource": "^1.0.5",
     "font-awesome": "^4.7.0",
     "font-awesome-svg-png": "^1.1.5",
     "fscreen": "^1.0.2",
@@ -78,6 +79,7 @@
     "taskcluster-client-web": "^8.0.4",
     "taskcluster-lib-scopes": "^1.9.0",
     "taskcluster-lib-urls": "^10.0.0",
+    "urlencode": "^1.1.0",
     "ws-shell": "0.1.5",
     "www-authenticate": "0.6.2",
     "xml2js": "^0.4.19"
--- a/src/views/PulseInspector/index.jsx
+++ b/src/views/PulseInspector/index.jsx
@@ -11,10 +11,11 @@ import {
   FormControl,
   ControlLabel
 } from 'react-bootstrap';
-import { WebListener } from 'taskcluster-client-web';
 import { nice } from 'slugid';
 import { pick } from 'ramda';
 import { parse, stringify } from 'qs';
+import { encode } from 'urlencode';
+import EventSource from 'eventsource';
 import MessageRow from './MessageRow';
 import HelmetTitle from '../../components/HelmetTitle';
 import { urls } from '../../utils';
@@ -46,8 +47,6 @@ export default class PulseInspector extends PureComponent {
       bindings,
       listening: false
     });
-
-    this.createListener(bindings);
   }
 
   componentWillUnmount() {
@@ -62,7 +61,6 @@ export default class PulseInspector extends PureComponent {
       const bindings = this.getBindingsFromProps(nextProps);
 
       this.setState({ bindings, listening: false });
-      this.createListener(bindings);
     }
   }
 
@@ -83,13 +81,14 @@ export default class PulseInspector extends PureComponent {
     }
 
     try {
-      const listener = new WebListener({
-        rootUrl: process.env.TASKCLUSTER_ROOT_URL
+      const jsonBindings = encode(JSON.stringify({ bindings }));
+      const evtUrl = `http://taskcluster-events-staging.herokuapp.com/api/events/v1/connect/?bindings=${jsonBindings}`;
+      const listener = new EventSource(evtUrl);
+
+      listener.addEventListener('message', this.handleListenerMessage);
+      listener.addEventListener('error', e => {
+        throw new Error(e.data);
       });
-
-      bindings.map(binding => listener.bind(binding));
-
-      listener.on('message', this.handleListenerMessage);
 
       this.listener = listener;
 
@@ -147,7 +146,7 @@ export default class PulseInspector extends PureComponent {
 
   handleStartListening = () => {
     this.setState({ listening: true });
-    this.listener.connect();
+    this.createListener(this.state.bindings);
   };
 
   /** Set expanded message, note we rely on object reference comparison here */
--- a/src/views/PulseInspector/index.jsx
+++ b/src/views/PulseInspector/index.jsx
@@ -79,14 +79,15 @@ export default class PulseInspector extends PureComponent {
     }
 
     const jsonBindings = encodeURIComponent(JSON.stringify({ bindings }));
-    const eventsUrl = `https://taskcluster-events-staging.herokuapp.com/api/events/v1/connect/?bindings=${jsonBindings}`;
-    const listener = new EventSource(eventsUrl);
+    const listener = new EventSource(
+      urls.api('events', 'v1', 'connect/?bindings=') + jsonBindings
+    );
 
     listener.addEventListener('message', this.handleListenerMessage);
-    listener.onerror = err => {
+    listener.addEventListener('error', err => {
       this.setState({ listeningError: err.data });
       this.handleStopListening();
-    };
+    });
 
     this.listener = listener;
 
--- a/src/views/PulseInspector/index.jsx
+++ b/src/views/PulseInspector/index.jsx
@@ -14,7 +14,6 @@ import {
 import { nice } from 'slugid';
 import { pick } from 'ramda';
 import { parse, stringify } from 'qs';
-import { encode } from 'urlencode';
 import EventSource from 'eventsource';
 import MessageRow from './MessageRow';
 import HelmetTitle from '../../components/HelmetTitle';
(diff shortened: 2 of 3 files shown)


src/views/PulseInspector/index.jsx
```diff
@@ -11,10 +11,11 @@ import {
   FormControl,
   ControlLabel
 } from 'react-bootstrap';
-import { WebListener } from 'taskcluster-client-web';
 import { nice } from 'slugid';
 import { pick } from 'ramda';
 import { parse, stringify } from 'qs';
+import { encode } from 'urlencode';
```
helfi92: The browser exposes `encodeURIComponent` that you can use.

src/views/PulseInspector/index.jsx
```diff
@@ -83,13 +81,14 @@ export default class PulseInspector extends PureComponent {
     }
 
     try {
-      const listener = new WebListener({
-        rootUrl: process.env.TASKCLUSTER_ROOT_URL
-      });
+      const jsonBindings = encode(JSON.stringify({ bindings }));
+      const evtUrl = `http://taskcluster-events-staging.herokuapp.com/api/events/v1/connect/?bindings=${jsonBindings}`;
+      const listener = new EventSource(evtUrl);
```
helfi92:
```js
const listener = new EventSource(`http://taskcluster-events-staging.herokuapp.com/api/events/v1/connect/?bindings=${jsonBindings}`);
```

src/views/PulseInspector/index.jsx
```diff
@@ -83,13 +81,14 @@ export default class PulseInspector extends PureComponent {
...
 
-      bindings.map(binding => listener.bind(binding));
-
-      listener.on('message', this.handleListenerMessage);
+      listener.addEventListener('message', this.handleListenerMessage);
+      listener.addEventListener('error', e => {
+        throw new Error(e.data);
+      });
```
helfi92: There doesn't seem to be a `data` field in the error:

screen shot 2018-07-10 at 10 11 19 am

src/views/PulseInspector/index.jsx
eliperelman: This URL will probably need to either come from the environment or taskcluster-lib-urls. We don't want this hardcoded when the cluster changes.

src/views/PulseInspector/index.jsx
eliperelman: There's also no point in throwing an error we don't handle. I would either skip handling the error, or setting it in state to be handled.

src/views/PulseInspector/index.jsx
helfi92: Agreed. This should probably be coming from tc-lib-urls.

ydidwania: Agreed. This is only till we shift out of heroku to `events.taskcluster.net`

src/views/PulseInspector/index.jsx
ydidwania: I want this `error` to be caught in the outer `try..catch` block. How can I do that?

src/views/PulseInspector/index.jsx
helfi92: I think we can still use `tc-lib-urls` even though we're using the staging url.

package.json
```diff
@@ -34,6 +34,7 @@
     "deepmerge": "^1.5.2",
     "docker-exec-websocket-server": "^1.3.1",
     "dotenv": "^5.0.1",
+    "eventsource": "^1.0.5",
```
helfi92: The browser expose an eventSource interface https://developer.mozilla.org/en-US/docs/Web/API/EventSource that we can use. `eventsource` might explain why there was a cache-control header sent with the request.

djmitche:
Using tc-lib-urls would require a "fake" rootUrl of http://taskcluster-events-staging.herokuapp.com (unrelated: why not https??), so it doesn't get around the need to embed a weird URL in the code.

src/views/PulseInspector/index.jsx
eliperelman: I didn't notice this was in a `try` block, so what you have is correct. If `error` `e` is already an error, why throw `e.data` instead of `e`?

Also, let's rename `e` to `err` or `error` as `e` is synonymous with events.

ydidwania: `e` is the message from the server with various [fields](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Fields)
```
e = { event : error
      message: Actual error from server
      id: -
...}
```
Yes I'll rename it to `err`.

src/views/PulseInspector/index.jsx
```diff
@@ -83,23 +79,26 @@ export default class PulseInspector extends PureComponent {
     }
 
     try {
-      const listener = new WebListener({
-        rootUrl: process.env.TASKCLUSTER_ROOT_URL
-      });
-
-      bindings.map(binding => listener.bind(binding));
+      const jsonBindings = encodeURIComponent(JSON.stringify({ bindings }));
+      const evtUrl = `https://taskcluster-events-staging.herokuapp.com/api/events/v1/connect/?bindings=${jsonBindings}`;
```
eliperelman: Let's rename this `eventsUrl`. Note that while this is fine for testing, we won't merge this PR until the service has landed in place of the events service and can use taskcluster-lib-urls.

src/views/PulseInspector/index.jsx
```diff
@@ -147,7 +146,8 @@ export default class PulseInspector extends PureComponent {
 
   handleStartListening = () => {
     this.setState({ listening: true });
-    this.listener.connect();
+    this.setState({ listeningError: null });
```
eliperelman: There is already a `setState` on the line before this, so re-use that one:

```js
this.setState({ listening: true, listeningError: null });
```

src/views/PulseInspector/index.jsx
```diff
@@ -83,23 +79,26 @@ export default class PulseInspector extends PureComponent {
...
+        this.setState({ listeningError: err.data });
+        this.handleStopListening();
+      };
 
       this.listener = listener;
 
       return listener;
     } catch (err) {
-      this.setState({ listeningError: err });
+      // this.setState({ listeningError: err });
```
eliperelman: Is there nothing else that can throw here? If not, we can remove the try/catch, and if so, it needs to be set in state.

src/views/PulseInspector/index.jsx
ydidwania: Yes. I will shift to `tc-lib-urls` once we land the server side

src/views/PulseInspector/index.jsx
ydidwania: All listener errors will be handled by `listener.onerror` .
I didn't realize I left the `catch` block empty. Will remove the `try/catch`

src/views/PulseInspector/index.jsx
```diff
@@ -235,7 +230,7 @@ export default class PulseInspector extends PureComponent {
   renderListeningError() {
     return (
       <Alert bsStyle="danger" onDismiss={this.handleDismissListeningError}>
-        <strong>Listening Error:</strong> {this.state.listeningError.message}
+        <strong>Listening Error:</strong> {this.state.listeningError}
```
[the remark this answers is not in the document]
ydidwania: because `listeningError.message` never existed. It was always 
```
      this.setState({ listeningError: <error-message> });
```

src/views/PulseInspector/index.jsx
```diff
@@ -82,24 +78,23 @@ export default class PulseInspector extends PureComponent {
...
-      });
-
-      bindings.map(binding => listener.bind(binding));
+    const jsonBindings = encodeURIComponent(JSON.stringify({ bindings }));
+    const eventsUrl = `https://taskcluster-events-staging.herokuapp.com/api/events/v1/connect/?bindings=${jsonBindings}`;
+    const listener = new EventSource(eventsUrl);
 
-      listener.on('message', this.handleListenerMessage);
+    listener.addEventListener('message', this.handleListenerMessage);
+    listener.onerror = err => {
```
eliperelman: Can this also use `addEventListener`?

src/views/PulseInspector/index.jsx
```diff
@@ -82,24 +78,23 @@ export default class PulseInspector extends PureComponent {
...
     }
 
-    try {
-      const listener = new WebListener({
-        rootUrl: process.env.TASKCLUSTER_ROOT_URL
-      });
-
-      bindings.map(binding => listener.bind(binding));
+    const jsonBindings = encodeURIComponent(JSON.stringify({ bindings }));
+    const eventsUrl = `https://taskcluster-events-staging.herokuapp.com/api/events/v1/connect/?bindings=${jsonBindings}`;
```
eliperelman: This will need to use tc-lib-urls:

```js
import { urls } from '../../utils';

// ...

const eventsUrl = urls.api('events', 'v1', 'connect/?bindings=${jsonBindings}`);
```

Something like that.

src/views/PulseInspector/index.jsx
ydidwania: yes.

src/views/PulseInspector/index.jsx
eliperelman: If `TASKCLUSTER_ROOT_URL` is set to `https://taskcluster.net` it generates the first URL, but if the service is anything else or not set, it will generate that second URL.

@djmitche any comments?

ydidwania: Okay, i realized this. I had set the `rootUrl` to `https://taskcluster-events-staging.herokuapp.com/`
I fixed it now and made the change. 
I think we can land this now :)

src/views/PulseInspector/index.jsx
```diff
@@ -82,24 +78,24 @@ export default class PulseInspector extends PureComponent {
...
 
-    try {
-      const listener = new WebListener({
-        rootUrl: process.env.TASKCLUSTER_ROOT_URL
-      });
-
-      bindings.map(binding => listener.bind(binding));
+    const jsonBindings = encodeURIComponent(JSON.stringify({ bindings }));
+    const listener = new EventSource(
+      urls.api('events', 'v1', 'connect/?bindings=') + jsonBindings
```
eliperelman: Switch to template string:

```js
urls.api('events', 'v1', `connect/?bindings=${jsonBindings}`)
```

src/views/PulseInspector/index.jsx
```diff
@@ -82,24 +78,24 @@ export default class PulseInspector extends PureComponent {
...
+    );
 
-      listener.on('message', this.handleListenerMessage);
+    listener.addEventListener('message', this.handleListenerMessage);
+    listener.addEventListener('error', err => {
```
eliperelman: If we only need `data`, let's destructure it in:

```js
listener.addEventListener('error', ({ data }) => {
  this.setState({ listeningError: data });
```

src/views/PulseInspector/index.jsx
```diff
@@ -82,24 +78,24 @@ export default class PulseInspector extends PureComponent {
...
 
-      return listener;
-    } catch (err) {
-      this.setState({ listeningError: err });
-    }
+    return listener;
   }
 
-  handleListenerMessage = message => {
+  handleListenerMessage = msg => {
```
eliperelman: Since we don't need anything but data, let's destructure it out:

```js
handleListenerMessage = ({ data }) => {
  const message = JSON.parse(data);
```

This way we avoid the confusion of 2 message variables in the same scope.

PR #545, opened 2018-07-11
Stop the running bar animation

helfi92:
Webrender doesn't support async animation of this property so it's causing a constant scene rebuild.

https://bugzilla.mozilla.org/show_bug.cgi?id=1465792


lissyx:
:1st_place_medal:

PR #551, opened 2018-07-19
Validate actions json schema before triggering

### LINKED ISSUE #548: actions.json is not fully validated before trying to trigger actions
I had an `actions.json` that was missing the `variables` field completely. That led to the following error when trying to trigger an action:
```
Error can't convert undefined to object
```

which is not really helpful.

Thankfully, the js debugger got me to:
https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/UnifiedInspector/ActionsMenu.jsx#L447

It turns out `actions` is used without validation of its schema, and `variables` is required per said schema. A schema validation error would have been more useful.

helfi92:
Resolves #548.

### DIFF
--- a/src/views/UnifiedInspector/ActionsMenu.jsx
+++ b/src/views/UnifiedInspector/ActionsMenu.jsx
@@ -11,12 +11,11 @@ import clone from 'lodash.clonedeep';
 import jsonSchemaDefaults from 'json-schema-defaults';
 import { safeLoad, safeDump } from 'js-yaml';
 import { satisfiesExpression } from 'taskcluster-lib-scopes';
-import libUrls from 'taskcluster-lib-urls';
 import ModalItem from '../../components/ModalItem';
 import Markdown from '../../components/Markdown';
 import CodeEditor from '../../components/CodeEditor';
 import Code from '../../components/Code';
-import { parameterizeTask } from '../../utils';
+import { parameterizeTask, urls } from '../../utils';
 
 export default class ActionsMenu extends PureComponent {
   static propTypes = {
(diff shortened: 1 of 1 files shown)


src/views/UnifiedInspector/ActionsMenu.jsx
```diff
@@ -11,6 +11,7 @@ import clone from 'lodash.clonedeep';
 import jsonSchemaDefaults from 'json-schema-defaults';
 import { safeLoad, safeDump } from 'js-yaml';
 import { satisfiesExpression } from 'taskcluster-lib-scopes';
+import libUrls from 'taskcluster-lib-urls';
```
eliperelman: Why not:

```js
import { urls } from '../../utils';
```

?

src/views/UnifiedInspector/ActionsMenu.jsx
```diff
@@ -51,6 +52,20 @@ export default class ActionsMenu extends PureComponent {
     };
   }
 
+  async componentDidMount() {
+    const actionsJsonSchemaResponse = await fetch(
+      libUrls.schema(
```
eliperelman: I think this should be (by using the already-set `TASKCLUSTER_ROOT_URL` from `urls` set in `utils`):

```js
urls.schema('common', 'action-schema-v1.json')
```

src/views/UnifiedInspector/ActionsMenu.jsx
```diff
@@ -428,6 +443,11 @@ export default class ActionsMenu extends PureComponent {
     const { validate, action } = actionData[name];
     const input = safeLoad(form);
     const valid = validate(input);
+    const validActionsJson = this.validateActionsJson(actions);
```
eliperelman: Is it possible to submit this/run this function prior to `componentDidMount` completing?

helfi92: Not really. To trigger that function, the user needs to open the dialog and click on the trigger button. By that time, the `componentDidMount` would have already run.

PR #552, opened 2018-07-20
Stop using WebListener in tools

ydidwania:
When we landed #544 , couple tools services like the `InteractiveConnect` and `UnifiedInspector` still used `WebListener`.
This PR makes them free of that. i have retained the heroku staging url here for some testing by @eliperelman 
Once tested, I will change to  `tc-libUrls`

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -98,24 +98,6 @@ export const parameterizeTask = task =>
     }
   );
 
-export const createListener = options => {
-  let { bindings } = options;
-
-  if (!bindings) {
-    bindings = options.exchanges.map(exchange =>
-      (({ exchange, routingKeyPattern }) => ({
-        exchange,
-        routingKeyPattern
-      }))(options.queueEvents[exchange](options.routingKey))
-    );
-  }
-
-  return new EventSource(`
-    https://taskcluster-events-staging.herokuapp.com/v1/connect/?bindings=${encodeURIComponent(
-      JSON.stringify({ bindings })
-    )}`);
-};
-
 // toArray :: a -> Array
 export const toArray = cond([[Array.isArray, identity], [T, of]]);
 
@@ -141,3 +123,22 @@ export const loadable = loader =>
   });
 
 export const urls = withRootUrl(process.env.TASKCLUSTER_ROOT_URL);
+
+export const createListener = options => {
+  const bindings =
+    options.bindings ||
+    options.exchanges.map(exchange =>
+      (({ exchange, routingKeyPattern }) => ({
+        exchange,
+        routingKeyPattern
+      }))(options.queueEvents[exchange](options.routingKey))
+    );
+
+  return new EventSource(
+    urls.api(
+      'events',
+      'v1',
+      `connect/?bindings=${encodeURIComponent(JSON.stringify({ bindings }))}`
+    )
+  );
+};
--- a/src/utils.js
+++ b/src/utils.js
@@ -99,12 +99,16 @@ export const parameterizeTask = task =>
   );
 
 export const createListener = options => {
-  const bindings = options.exchanges.map(exchange =>
-    (({ exchange, routingKeyPattern }) => ({
-      exchange,
-      routingKeyPattern
-    }))(options.queueEvents[exchange](options.routingKey))
-  );
+  let { bindings } = options;
+
+  if (!bindings) {
+    bindings = options.exchanges.map(exchange =>
+      (({ exchange, routingKeyPattern }) => ({
+        exchange,
+        routingKeyPattern
+      }))(options.queueEvents[exchange](options.routingKey))
+    );
+  }
 
   return new EventSource(`
     https://taskcluster-events-staging.herokuapp.com/v1/connect/?bindings=${encodeURIComponent(
--- a/src/utils.js
+++ b/src/utils.js
@@ -98,7 +98,7 @@ export const parameterizeTask = task =>
     }
   );
 
-export const createEventsListener = options => {
+export const createListener = options => {
   const bindings = options.exchanges.map(exchange =>
     (({ exchange, routingKeyPattern }) => ({
       exchange,
--- a/src/utils.js
+++ b/src/utils.js
@@ -99,15 +99,11 @@ export const parameterizeTask = task =>
   );
 
 export const createEventsListener = options => {
-  const bindings = [];
-
-  options.exchanges.map(exchange =>
-    bindings.push(
-      (({ exchange, routingKeyPattern }) => ({
-        exchange,
-        routingKeyPattern
-      }))(options.queueEvents[exchange](options.routingKey))
-    )
+  const bindings = options.exchanges.map(exchange =>
+    (({ exchange, routingKeyPattern }) => ({
+      exchange,
+      routingKeyPattern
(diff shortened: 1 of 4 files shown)


ydidwania:
@helfi92 can you test and review this?

src/views/UnifiedInspector/Inspector.jsx
```diff
@@ -340,14 +331,12 @@ export default class Inspector extends PureComponent {
 
     const { queueEvents } = this.props;
     const routingKey = { taskId };
-    const { exchange, routingKeyPattern } = queueEvents.artifactCreated(
+    const exchanges = ['artifactCreated'];
+    const listener = createEventsListener({
+      queueEvents,
+      exchanges,
```
helfi92: `exchanges: ['artifactCreated'],`

helfi92: `exchanges: ['artifactCreated'],`

src/views/UnifiedInspector/Inspector.jsx
```diff
@@ -340,14 +331,12 @@ export default class Inspector extends PureComponent {
...
       routingKey
```
helfi92: `routingKey: { taskId },`

src/views/UnifiedInspector/Inspector.jsx
```diff
@@ -296,28 +296,19 @@ export default class Inspector extends PureComponent {
...
-    );
-
-    const listener = new EventSource(`
-      https://taskcluster-events-staging.herokuapp.com/v1/connect/?bindings=${encodeURIComponent(
-        JSON.stringify({ bindings })
-      )}`);
+    ];
+    const listener = createEventsListener({
+      queueEvents,
+      exchanges,
```
helfi92:
```js
exchanges: [
 // ...
],
```

src/views/UnifiedInspector/Inspector.jsx
```diff
@@ -296,28 +296,19 @@ export default class Inspector extends PureComponent {
...
+      routingKey
```
helfi92: `routingKey: { taskGroupId },`

src/views/InteractiveConnect/InteractiveConnect.jsx
```diff
@@ -115,28 +115,19 @@ export default class InteractiveConnect extends PureComponent {
...
-
-    const listener = new EventSource(`
-      https://taskcluster-events-staging.herokuapp.com/v1/connect/?bindings=${encodeURIComponent(
-        JSON.stringify({ bindings })
-      )}`);
+    ];
+    const listener = createEventsListener({
+      queueEvents,
+      exchanges,
+      routingKey
```
helfi92: `routingKey: { taskId },`

src/views/InteractiveConnect/InteractiveConnect.jsx
```diff
@@ -115,28 +115,19 @@ export default class InteractiveConnect extends PureComponent {
...
-    );
-
-    const listener = new EventSource(`
-      https://taskcluster-events-staging.herokuapp.com/v1/connect/?bindings=${encodeURIComponent(
-        JSON.stringify({ bindings })
-      )}`);
+    ];
+    const listener = createEventsListener({
+      queueEvents,
+      exchanges,
```
helfi92:
```js
exchanges: [
  // ...
],
```

src/views/InteractiveConnect/InteractiveConnect.jsx
```diff
@@ -115,28 +115,19 @@ export default class InteractiveConnect extends PureComponent {
...
-        }))(queueEvents[binding](routingKey))
-      )
-    );
-
-    const listener = new EventSource(`
-      https://taskcluster-events-staging.herokuapp.com/v1/connect/?bindings=${encodeURIComponent(
-        JSON.stringify({ bindings })
-      )}`);
+    ];
+    const listener = createEventsListener({
```
helfi92: I think you can just rename to `createListener`.

ydidwania:
@helfi92 is it okay now?

src/views/InteractiveConnect/InteractiveConnect.jsx
```diff
@@ -115,28 +115,19 @@ export default class InteractiveConnect extends PureComponent {
 
     const { queueEvents } = this.props;
     const routingKey = { taskId };
-    const bindings = [];
-
-    [
+    const exchanges = [
```
helfi92: We can inline `exchanges` and `routingKey` since we don't reuse these variables.

src/utils.js
```diff
@@ -98,6 +98,24 @@ export const parameterizeTask = task =>
     }
   );
 
+export const createEventsListener = options => {
+  const bindings = [];
```
eliperelman: Since you are using `map`, you can directly return the new array without needing to create a placeholder array or `push`:

```js
const bindings = options.exchanges.map(exchange =>
  (({ exchange, routingKeyPattern }) => ({
    exchange,
    routingKeyPattern
  }))(options.queueEvents[exchange](options.routingKey))
);
```

src/utils.js
```diff
@@ -98,6 +98,20 @@ export const parameterizeTask = task =>
     }
   );
 
+export const createListener = options => {
```
helfi92: The pulse inspector should be making use of this utility function.

eliperelman: The pulse listener changes were landed last week with the new events instance, but that was backed out. It should work as expected when we ship tools and events again, but right now, this should only work in the parts changed here, not the pulse listener.

helfi92: Agreed. What I'm suggesting is to use `createListener` in `PulseInspector` so that we don't have two places with `new EventSource(...)`. Once we are happy with the PR, we can change `createListener` to use `urls.api('events', 'v1', ...)` before merging.

eliperelman: Ooo yeah, totally agree.

src/views/InteractiveConnect/InteractiveConnect.jsx
```diff
@@ -115,24 +114,23 @@ export default class InteractiveConnect extends PureComponent {
...
-      'taskException'
-    ].map(binding => listener.bind(queueEvents[binding](routingKey)));
-
-    listener.on('message', this.handleTaskMessage);
-    listener.on('reconnect', () => this.load(this.props));
-    listener.connect();
+
+    listener.addEventListener('message', msg =>
+      this.handleTaskMessage(JSON.parse(msg.data))
+    );
```
helfi92: What do we do in case of an `error` event? Should we be setting the `last-event-id` header somewhere in the client?

ydidwania: The `last-event-header` is automatically set by the `EventSource` object

src/utils.js
```diff
@@ -98,6 +98,24 @@ export const parameterizeTask = task =>
...
+  let { bindings } = options;
```
eliperelman: Instead of a `let`, this can be a `const` with:

```js
const bindings = options.bindings || options.exchanges.map(// ...
```

helfi92:
I'm ok with this PR. I'll leave the merge to @eliperelman.

PR #554, opened 2018-07-27
Bug 1478043 - remove scope inspector

djmitche:
For the moment, this links assume:... without a trailing star to the
corresponding role.  That will lead to some 404's sometimes, when the
named role doesn't exist but is matched by other, shorter roleIds.  But
it's better than not linking at all.


petemoore:
@djmitche @eliperelman I don't understand why we are deleting this tool. It was useful for discovering which roles/clients had a particular scope. Is there a better / different way of doing this?

PR #559, opened 2018-08-16
Bug 1483704 - Enhance CSP

helfi92:
Second attempt.

### DIFF
--- a/static.json
+++ b/static.json
@@ -22,7 +22,7 @@
       "X-Content-Type-Options": "nosniff",
       "X-XSS-Protection": "1; mode=block",
       "Referrer-Policy": "origin",
-      "Content-Security-Policy": "default-src 'none'; script-src 'self'; font-src 'self'; object-src 'none'; frame-ancestors 'self'",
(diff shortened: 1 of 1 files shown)


helfi92:
@imbstack I modified the CSPs to be somewhat similar to https://github.com/mozilla-frontend-infra/codetribute/blob/master/netlify.toml#L34 which has a similar setup as tools and also an [A+ on the observatory](https://observatory.mozilla.org/analyze/codetribute.mozilla.org).

> Content Security Policy: The page’s settings blocked the loading of a resource at https://tools.taskcluster.net/index.b9ec9f27fa848ae9cea54954892eb073.css (“default-src”).

To fix this issue, I added the `style-src` directive.

> Content Security Policy: The page’s settings blocked the loading of a resource at https://s.gravatar.com/avatar/4735e831dedefec5b5dd614aaa049692?s=480&r=pg&d=https%3A%2F%2Fcdn.auth0.com%2Favatars%2Fbs.png

To fix this issue, I added the `img-src` directive.

I'm not sure we currently have a way to test if this still works (no staging environment for tools to test out the CSPs). Do you think we should give it a try and rollback if still broken or should we create a staging environment.

I also added the `connect-src` in case we need it for events now that we are using `EventSource` for pulse messages.

PR #582, opened 2018-10-25
Bug 1474953 - catch task status 404s and show in table

davehouse:
When a task is not found (queue task and status requests), the worker explorer currently shows an error for the first task not found. As a user, I would like to see the worker information (first claim, and action buttons), and any valid tasks as well as the unknown task ids.

Example worker view:
https://tools.taskcluster.net/provisioners/releng-hardware/worker-types/gecko-t-osx-1010/workers/mdc1/t-yosemite-r7-273

Before:
image

After:
image

### DIFF
--- a/src/views/WorkerManager/WorkerManager.jsx
+++ b/src/views/WorkerManager/WorkerManager.jsx
@@ -128,13 +128,17 @@ export default class WorkerManager extends PureComponent {
           workerGroup,
           workerId
         );
+
+        // Display worker data while loading tasks.
+        this.setState({ worker, recentTasks: [{}] });
+
         const recentTasks = await this.handleLoadRecentTasks(
           worker.recentTasks
         );
 
-        this.setState({ worker, recentTasks, loading: false, error: null });
+        this.setState({ recentTasks, loading: false });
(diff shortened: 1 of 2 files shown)


davehouse:
Great! Thank you for the Nit fixes.

davehouse:
@helfi92 I noticed that this breaks sorting because the recentTasks are always(?) in reverse order and the sort reverses the list for the ones with times, but undefined ones are left in their original place.

davehouse:
I would like to see the tasks in their chronological order, which seems to be preserved in the task list despite the tasks not being found.
What do you think of reversing the list before we sort it? It is a kludge, but the api doesn't give the times although it gives the taskids and runids.

davehouse:
https://github.com/taskcluster/taskcluster-tools/commit/8563290fd8ec2b382e4372115b6cb837c321c251

helfi92:
Sounds good to me. Feel free to open the PR.

davehouse:
Thank you. I opened https://github.com/taskcluster/taskcluster-tools/pull/585 to add the .reverse()

PR #585, opened 2018-10-26
Bug1474953 show worker despite unknown tasks reverse tasklist

davehouse:
The taskcluster.queue.getWorker(https://docs.taskcluster.net/docs/reference/platform/taskcluster-queue/references/api#getWorker) recentTasks list is usually in reverse chronological order. And we sort by default in reverse chronological order. So, reverse the list when we get it and before we manually sort the elements.

### DIFF
--- a/src/views/WorkerManager/WorkerManager.jsx
+++ b/src/views/WorkerManager/WorkerManager.jsx
@@ -133,7 +133,7 @@ export default class WorkerManager extends PureComponent {
         this.setState({ worker, recentTasks: [{}] });
 
         const recentTasks = await this.handleLoadRecentTasks(
-          worker.recentTasks
+          worker.recentTasks.reverse()
         );
 
         this.setState({ recentTasks, loading: false });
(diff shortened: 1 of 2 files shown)


davehouse:
@helfi92 I'm sorry my github understanding is rusty, and I misunderstood the diff it showed me (with the already-merged changes).

davehouse:
When I test this change locally, it fixes the ordering when elements do not have times. Since it is still applying the .sort(), I think this does not create a new problem if the api results change and are not in the same order (we'd just see the undefined tasks out of order again).

helfi92:
You'll need to rebase on master. From your branch, I think you can do:

```bash
git pull https://github.com/taskcluster/taskcluster-tools.git master --rebase
git push origin bug1474953_show-worker-despite-unknown-tasks_reverse-tasklist -f
```

Hopefully there will be no conflicts when rebasing. If there are, you will need to fix the conflicts after `git pull`ing and then `git add <file>` followed by `git rebase --continue`. Let me know if that works.

davehouse:
I tried a rebase, but that would reapply my other commits and causes merge conflicts. So I re-made the change since it was a single line. https://github.com/taskcluster/taskcluster-tools/pull/586

Issue #587, opened 2018-10-29
New Heroku API key needed (yes, again)


hwine:
The key committed in #575 will not actually work after cutover to SSO. A new API key from a service account will be needed.

I'm sending emails to @helfi92 & @djmitche with details on approach. There need not be any downtime following those instructions.

edmorley:
Out of curiosity, what are the reasons for deploying from Travis rather than using Heroku's GitHub integration?

(I can think of one: being able to speed up the deploy slightly since re-uses the assets built on Travis, however now that builds are much faster with Neutrino 9 the benefit will be much less and so perhaps not worth the security/hassle trade-off)

helfi92:
taskcluster-tools doesn't use Neutrino v9. fwiw, we plan to move to taskcluster-web (uses neutrino v9) in the near future.

edmorley:
Ah yes. But taskcluster-tools could still switch to using the more conventional Heroku GitHub integration in the meantime (at a slight deploy-time regression) right? And then either later upgrade to Neutrino 9, or else wontfix in favour of switching to taskcluster-web depending on the expected timeframe for it being ready.

helfi92:
This sounds good to me.

hwine:
> 
> 
> This sounds good to me.

@helfi92 I'm confused about which "this" this is ;)  -- if you no longer need the heroku service account, please let me know in [bug 1504696](https://bugzilla.mozilla.org/show_bug.cgi?id=1504696)

helfi92:
Commented.

PR #589, opened 2018-11-05
Add Heroku GitHub integration

helfi92:
I added the following env vars in Heroku https://github.com/taskcluster/taskcluster-tools/blob/46b8ccb2f209f4dc7a5ea667be29d0233a9b82b5/.env. I did however notice there were [2 additional variables](https://travis-ci.org/taskcluster/taskcluster-tools/settings) only defined in the Travis site, namely `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. I don't see them being used in the repository. Do we need them?

Once this is merged and things look ok, we should enable automatic deploys from Heroku.

Relates to #587.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,11 +24,3 @@ after_failure:
 script:
 - yarn lint
 - travis_wait yarn build
-deploy:
-  app: taskcluster-tools
-  provider: heroku
-  skip_cleanup: true
-  on:
-    branch: master
-  api_key:
-    secure: at4cJRmc98kgJSypIeVDWOZP62Hk707d4Q4GhZ1Xr7QsZQV8rIJpbTcaS+naTqPp5tgrigu6if1LFpwCa6mjVACe6HxqeHqfUrmWPU/4/139GG0Jh240Z1CmyOGT3CNfcBrquGcYrFY7y//EBMW2eBG6XW47lFlA3joWLGBAqVI=
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "lint": "neutrino lint",
     "debug": "node --inspect ./node_modules/.bin/neutrino start --require dotenv/config",
     "build": "neutrino build --require dotenv/config",
-    "precommit": "lint-staged"
+    "precommit": "lint-staged",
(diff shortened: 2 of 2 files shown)


eliperelman:
One reason this was moved from Heroku to Travis was to not pay the cost of 2 builds per deploy. Meaning from a merge to production could take upwards of 30 minutes.

helfi92:
@eliperelman Which 2 builds are you referring to? This PR removes the build from Travis.

eliperelman:
The settings in Heroku, in the past, are to not deploy a master branch until the build passes in CI. This is to keep broken builds from being deployed to production. This means the master branch builds in both Travis and Heroku.

eliperelman:
I guess I'm wondering why you would want to remove Travis and risk deploying broken builds.

helfi92:
> This is to keep broken builds from being deployed to production

I thought deploys won't happen if `yarn build` exits with a non-zero code. I guess I was wrong. I'll add back Travis.

@eliperelman We are switching the Mozilla Heroku team to have SSO login and so there are extra security guidelines to follow to generate non-expiry access tokens (like regenerating the access token). Given that taskcluster-tools is going away soon, we thought we should just stop deploying from Travis even if there's a performance penalty to avoid having to generate a new access token. The new process at Mozilla requires creating a new service account to manage the app for non-expiry tokens.

djmitche:
I'm for whatever is quickest here to keep the focus on tc-web.  If that means we run a risk of deploying something with a broken build, that's OK by me -- I think it's rare to click "merge" on a red PR anyway.  And, since we don't have any testing, there's not much to prevent shipping a "broken" master anyway, outside of "so broken that `yarn build` fails" :)

helfi92:
@djmitche I noticed there were 2 additional variables only defined in the Travis site, namely AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY. I don't see them being used in the repository. Do we need them? I wonder why they were there in the first place.

djmitche:
We used to upload the built site from Travis to an S3 bucket, so those were probably in place to support that upload.  If you look up the Access Key ID in the IAM console, you can probably find the relevant IAM user and delete it.

PR #590, opened 2018-11-08
Bug 1478416 - upgrade tc-client-web to stop 4xx retries

djmitche:


djmitche:
@helfi92 is auto-deployment working now or do I need to do something to deploy?

helfi92:
I still haven't done the switch yet. I was planning to tackle that today or tomorrow. Auto-deployment should be working.

PR #595, opened 2018-12-26
Bug 1225243 - Support triggering tasks in response to pulse messages

ydidwania:
This PR allows editing of `bindings` as a part of the `hook` definition
I have considered `bindings` to be similar in behavior to the `schedule` property so a lot of code might be similar.
Tagging @djmitche  and @helfi92  for review

### DIFF
--- a/src/views/HooksManager/HookEditor.jsx
+++ b/src/views/HooksManager/HookEditor.jsx
@@ -468,7 +468,7 @@ export default class HookEditor extends PureComponent {
               ))}
             </ul>
             <div className="input-group">
-              <div>
+              <div className="div-horizontal">
                 <label className="control-label">Pulse Exchange</label>
                 <input
                   type="text"
@@ -478,7 +478,7 @@ export default class HookEditor extends PureComponent {
                   onChange={this.handleNewBindingsExchangeChange}
                 />
               </div>
-              <div>
+              <div className="div-horizontal">
                 <label className="control-label">Routing Key Pattern</label>
                 <input
                   type="text"
--- a/src/views/HooksManager/HookEditor.jsx
+++ b/src/views/HooksManager/HookEditor.jsx
@@ -59,7 +59,7 @@ export default class HookEditor extends PureComponent {
     this.state = {
       newScheduleValue: '',
       newBindingsExchangeValue: '',
-      newBindingsRkpValue: '',
+      newBindingsRkpValue: '#',
       hook: null,
       hookValidJson: false,
       triggerSchemaValidJson: false
@@ -187,13 +187,13 @@ export default class HookEditor extends PureComponent {
     const { newBindingsExchangeValue, newBindingsRkpValue } = this.state;
     const newBindingsValue = {
       exchange: newBindingsExchangeValue,
-      routingKeyPattern: newBindingsRkpValue || '#'
+      routingKeyPattern: newBindingsRkpValue
     };
 
     if (newBindingsExchangeValue) {
       this.setState({
         newBindingsExchangeValue: '',
-        newBindingsRkpValue: '',
+        newBindingsRkpValue: '#',
         hook: assocPath(
           ['bindings'],
           this.state.hook.bindings.concat(newBindingsValue),
@@ -472,7 +472,7 @@ export default class HookEditor extends PureComponent {
                 <label className="control-label">Pulse Exchange</label>
                 <input
                   type="text"
-                  className="form-control"
+                  className="form-control horizontal-left"
                   placeholder="exchange/<username>/some-exchange-name"
                   value={newBindingsExchangeValue}
                   onChange={this.handleNewBindingsExchangeChange}
@@ -482,7 +482,7 @@ export default class HookEditor extends PureComponent {
                 <label className="control-label">Routing Key Pattern</label>
                 <input
                   type="text"
-                  className="form-control"
+                  className="form-control horizontal-right"
                   placeholder="*.some-interesting-key.#"
                   value={newBindingsRkpValue}
                   onChange={this.handleNewBindingsRkpChange}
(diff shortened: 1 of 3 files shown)


src/views/HooksManager/HookEditor.jsx
```diff
@@ -166,6 +175,34 @@ export default class HookEditor extends PureComponent {
...
+      routingKeyPattern: newBindingsRkpValue
+    };
+
+    if (newBindingsExchangeValue && newBindingsRkpValue) {
+      this.setState({
+        newBindingsExchangeValue: '',
+        newBindingsRkpValue: '',
+        hook: assocPath(
+          ['bindings'],
+          [...clone(this.state.hook.bindings), newBindingsValue],
```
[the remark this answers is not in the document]
ydidwania: Hmm, that looks better. I did the `clone`maybe because it was done that way for `schedule`.
Will make this change

src/views/HooksManager/HookEditor.jsx
```diff
@@ -406,6 +444,55 @@ export default class HookEditor extends PureComponent {
...
+                  onChange={this.handleNewBindingsRkpChange}
+                />
+              </div>
+              <span className="input-group-btn">
+                <button
+                  className="btn btn-success"
+                  type="button"
+                  onClick={this.handleNewBindingsItem}>
+                  <Glyphicon glyph="plus" /> Add
+                </button>
```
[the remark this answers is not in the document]
ydidwania: Those are valid points. 
I actually wanted the two fields for `exchange` and `routingKeyPattern` to be horizontally linear instead of occupying the whole width as they do now. That  would look much better I suppose. Any idea how to do that?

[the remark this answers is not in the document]
ydidwania: image
Does this look better?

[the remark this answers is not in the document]
helfi92: Yeah, I think it looks a lot better :)

src/views/HooksManager/HookEditor.jsx
```diff
@@ -406,6 +450,57 @@ export default class HookEditor extends PureComponent {
...
+                  className="btn btn-success"
+                  type="button"
+                  disabled={!this.validBindingsItem()}
+                  onClick={this.handleNewBindingsItem}>
+                  <Glyphicon glyph="plus" /> Add
+                </button>
+              </span>
+            </div>
+          </div>
+        </div>
```
helfi92: Instead of

screen shot 2019-01-03 at 11 23 33 am

Let's have it look like

screen shot 2019-01-03 at 11 22 09 am


Differences:

1. The left border and right border of the first input are opposite. You can achieve that with the following style on the first input:

```css
border-top-right-radius: 0;
border-bottom-right-radius: 0;
border-top-left-radius: 4px;
border-bottom-left-radius: 4px;
```

2. The second input has a 0 right border radius

```css
border-top-right-radius: 0;
border-bottom-right-radius: 0;
```

3.  The `margin-top` property of the button is now set to `26px` instead of `25px`

Instead of hardcoding the values straight into the `js` file, you can make use of css modules similar to how it's done in https://github.com/taskcluster/taskcluster-tools/blob/master/src/views/Home/index.jsx#L9-L16. Simply create `styles.module.css` in the same directory as HooksEditor then you can import the styles from the js file. You can see how they're used all around the site.

src/views/HooksManager/HookEditor.jsx
```diff
@@ -57,6 +58,8 @@ export default class HookEditor extends PureComponent {
 
     this.state = {
       newScheduleValue: '',
+      newBindingsExchangeValue: '',
+      newBindingsRkpValue: '',
```
helfi92: `newBindingsRkpValue: '#',`

src/views/HooksManager/HookEditor.jsx
```diff
@@ -166,6 +175,40 @@ export default class HookEditor extends PureComponent {
...
+    bindings.splice(index, 1);
+
+    this.setState({ hook: { ...this.state.hook, bindings } });
+  };
+
+  handleNewBindingsItem = () => {
+    const { newBindingsExchangeValue, newBindingsRkpValue } = this.state;
+    const newBindingsValue = {
+      exchange: newBindingsExchangeValue,
+      routingKeyPattern: newBindingsRkpValue || '#'
```
helfi92: `routingKeyPattern: newBindingsRkpValue`

src/views/HooksManager/HookEditor.jsx
helfi92: Definitely.

src/views/HooksManager/styles.css
```diff
@@ -40,6 +40,16 @@
     display: inline-block;
     width: 50%;
   }
+  .horizontal-left {
```
helfi92: Rename to `.bindings-exchange-input`.

src/views/HooksManager/styles.css
```diff
@@ -40,6 +40,16 @@
...
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+    border-top-left-radius: 4px !important;
+    border-bottom-left-radius: 4px !important;
+  }
+  .horizontal-right {
```
helfi92: Rename to `.bindings-rkp-input`.

helfi92:
@ydidwania I created https://github.com/taskcluster/taskcluster-web/issues/352. Do you want to try implementing the same thing but in the new UI?

ydidwania:
@helfi92 I want to. It'll just take me some time to go through the new UI

helfi92:
That's okay. I'm also here to help :) Let's continue the discussion in taskcluster/taskcluster-web#352.

PR #596, opened 2019-01-24
Update Quickstart to use the proper version

helfi92:


src/views/QuickStart/index.jsx
```diff
@@ -22,7 +22,7 @@ import { info } from './styles.module.css';
 /* eslint-disable react/no-unused-state */
 
 const initialYaml = {
-  version: 0,
+  version: 1,
```
owlishDeveloper: This is not all - we need the whole yml to be `v1` (i.e. no `extra` section, etc. [See here](https://github.com/taskcluster/taskcluster/blob/master/services/github/docs/taskcluster-yml-v1.md) )

helfi92: Would there be anything breaking?

From https://github.com/taskcluster/taskcluster/blob/dfee9af2438503514b00b0df8dbbfb8032bda517/services/github/docs/taskcluster-yml-v0.md

> If you already use a version 0 file, we strongly recommend you switch to version 1. We will support the old version for a while though.

This commit was to resolve [Comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1515673#c1). Changing the whole yml should probably be another bug? If you think it can be a good-first-bug, we could create a new good-first-bug that you can mentor given your tc-github knowledge :) Let me know what works best.

PR #599, opened 2019-01-26
bug1520857-add many lastfire info

Biboswan:


Biboswan:
tools
 @djmitche

djmitche:
Could we hide that big wall of JSON behind some kind of click?

I think we will end up going back to make that a rendered markdown error, since I expect InsufficientScopes is the most common kind, so don't worry about parsing the JSON (it might not be valid anyway!)

djmitche:
Let's see what @helfi92 says when he's back!

helfi92:
The tools site will no longer be used soon. I wonder if we should bother with trying to have this functionality in tools. @djmitche thoughts?

djmitche:
I think that's OK.

helfi92:
> I think that's OK.

What do you think is okay? :)

djmitche:
Haha, sorry -- I think it's OK to skip implementing this in tools.

PR #610, opened 2019-11-06
Add a banner pointing users to new deployments

djmitche:
I dropped the banner sending people to the taskcluster-ui.herokuapp.com site since that will soon be the default.  I also dropped the "Welcome To Taskcluster" banner on Home since it didn't look good with this banner.


djmitche:
Good point, good idea :)

helfi92:
> Good point, good idea :)

Is it possible you forgot to commit the additional changes?

djmitche:
Apparently.. I wonder what I force-pushed..

I'll make a hotfix.  Thanks for spotting that!

